PR #86, opened 2024-04-09
Feature/ai service/refine ask pipeline v2

cyyeh:
- add sample dataset evaluation script
- remove duplicates of ask pipeline generation results in ask pipeline
- return original sql and summary instead of showing no relevant sql in ask details pipeline


paopa (approved):
there are some breaking changes for the ask details evaluation process. but, let's us skip them here. and fix in another PR because there are some changes here we need to put into the next release.

PR #101, opened 2024-04-11
update ask evaluation outputs path in streamlit app

imAsterSun:
Update ask evaluation streamlit path of outputs folder


cyyeh (approved):
LGTM

PR #102, opened 2024-04-11
fix(ai-service): correct the breaking change for ask details evaluation process

paopa:
in #86, we mentioned the breaking changes for the ask details evaluation process. and fixed them here. 😃


cyyeh (approved):
LGTM

PR #107, opened 2024-04-12
feat(wren-ui): Support deploy with show status on modeling page 

fredalai:
## Description

- `Deploy` Button
    - Status: `Undeployed changes`, `Deploying`, `synced`
    - Connect API

### UI
截圖 2024-04-12 上午9 58 23

截圖 2024-04-12 上午9 58 10

截圖 2024-04-12 上午9 57 44

## Tasks
 - [x] Add **Deploy** Button
 - [x] Show sync status

### DIFF
--- a/wren-ui/src/components/Deploy.tsx
+++ b/wren-ui/src/components/Deploy.tsx
@@ -1,0 +1,65 @@
+import { Button, Space, Typography } from 'antd';
+import CheckCircleOutlined from '@ant-design/icons/CheckCircleOutlined';
+import LoadingOutlined from '@ant-design/icons/LoadingOutlined';
+import WarningOutlined from '@ant-design/icons/WarningOutlined';
+import {
+  useDeployMutation,
+  useDeployStatusQuery,
+} from '@/apollo/client/graphql/deploy.generated';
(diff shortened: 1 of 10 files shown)


wren-ui/src/components/Deploy.tsx
```diff
@@ -0,0 +1,65 @@
...
+      ),
+    }[syncStatus] || ''
+  );
+};
+
+export default function Deploy() {
+  const [deployMutation, { loading: deploying }] = useDeployMutation();
+  const { data, loading } = useDeployStatusQuery({
+    pollInterval: 1000,
+  });
```
andreashimin, line 44: I have some concerns about long polling with one sec.
It seems like we will infinite polling from the page start to end, does it really need?

If it is necessary, I will suggest we can add a Visibility API to have a switch of it.

fredalai: According to the results of the previous discussion, since there are not many trigger behaviors for changing MDL, I first adjusted this version - Use trigger deployStatusQueryResult.refetch() for actions instead of the polling method to get the latest sync status.

I think... we can go through how to implement better ways if the product has complex behavior in the future.

andreashimin (approved):
LGTM

PR #133, opened 2024-04-18
chore(wren-ui): Enhance view node

fredalai:
## Description
 - handle view fields more than config.limit length (50)
 - more icon - visible dropdown menu


andreashimin (approved):
LGTM

PR #134, opened 2024-04-18
chore(wren-ai-service): refactor makefile command

paopa:
This PR aims to refactor the Makefile and also adjust the related document such as README, CI.


wren-ai-service/Makefile
```diff
@@ -1,61 +1,243 @@
...
+	@$(MAKE) check-poetry
+	@$(MAKE) install-python-dependencies
+	@$(MAKE) install-precommit-hooks
+	@$(MAKE) setup-envfile
+	@echo "$(GREEN)Development environment setup complete!$(RESET)"
 
-down:
-	make stop-wren-engine && \
-	docker compose -f docker/docker-compose.yml --env-file .env.prod down
+check-python:
```
cyyeh, line 20: as we said in requirements in README, we need python 3.12. so can we also check the python version should be 12.xxx here?

wren-ai-service/Makefile
```diff
@@ -1,61 +1,243 @@
...
+	@if docker container ls -a --format '{{.Names}}' | grep 'qdrant' 1> /dev/null; then \
+		docker rm qdrant > /dev/null; \
+		echo "$(GREEN)Qdrant container removed!$(RESET)"; \
+	else \
+		echo "$(BLUE)Qdrant container does not exist.$(RESET)"; \
+	fi
 
-run-wren-engine:
-	docker compose -f ./src/eval/wren-engine/docker-compose.yml --env-file ./src/eval/wren-engine/.env up -d
+start-wren-engine:
```
cyyeh, line 103: for the start-wren-engine part, users also need to prepare an etc folder first, and put the config.properties file inside and fill in environment variables.(generated from config.properties.example in the src/eval/wren-engine)

wren-ai-service/README.md
```diff
@@ -1,49 +1,104 @@
...
+  - If you change the host and port in env file, you could follow the format to access the API documentation and 
+  try the Web APIs. (`http://WREN_AI_SERVICE_HOST:WREN_AI_SERVICE_PORT`)
+- To stop the services, execute the following command:
+  ```shell
+  make stop
+  ```
 
-- copy `.env.prod.example` file to `.env.prod` and fill in the environment variables
+## Run all test cases
+- All the test cases are located in the `tests` folder and the following command will run all the test cases:
```
cyyeh, line 34: for make test to work, users need to start qdrant and wren-engine first, and create book_2 mdl file. you can see the command used before for reference:

```Makefile
test:
	poetry run python -m src.prepare_mdl_json --dataset_name book_2 && \
	make run-qdrant && \
	make run-wren-engine && \
	poetry run pytest -s $(args) && \
	make stop-all
```

wren-ai-service/README.md
```diff
@@ -1,49 +1,104 @@
...
-- in the `demo` folder, open three terminals
-  - in the first terminal, run `make prepare` to start the docker containers and `make run` to start the demo service
-  - in the second terminal, run `make ui` to start the wren-ui service
-  - in the third terminal, run `make ai` to start the wren-ai service
-- ports of the services:
-  - wren-engine: ports should be 8080
-  - wren-ai-service: port should be 5556
-  - wren-ui: port should be 3000
-  - qdrant: ports should be 6333, 6334
+## Interative the system with Demo application or UI
```
cyyeh, line 67: I am not sure if this is typo?

wren-ai-service/README.md
```diff
@@ -1,49 +1,104 @@
...
+
+- Go to the `../docker` folder and prepare the `.env.local` file
+- Execute the follow command to start the containers and choose App or UI to interact with the system
```
cyyeh, line 70: maybe we need to list `make start-ui` and `make start-app` commands explicitly here

cyyeh: also in order to make `make start-ui` work, we need to explicitly tell users these prerequisites

  - make sure the node version is v16.19.0
  - if you are using Python 3.12+, please also install `setuptools` in order to successfully install the dependencies of the wren-ui service

cyyeh: in order to make make start-app to work, users should run poetry install in the demo folder first

cyyeh:
It's a cool work! Thanks for your hard work :)

wren-ai-service/Makefile
paopa: I think it is a one-time behavior. I'd prefer to mention this in README instead of generating the config every time. or add another make file command to generate the config file. what do you think?

cyyeh: sure. it's fine to me

wren-ai-service/README.md
paopa: I removed them because in my case, I already started the dependent services so I don't want to start them every time, and also don't want to stop them after finishing the test cases. so, I think making the behavior more unit will be better.

PR #136, opened 2024-04-18
Feature/wren ai service/refine pipelines

cyyeh:
- refine prompts for ask and ask details pipelines
  - lower case for case-insensitive comparison
  - must use join if llm uses columns appeared in different tables
- refine data types for sample data's mdl
- add example data for ecommerce sample data
- fix dupicated column issue for demo app
- fix eval sample data issue


wren-ai-service/src/utils.py
```diff
@@ -1,3 +1,6 @@
+"""
+CAUTION: before running this code, please ensure the given dataset's mdl model is deployed already
+"""
```
paopa, line 3: curiously, why did you add this comment, seems we didn't directly run any code in this module.

paopa (approved):
overall LGTM. only have a minor comment for the code comment

PR #138, opened 2024-04-19
Fix(wren-ui): fix invalid session props cause saveDataSource action to fail

onlyjackfrost:
clear prior session props setting before running init sql.


andreashimin (approved):
LGTM 👍🏼  Thanks for fixing it

PR #139, opened 2024-04-19
Fix(wren-ui): fix incorrect data type in the sample data

onlyjackfrost:
- fix incorrect data type in the sample data
 - implement function to adding description on model/column


wwwy3y3 (approved):
LGTM

PR #149, opened 2024-04-22
fix wording in launcher

cyyeh:
Do you agree to help us by sending anonymous usage data? (yes/no, default is y) -> Do you agree to help us by sending anonymous usage data? (y/n, default is y)


onlyjackfrost (commented):
the user response should be exactly "y" or "n"

onlyjackfrost (approved):

PR #150, opened 2024-04-23
fix(wren-ui): Fix pg save data source missing await

andreashimin:
## Description
The onboarding saves PG data source that won't return errors because of missing await the promise of connection testing


onlyjackfrost (approved):
ＬＧＴＭ

PR #151, opened 2024-04-23
Fix(wren-ui): provide better error message

onlyjackfrost:
Provide better error message when using duckdb as the datasource
image


wwwy3y3 (approved):
LGTM

PR #153, opened 2024-04-23
feature(ai-service): MDL integration with View

paopa:
# Description

This PR aims to support the indexing process to store the view in the vector store. and also, update the prompt to hint LLM to use the view to simplify the SQL.

# Screenshots

image

image


cyyeh (approved):
Overall LGTM.

I think we need to have more test cases though in the future. Also, we may need to think if there is a better way to test various MDL integration cases?

PR #158, opened 2024-04-25
feature(ai-service): MDL calculated field integration

imAsterSun:


wren-ai-service/src/utils.py
```diff
@@ -207,6 +205,12 @@ def generate_ddls_from_semantics(
                     comment = f"-- {json.dumps(column['properties'])}\n  "
                 else:
                     comment = ""
+                if column["isCalculated"]:
+                    comment = comment + "-- This column is a Calculated Field\n  "
+                if "expression" in column:
```
cyyeh, line 210: I suppose we should only add expression if "isCalculated" is true?

imAsterSun: Sure. Code modified, thanks for the advice.

cyyeh (approved):
LGTM

PR #159, opened 2024-04-25
Feature/ai service/improve pipelines

cyyeh:
- remove "Data example" in e-commerce sample data
- adjust adding quotes in sql through using sqlglot instead of using llm
- update eval_sampledata.py by adding another file called sampledata_questions.json instead of writing all questions inside the testing script
- adjust ai pipelines prompts


onlyjackfrost (approved):
LGTM for the sample data

paopa (approved):

PR #166, opened 2024-04-25
Chore: update sample dataset model/column properties

onlyjackfrost:
allow setting model/column properties on demo dataset


wwwy3y3 (approved):
LGTM

PR #176, opened 2024-04-29
fix(wren-ui): update primary key validator for model form

fredalai:
## Description
 - Update the validator since the primary key is not the required field
 - Allow a user to clear the primary key value


andreashimin (approved):
LGTM

PR #197, opened 2024-05-06
fix(wren-ui): Fix diagram node height calculation

andreashimin:
## Description
The modeling page could not render after the user saved multiple views

## Reason
The diagram couldn't find view data to calculate the node with the most column height

## Tasks
- Replenish views data for calculation
- Add calculated field columns height calculation
- Rearrange all column height config


fredalai (approved):
LGTM

PR #199, opened 2024-05-06
Fix(wren-ai-service): Clear all documents before deploying the new MDL

paopa:
Currently, our document updates utilize policy overwriting, which identifies documents by their ID. However, this approach poses a risk as it heavily relies on the order of entries in the MDL. This dependency can lead to unintended outcomes that need addressing.

For example, consider the following scenario:
- Original MDL: 1 → A, 2 → B, 3 → C, 4 → D
- Updated MDL: 1 → A, 2 → B, 3 → C

After deploying the new MDL, the current vector store still retains:
- 1 → A, 2 → B, 3 → C, 4 → D

The ID 4 with document D remains in the vector store, which is not the intended behavior. We need to resolve this issue to ensure only relevant documents are preserved after updates.


cyyeh (approved):
LGTM

PR #202, opened 2024-05-07
fix(wren-ui): show the connected data source when on the original SQL mode only

fredalai:
## Description
 - Fix
   - show the connected data source when on the original SQL mode only
   - redirect to the home page if there is no relationship data needs to be saved

### UI
截圖 2024-05-07 下午12 19 53


andreashimin (approved):
LGTM

PR #205, opened 2024-05-07
fix(wren-ui): Fix svg embed definitions styles that affected user operation

andreashimin:
## Description
The SVG embed definitions have affected user operation.


fredalai (approved):
LGTM.

PR #208, opened 2024-05-07
Fix(wren-ui):  change migration file order to prevent relation data lost

onlyjackfrost:
CASCADE will lead to relation deletion when column drop


wwwy3y3 (approved):
LGTM

PR #213, opened 2024-05-08
fix(wren-ui): handle network error issues for asking a question

### LINKED ISSUE #175: Show internal server error message
Currently when a internal server error occurs, we only show 'internal server error'.
In this use case, this error occurs because wifi is down:
SCR-20240429-mmgv

Should show error message(e.g. network error, unable to connect to network....) to let users know what is wrong.

fredalai:
## Description
 Handle network error issue for https://github.com/Canner/WrenAI/issues/175

## Issue ticket number
closes #175

## Solution
 - Based on the reference background context, considering the **OTHERS** error code does not directly mean network error, I think we can pass an error message from the AI service to the web and show it on the UI at this stage.

### Reference background context

 - When there is no network, after the web backend sends the API request to the AI ​​Service, the final return data will be:
```json
{
  "status": "failed",
  "response": null,
  "error": {
    "code": "OTHERS",
    "message": "Connection error."
  }
}
```

 - Snippet code for AI Service

We can see the below snippet code of handle exception error:

https://github.com/Canner/WrenAI/blob/d24cdc1090fc9bdaa8722a720b02b1b2281b9786/wren-ai-service/src/web/v1/services/ask_details.py#L106-L113

https://github.com/Canner/WrenAI/blob/e997bd0d5142e4c40dfc72cda07e2e243808d884/wren-ai-service/src/web/v1/services/ask.py#L271-L278

## Tasks
 - [x] Defaults to internal server error if not mapped to short message (**shortMessage**)
 - [x] Show default message for network error when calling an API

## UI screenshots

 - Step 1: Go to /home page
 - Step 2: Turn off Wi-Fi
 - Step 3: Ask a question
 

截圖 2024-05-09 上午9 34 57

### DIFF
--- a/wren-ui/src/apollo/server/utils/error.ts
+++ b/wren-ui/src/apollo/server/utils/error.ts
@@ -8,6 +8,9 @@ export enum GeneralErrorCodes {
   NO_RELEVANT_DATA = 'NO_RELEVANT_DATA',
   NO_RELEVANT_SQL = 'NO_RELEVANT_SQL',
 
+  // Exception error for AI service (e.g., network connection error)
+  AI_SERVICE_UNDEFINED_ERROR = 'OTHERS',
+
   // Connector errors
   // duckdb
   INIT_SQL_ERROR = 'INIT_SQL_ERROR',
@@ -77,7 +80,9 @@ export const create = (
       originalError,
       code,
       message,
-      shortMessage: shortMessages[code],
+      shortMessage:
+        shortMessages[code] ||
+        shortMessages[GeneralErrorCodes.INTERNAL_SERVER_ERROR],
     },
   });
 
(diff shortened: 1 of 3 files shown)


wren-ui/src/apollo/server/utils/error.ts
```diff
@@ -8,6 +8,9 @@ export enum GeneralErrorCodes {
   NO_RELEVANT_DATA = 'NO_RELEVANT_DATA',
   NO_RELEVANT_SQL = 'NO_RELEVANT_SQL',
 
+  // Exception error for AI service (e.g., network connection error)
+  OTHERS = 'OTHERS',
```
wwwy3y3, line 12: What does `OTHERS` means ? Can we give a more specific error ?

wwwy3y3: I see. Didn't notice AI service error code called `OTHERS`. Maybe we can call it `AI_SERVICE_UNDEFINED_ERROR` in our enum ? since it basically means an error that AI service doesn't specifically define.

fredalai: Done by 7fb1273

andreashimin (approved):
LGTM

PR #214, opened 2024-05-08
feat(wren-ui): allow column case insensitive search for column selector

fredalai:
## Description
 - Allow column case insensitive search for column selector

### Before
case sensitive_before

### After
截圖 2024-05-08 下午1 23 12


andreashimin (approved):
LGTM

PR #215, opened 2024-05-08
fix(wren-ui): in pg connector, we return column types in enum

wwwy3y3:
## Description
Fix that postgres connector return native data types instead of types we defined in enum


wren-ui/src/apollo/server/connectors/types.ts
```diff
@@ -0,0 +1,59 @@
+export enum ColumnTypes {
```
onlyjackfrost, line 1: nit) can we rename to WrenEngineColumnType?

onlyjackfrost (approved):
LGTM

PR #216, opened 2024-05-08
fix(wren-ui): Align column types with backend connectors provided types

andreashimin:
## Description

- align column types
- update id, binary icon 
   image
   image

- update calculated field functions type checking


wwwy3y3 (approved):
LGTM on types

fredalai (approved):

PR #218, opened 2024-05-08
feature(wren-ai-service): Indexing saved view into document store

paopa:
This PR is designed to convert the saved view into the specified format and store it in the document store. Additionally, it lays the groundwork for future functionalities where we can perform searches to retrieve past questions and answers associated with the view (this functionality will be addressed in a subsequent PR). The PR also aims to enhance modularity by extracting functions into separate components and utilizing the Haystack pipeline to manage the entire workflow.

```json
{
	"question":"user original query",
	"description":"the description generated by LLM",
	"statement":"the SQL statement generated by LLM"
}
```

## The flow chart for the indexing process
indexing_pipeline


wren-ai-service/src/pipelines/ask/indexing_pipeline.py
```diff
@@ -23,46 +23,112 @@
...
+@component
+class DocumentCleaner:
+    """
+    This component is used to clear all the documents in the specified document store(si).
+
+    """
+
+    def __init__(self, stores: List[DocumentStore]) -> None:
+        self._stores = stores
+        pass
```
cyyeh, line 35: I am curious what's the purpose of "pass" here

wren-ai-service/src/pipelines/ask/indexing_pipeline.py
```diff
@@ -23,46 +23,112 @@
...
+
+    @component.output_types(mdl=str)
+    def run(self, mdl: str) -> str:
+        def _clear_documents(store: DocumentStore) -> None:
+            ids = [str(i) for i in range(store.count_documents())]
+            if ids:
+                store.delete_documents(ids)
+
+        logger.info("Ask Indexing pipeline is clearing old documents...")
+        [_clear_documents(store) for store in self._stores]
```
cyyeh, line 45: I think we should not directly delete all document in all stores here.

wren-ai-service/src/pipelines/ask/indexing_pipeline.py
```diff
@@ -134,9 +200,39 @@ def _format(view: Dict[str, Any]) -> str:
...
+    def __init__(
+        self,
+        document_store: DocumentStore,
+        view_store: DocumentStore,
+    ) -> None:
+        pipe = Pipeline()
+        pipe.add_component("cleaner", DocumentCleaner([document_store, view_store]))
+        pipe.add_component("view_indexer", ViewIndexer(view_store))
+        pipe.add_component("converter", MDLToDDLConverter())
+        pipe.add_component(
```
cyyeh, line 213: due to consistency, should we also have a writer for view_store here instead of embedding it inside ViewIndexer?

maybe the writer for document_store(should be renamed to ddl_store?) is called ddl_writer, and the writer for view_store is view_writer?

wren-ai-service/src/pipelines/ask/indexing_pipeline.py
paopa: no more purpose. this is just that I use to put a pass before implementing a function. I can remove it. 👍

wren-ai-service/src/pipelines/ask/indexing_pipeline.py
cyyeh: after discussion, there is no issue deleting all documents in all stores. since we assume mdl should contain the latest spec.

paopa: after the discussion, we won't modfiy the behavior here.

wren-ai-service/src/pipelines/ask/indexing_pipeline.py
paopa: after discussion, we'll split the view indexer into two components and rename these components for better understanding.

cyyeh (approved):
LGTM

PR #228, opened 2024-05-09
Feature(wren-ui): adjust display name validation logic of Calculated field & View 

onlyjackfrost:
Adjust logic:
 - use input displayName to generate referenceName
 - validation target is input displayName

### Example
created view 
image

created calculated field
image


wwwy3y3 (approved):
besides what @andreashimin said, others LGTM.

PR #230, opened 2024-05-10
chore(wren-ui): Replace view, calculated field & relationship name to alias

andreashimin:
## Description
To replace view, calculated field & relationship reference name to alias name

## Tasks
- [x] Replace view referenceName with displayName
- [x] Replace relationship referenceName to displayName
- [x] Provide relationship model & column displayName to client (BE)
- [x] Replace calculated field referenceName to displayName


wwwy3y3 (approved):
LGTM on backend

onlyjackfrost (approved):
LGTM for the Backend part

PR #237, opened 2024-05-13
add logger.error messages

cyyeh:
- added logger.error to dump error messages to terminal in each ai pipeline for every failed case
- fixed view indexing issue


paopa (approved):
LGTM.

PR #244, opened 2024-05-13
allow users to change openai model in launcher

cyyeh:
before choosing generation model:
image

after choosing generation model:
image

wren-ai-service console will show the generation model user chooses:
image


wwwy3y3 (approved):
Others LGTM

PR #246, opened 2024-05-14
fix(wren-ui): Fix Posthog related env in client

andreashimin:
## Description

By design, Nextjs client side cannot receive runtime env, the `NEXT_PUBLIC` env will be converted to frozen value, so we need to provide env at the build time.

> If you build and deploy a single Docker image to multiple environments, all NEXT_PUBLIC_ variables will be frozen with the value evaluated at build time, so these values need to be set appropriately when the project is built.

As an alternative, add an API to retrieve the user option from the backend such as telemetry enabling


wwwy3y3 (approved):
LGTM on backend

onlyjackfrost (approved):
LGTM as well

PR #250, opened 2024-05-14
chore(wren-ui): Tracking page view when router change

andreashimin:
## Description
Add tracking page view when router changing


fredalai (approved):

wwwy3y3 (approved):
LGTM

PR #257, opened 2024-05-15
feature(wren-ai-service): support to return a saved view as a Candidate

paopa:
This PR aims to complete the changes to the ask result API, which now returns the candidate's type to indicate the source of the candidate. Additionally, it introduces a mechanism to retrieve historical questions of views as candidates. Lastly, it adds an exclude section to ensure diversity in the LLM-generated SQL statements.

# Screenshots

image
image


wren-ai-service/src/pipelines/ask/historical_question.py
```diff
@@ -0,0 +1,90 @@
...
+class OutputFormatter:
+    @component.output_types(
+        documents=List[Optional[Dict]],
+    )
+    def run(self, documents: List[Document]):
+        list = []
+        logger.debug(f"historical_question_output_formatter: {documents}")
+
+        for doc in documents:
+            content = eval(doc.content)
```
cyyeh, line 75: I am not sure the reason to use eval here

cyyeh (commented):
I have few questions:
1. how do we ensure to generate at most 3 candidates to users(including saved view)?
2. does ai service need to do anything once a user clicks the saved view after asking a question?

wren-ai-service/src/pipelines/ask/historical_question.py
paopa: after a discussion, we will use other function instead of python eval function to convert str to dict.

cyyeh (approved):
LGTM

PR #260, opened 2024-05-16
chore(wren-ui): Replace model preview data column name to alias

andreashimin:
## Description
The model preview data should show the alias as column name.


fredalai (approved):
LGTM

PR #264, opened 2024-05-16
chore(wren-ui): Show AI exception errors to client when deploying manifest

andreashimin:
## Description
Show deploying error toast, and remind the user to check the detail in logs.

## UI Screenshot
image

### DIFF
--- a/wren-ui/src/apollo/server/adaptors/wrenAIAdaptor.ts
+++ b/wren-ui/src/apollo/server/adaptors/wrenAIAdaptor.ts
@@ -237,11 +237,11 @@ export class WrenAIAdaptor implements IWrenAIAdaptor {
       }
     } catch (err: any) {
       logger.debug(
-        `Got error when deploying to wren AI, hash: ${hash}. Error: ${getAISerciceError(err)}`,
+        `Got error when deploying to wren AI, hash: ${hash}. Error: ${err}`,
       );
       return {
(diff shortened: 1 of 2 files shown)


fredalai (approved):
LGTM

wren-ui/src/apollo/server/adaptors/wrenAIAdaptor.ts
```diff
@@ -282,12 +278,13 @@ export class WrenAIAdaptor implements IWrenAIAdaptor {
         `${this.wrenAIBaseEndpoint}/v1/semantics-preparations/${deployId}/status`,
       );
       if (res.data.error) {
-        logger.debug(`deploy error: ${res.data.error}`);
+        // passing AI response error string to catch block
+        throw res.data.error;
```
wwwy3y3, line 282: If `res.data.error` is not an error object. Could we wrap it with `Error` ? Maybe we can use our `Errors.create` function ?

andreashimin: Maybe wrap it with `Error`? 
The error will be retrieved in the response body as the structure in the image from the deploy API.
image

Using `Errors.create` function currently seems unnecessary here.

wwwy3y3: sure

wwwy3y3 (approved):
LGTM

PR #265, opened 2024-05-17
Chore/ai service/minor update

cyyeh:
- [orjson](https://github.com/ijl/orjson) is a lot faster than python's native json
- fix python dependency version


paopa (approved):
LGTM

PR #267, opened 2024-05-17
feature(wren-ai-service): Automatically loading the providers

paopa:
This PR aims to simplify support for providers (such as LLM providers or Document Store providers) for both users and ourselves, enhancing the module's scalability. The following section will outline the steps to customize a new provider.

## Adding your preferred LLM or Document Store

1. Create a provider module under the llm or document_store package. The file structure should look like this:
```text
src
├── providers
│   ├── llm
│   └── document_store
```

2. Create a class that inherits from `LLMProvider` or `DocumentStoreProvider` and implement the necessary methods. Detailed information about the base classes can be found in the `core.provider` module.

image

3. Use the provider decorator (import it from the src.providers.loader module) and assign a name to your provider.

image

4. In the last step, update the env file to set the LLM_PROVIDER or DOCUMENT_STORE_PROVIDER variable to your specified provider. Additionally, add any custom variables for your provider into the env file.

image


wren-ai-service/src/providers/llm/openai.py
```diff
@@ -34,8 +37,22 @@ def run(self, prompt: str, generation_kwargs: Optional[Dict[str, Any]] = None):
...
+@provider("openai")
 class OpenAILLMProvider(LLMProvider):
-    def __init__(self, api_key: Secret, generation_model: str = GENERATION_MODEL_NAME):
+    def __init__(
+        self,
+        api_key: Secret = Secret.from_env_var("OPENAI_API_KEY"),
+        generation_model: str = os.getenv("OPENAI_GENERATION_MODEL")
+        or GENERATION_MODEL_NAME,
+    ):
+        def _verify_env_vars() -> None:
```
cyyeh, line 48: I suggest we still explicitly pass open ai key inside the OpenAI constructor.

cyyeh (commented):
a comment, others LGTM. thanks for your hard work!

cyyeh:
also for the documentation, I think users also need to add provider specific environment variables there?

cyyeh (approved):

Issue #193, opened 2024-05-06
failed to deploy model if I apply the auto-generated table relationships


cyyeh:
**Describe the bug**
I used the [instructions here to dump pagila database to my local postgresql db](https://neon.tech/docs/import/import-sample-data#pagila-database). If I skip the relationship connection step, the model is deployed successfully; however, the model is failed to deploy if I apply the auto-generated table relationships

error message:
```json
{
    "errors": [
        {
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "saveRelations"
            ],
[...]

PR #268, opened 2024-05-17
feat(wren-ui): Update relationship rule

fredalai:
## Description
 Based on the [issue #193](https://github.com/Canner/WrenAI/issues/193), Considering that there will be multiple relationships between two models in most usage scenarios, we decided to allow multiple relationships at this stage.

## Tasks
 - [x] [API] Update logic for findExistedRelationBetweenModels function
 - [x] [Relationship Modal UI] update relationship modal UI
   - Rules that do not allow the creation of new relationships:
      - Same as from and to of existing relationship
      - Reverse of from and to of existing relationship

### DIFF
--- a/wren-ui/src/apollo/server/repositories/relationshipRepository.ts
+++ b/wren-ui/src/apollo/server/repositories/relationshipRepository.ts
@@ -52,7 +52,6 @@ export interface IRelationRepository extends IBasicRepository<Relation> {
     },
     queryOptions?: IQueryOptions,
   ): Promise<RelationInfo[]>;
-  findExistedRelationBetweenModels(modelIds): Promise<RelationInfo[]>;
 }
 
 export class RelationRepository
@@ -185,27 +184,6 @@ export class RelationRepository
       'tmc.reference_name AS toColumnName',
       'tmc.display_name AS toColumnDisplayName',
     );
-    return result.map((r) => this.transformFromDBData(r)) as RelationInfo[];
-  }
-
-  public async findExistedRelationBetweenModels(modelIds) {
-    const query = this.knex(this.tableName)
-      .join(
-        'model_column AS fmc',
-        `${this.tableName}.from_column_id`,
-        '=',
-        'fmc.id',
-      )
-      .join(
-        'model_column AS tmc',
-        `${this.tableName}.to_column_id`,
-        '=',
-        'tmc.id',
-      )
-      .whereIn('fmc.model_id', modelIds)
-      .whereIn('tmc.model_id', modelIds)
-      .select(`${this.tableName}.*`);
-    const result = await query;
     return result.map((r) => this.transformFromDBData(r)) as RelationInfo[];
   }
 }
--- a/wren-ui/src/apollo/server/repositories/relationshipRepository.ts
+++ b/wren-ui/src/apollo/server/repositories/relationshipRepository.ts
@@ -4,6 +4,7 @@ import {
   IBasicRepository,
   IQueryOptions,
 } from './baseRepository';
+import { RelationData } from '../types';
 
 export interface Relation {
   id: number; // ID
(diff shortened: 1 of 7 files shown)


wren-ui/src/apollo/server/repositories/relationshipRepository.ts
```diff
@@ -187,25 +186,4 @@ export class RelationRepository
     );
     return result.map((r) => this.transformFromDBData(r)) as RelationInfo[];
   }
-
-  public async findExistedRelationBetweenModels(modelIds) {
```
onlyjackfrost, line 191: I think this should be rewrite instead of deleted.
Can rewrite to check if columnIds exist in the table.

onlyjackfrost (commented):
The validation logic in the client side LGTM

wren-ui/src/utils/validator/relationshipValidator.ts
```diff
@@ -0,0 +1,113 @@
...
+    relationships[formValues.fromField.modelName];
+  const isDuplicate = relationshipsByFromFieldModel.find(
+    (relationship) =>
+      relationship.fromField.modelId === formValues.fromField.modelId &&
+      relationship.fromField.fieldId === formValues.fromField.fieldId &&
+      relationship.toField.modelId === formValues.toField.modelId &&
+      relationship.toField.fieldId === formValues.toField.fieldId,
+  );
+
+  if (Boolean(isDuplicate)) return true;
```
andreashimin, line 36: I will assume isDuplicate is already a boolean value. I think the type conversion should be done beforehand.

wren-ui/src/utils/validator/relationshipValidator.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (!skip) {
+      const toField = getFieldValue(FormFieldKey.TO_FIELD);
+      if (toField && toField.model && toField.field) {
+        if (
+          isExistRelationship(
+            relationships,
+            convertFormValuesToIdentifier({
+              fromField: value,
+              toField,
+              type: '',
```
andreashimin, line 71: Why does the type need to pass an empty string? 🤔

wren-ui/src/utils/validator/relationshipValidator.ts
```diff
@@ -0,0 +1,113 @@
...
+      relationship.fromField.modelId === formValues.toField.modelId &&
+      relationship.fromField.fieldId === formValues.toField.fieldId &&
+      relationship.toField.modelId === formValues.fromField.modelId &&
+      relationship.toField.fieldId === formValues.fromField.fieldId,
+  );
+
+  return Boolean(isReverseDuplicate);
+};
+
+export const relationshipFromFieldValidator =
```
andreashimin, line 51: Should it be unified with other validators' names?
```suggestion
export const createRelationshipFromFieldValidator =
```

wren-ui/src/apollo/server/repositories/relationshipRepository.ts
fredalai: Hmm... I rewrote the checking logic to check if it is a duplicate relationship

wren-ui/src/utils/validator/relationshipValidator.ts
```diff
@@ -0,0 +1,113 @@
...
+        ) {
+          return Promise.reject(ERROR_TEXTS.ADD_RELATION.RELATIONSHIP.EXIST);
+        }
+      }
+    }
+
+    return Promise.resolve();
+  };
+
+export const relationshipToFieldValidator =
```
andreashimin, line 83: Should it be unified with other validators' names?
```suggestion
export const createRelationshipToFieldValidator =
```

wren-ui/src/utils/validator/relationshipValidator.ts
fredalai: Hmm... Oh, the type is unused props for judgment.
I can  pass the type field value to reduce any confusion

fredalai: I will remove it

andreashimin (approved):
LGTM on FE

PR #280, opened 2024-05-21
feature(wren-launcher): enable dynamic-fields config for Wren Engine

wwwy3y3:
Enable dynamic-fields config for Wren Engine


goldmedal (approved):
LGTM

PR #288, opened 2024-05-21
fix(wren-ui): Exclude self-referential relationship in auto-generated relationship

andreashimin:
## Description 

Self-Referential Relationship: In a self-referential relationship, a column within a model refers back to the same model. For example, consider an `Employee` model where the `reportsTo` column holds the `ID` of another `Employee`, indicating who each employee reports to. This means that the `Employee` model is linked to itself through the `reportsTo` column. 

**Currently, WrenAI does not support creating such relationships where a column in a model references the same model**. 

## Fix
- Exclude self-referential relationship in auto-generated relationship
- Skip the edge in modeling diagram if model not existed

### DIFF
--- a/wren-ui/src/apollo/server/resolvers/projectResolver.ts
+++ b/wren-ui/src/apollo/server/resolvers/projectResolver.ts
@@ -269,7 +269,12 @@ export class ProjectResolver {
       id,
       displayName,
       referenceName,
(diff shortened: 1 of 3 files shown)


wwwy3y3 (approved):
backend LGTM

fredalai (approved):
LGTM for FE

PR #309, opened 2024-05-23
make pipelines cleaner

cyyeh:
- simplify codebase
  - pass `llm_provider` or `document_store_provider` to pipelines
  - put each pipeline's non-shared component to each pipeline
- update dependencies
  - upgrade Haystack to  v2.1.2 and include the argument `include_outputs_from` to each pipeline's run method for easily debugging: https://github.com/deepset-ai/haystack/releases/tag/v2.1.0


wren-ai-service/src/web/v1/services/ask.py
```diff
@@ -116,6 +116,7 @@ def __init__(self, pipelines: dict[str, Pipeline]):
 
     def prepare_semantics(self, prepare_semantics_request: SemanticsPreparationRequest):
         try:
+            print(prepare_semantics_request.mdl)
```
paopa, line 119: using the logger instead of the print function will be better, I think.

wren-ai-service/src/web/v1/services/ask_details.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, pipelines: dict[str, Pipeline]):
     def ask_details(
         self,
         ask_details_request: AskDetailsRequest,
-    ) -> AskDetailsResponse:
+    ):
```
paopa, line 63: curious, why remove this returning declaration?

paopa (commented):
I have a question about the include_outputs_from parameter that I noticed being added to all pipelines. It doesn't seem to be used in any case. Why is it necessary to include it?

paopa:
btw, if it is possible, adding some description to mention the changes for the PR would be better. 😄

cyyeh:
> btw, if it is possible, adding some description to mention the changes for the PR would be better. 😄

sorry, forgot to add the descriptions. They're added now

paopa (approved):
LGTM.

PR #314, opened 2024-05-24
cb interface

pberton:


goldmedal:
Hi @pberton,

I'm so glad to see that you want to contribute to the Couchbase-related feature. I have some suggestions and thoughts on this issue.

Since WrenAI is based on SQL for modeling, we're working on using the [ibis-project](https://github.com/ibis-project/ibis) to connect to the datasource. The concept is that WrenAI generates ibis-executable SQL, which is then transformed to the specific datasource using the appropriate dialect. You can see how we invoke the ibis-project in the [Wren Engine repo branch](https://github.com/Canner/wren-engine/tree/feature/ibis-server). It's still a work in progress.

As far as I know, Ibis doesn't have a dialect for Couchbase yet. So, if you could contribute this connector to Ibis, it would make it easier for WrenAI to integrate with Couchbase in the future.

PR #341, opened 2024-05-31
Chore: refactor database schema to reflect how we use them.

onlyjackfrost:
- wrap data source connection related columns(host, port, ...) with a new column "connection_info"
- use `tableReference` in MDL manifest instead of refSql (ref: Wren Engine [PR](https://github.com/Canner/wren-engine/pull/491)). This will benefit performance for small queries and no longer have to deal with SQL format.


wwwy3y3 (approved):
LGTM

PR #342, opened 2024-05-31
Fix(wren-ui): present view displayName instead of name in home page 

onlyjackfrost:
present view displayName instead of name in home page


fredalai (approved):
LGTM for FE

wwwy3y3 (approved):
LGTM

PR #362, opened 2024-06-01
update sqlparse version

cyyeh:
- update sqlparse version to 0.5.0 to solve the security issue: [sqlparse parsing heavily nested list leads to Denial of Service](https://github.com/Canner/WrenAI/security/dependabot/6)
- upload poetry.lock, otherwise security check would be failed


paopa (approved):
LGTM

PR #363, opened 2024-06-02
Feature/ai service/refactor pipeline using hamilton

cyyeh:
- rewrote all pipelines without document store or embedder using Hamilton style
- added a TODO comment in the semantics_description pipeline

PR #367, opened 2024-06-03
feature(wren-ai-service): refactor pipelines of retrieval and historical question

paopa:
After #363, we refactored almost all pipelines and completed the remaining ones in this PR. We also corrected some errors to ensure all functions work as they did before the refactoring.


cyyeh (approved):
LGTM

PR #373, opened 2024-06-04
feat(wren-ui): Provide limit property in preview data API for excel-add-in used

andreashimin:
## Description
- Add limit property for preview data
- Fix all view related APIs that need to provide displayName in the `ViewInfo` type


onlyjackfrost (approved):
LGTM

PR #374, opened 2024-06-04
feature(wren-ai-service): support async SQL dry-run and async Qdrant component 

paopa:
This PR aims to make the SQL dry-run and Qdrant components support async tasks. and also extract the function related to the engine to a module.

- https://python-client.qdrant.tech/qdrant_client.async_qdrant_client
- https://docs.aiohttp.org/en/stable/client_quickstart.html#make-a-request

### DIFF
--- a/wren-ai-service/pyproject.toml
+++ b/wren-ai-service/pyproject.toml
@@ -21,6 +21,7 @@ sqlparse = "==0.5.0"
 sqlglot = "==22.5.0"
 orjson = "==3.10.3"
 sf-hamilton = {version = "1.63.0", extras = ["visualization"]}
+aiohttp = "3.9.5"
 
 [tool.poetry.group.dev.dependencies]
 pytest = "==8.2.0"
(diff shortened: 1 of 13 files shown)


wren-ai-service/pyproject.toml
```diff
@@ -21,6 +21,7 @@ sqlparse = "==0.5.0"
 sqlglot = "==22.5.0"
 orjson = "==3.10.3"
 sf-hamilton = {version = "1.63.0", extras = ["visualization"]}
+aiohttp = "3.9.5"
```
cyyeh, line 24: should be `aiohttp = "==3.9.5"`?

wren-ai-service/src/engine.py
cyyeh, line 1: nice abstraction

cyyeh (commented):
Overall LGTM, only minor issue

wren-ai-service/pyproject.toml
cyyeh: also `sf-hamilton = {version = "==1.63.0", extras = ["visualization"]}`?

paopa: according to https://python-poetry.org/docs/dependency-specification/#exact-requirements, both of them are the way to specify the exact version for a package. I think we could unify using `==` to specify the version. I'll modify them in the next commit. 👍

cyyeh (approved):
LGTM

PR #377, opened 2024-06-04
Chore/ai service/minor updates

cyyeh:
- restructure tests folder
- fix `WREN_ENGINE_ENDPOINT` empty issue
- add `async_timer`
- add `WORKER` as env for setting up number of workers, default is 1(also enabling reloading)


wren-ai-service/src/utils.py
```diff
@@ -101,8 +106,38 @@ def wrapper_timer(*args, **kwargs):
...
-            return value
+            return result
 
         return func(*args, **kwargs)
 
     return wrapper_timer
+
+
+def async_timer(func):
+    load_env_vars()
```
paopa, line 117: removing invoking the loading env vars function can reduce the impact on the function with this decorator.

wren-ai-service/src/utils.py
```diff
@@ -101,8 +106,38 @@ def wrapper_timer(*args, **kwargs):
...
+            endTime = time.perf_counter()
+            elapsed_time = endTime - startTime
+
+            logger.info(
+                f"{func.__qualname__} Elapsed time: {elapsed_time:0.4f} seconds"
+            )
+
+            return result
+
+        return await process(func, *args, **kwargs)
```
paopa, line 141: the else-case in the process function will cause an error if it is not a coroutine, the python interpreter will throw the error like 
```
TypeError: object int can't be used in 'await' expression
```

wren-ai-service/src/__main__.py
```diff
@@ -67,6 +69,20 @@ def health():
...
+    """
+    Dummy endpoint to test async behavior by sleeping for several seconds
+    """
+    if should_sleep:
+        if is_async:
+            await asyncio.sleep(sleep)
+        else:
+            time.sleep(sleep)
+
+    return {"status": "dummy"}
```
paopa, line 83: note: we will move this to a development router in the upcoming PR

paopa (approved):

PR #379, opened 2024-06-05
Fix: fix sql not show after clicking "View Full SQL" button

onlyjackfrost:
fix sql not show after clicking "View Full SQL" button


fredalai (approved):
LGTM

andreashimin (approved):
LGTM

PR #382, opened 2024-06-05
Chore/ai service/minor update

cyyeh:
- update missing configs
- add redis to support multi-worker setup
- fix document store duplicate collection issue by adding a pre_run script
- add dummy_ask api


wren-ai-service/src/web/development.py
```diff
@@ -1,20 +1,128 @@
...
+    REDIS_DB.hset(
+        "test_ask_results",
+        ask_request.query_id,
+        AskResultResponse(
+            status="finished",
+        ).model_dump_json(),
+    )
+
+
+def get_dummy_ask_task_result(
```
paopa, line 71: it seems only be used in get_dummy_ask_result, I think we can merge them into the same one.

paopa (approved):
only have a minor comment, others LGTM.

PR #386, opened 2024-06-07
small refactor and upgrade uvicorn version to allow process manager

cyyeh:
- small refactor to `src/__main__.py`
- upgrade uvicorn version so we don't need gunicorn in the future: https://www.uvicorn.org/release-notes/#0300-2024-05-28


paopa (approved):
LGTM

PR #392, opened 2024-06-10
refine readme

cyyeh:
- update instructions for running wren-ai-service for development
- simplify demo app setup
- update Poetry version
  - readme
  - ci
- minor performance improvement(reference: https://www.youtube.com/watch?v=7jtzjovKQ8A)
  - uvicorn
    - explicitly use `uvloop`
    - explicitly use `httptools`
  - fastapi
    - use `ORJSONResponse` instead of `JSONResponse`


wren-ai-service/demo/Makefile
paopa, line 1: Would you be able to move the run command to the primary make file? Almost all commands here are the same in the primary make file. thus, I think we can consider moving the one to start the AI demo-site to there and delete this file.

paopa (commented):
Just curious do we have the chart or table to compare the original performance and this one with Uvloop and Orjson response

cyyeh:
> Just curious do we have the chart or table to compare the original performance and this one with Uvloop and Orjson response

by default uvicorn used "uvloop" and "httptools"; however, I just want to make it more explicit here :)

for orjson, it's not easier to see the difference, since the payload we are using is not big enough; however, I think we still should use orjson if it's already available :)

paopa (approved):
LGTM. thanks for your work! after the CI done, we could merge this PR.

PR #400, opened 2024-06-12
feature(ai-service): wren-ui dry-run SQL API instead of wren-engine API

paopa:
As the title suggests, this PR uses the Wren-UI endpoint to dry-run the SQL for validation. This PR only replaces the endpoint, but there are still some details that need to be fixed, such as the CI and the development tool, which also need to start the UI and Ibis server. We will create a follow-up PR to address these issues.


cyyeh (commented):
Would u like to add wren ui endpoint env to the docker-compose.yaml in the docker folder from the root directory?

cyyeh (approved):
LGTM

PR #440, opened 2024-06-24
feature(wren-ai-service): implement Engine Adapter to validate queries

paopa:


wren-ai-service/src/providers/engine/wren.py
```diff
@@ -0,0 +1,68 @@
...
+            res = await response.json()
+            if res.get("data"):
+                return True, None
+            return False, res.get("errors", [{}])[0].get("message", "Unknown error")
+
+
+@provider("wren-ibis")
+class WrenIbis(Engine):
+    def __init__(self, endpoint: str = os.getenv("WREN_IBIS_ENDPOINT")):
+        self._endpoint = endpoint
```
cyyeh, line 46: I suggest we put `source`, `manifest`, `connection info` parameters to the `dry_run_sql` with default value from environment variable, so it's more flexible for users to decide what source to use in runtime

```python
    def __init__(self, endpoint: str = os.getenv("WREN_IBIS_ENDPOINT")):
        self._endpoint = endpoint

    async def dry_run_sql(
        self,
        sql: str,
        session: aiohttp.ClientSession,
        source: str = os.getenv("WREN_IBIS_SOURCE"),
        manifest: str = os.getenv("WREN_IBIS_MANIFEST"),
        connection_info: str = os.getenv("WREN_IBIS_CONNECTION_INFO"),
    ) -> Tuple[bool, Optional[Dict[str, Any]]]:
```

cyyeh (commented):
only one minor comment

wren-ai-service/src/providers/engine/wren.py
paopa: sure. but a bit different, using a dict and also adding to the abstract method to consist the interface will be better, I do believe. the code will be like:
```python
## engine
    @abstractmethod
    async def dry_run_sql(
        self,
        sql: str,
        session: aiohttp.ClientSession,
        properties: Dict[str, Any] = {},
    ) -> Tuple[bool, Optional[Dict[str, Any]]]:
        ...


## wren-ibis
    async def dry_run_sql(
        self,
        sql: str,
        session: aiohttp.ClientSession,
        properties: Dict[str, Any] = {
            "source": os.getenv("WREN_IBIS_SOURCE"),
            "manifest": os.getenv("WREN_IBIS_MANIFEST"),
            "connection_info": orjson.loads(
                os.getenv("WREN_IBIS_CONNECTION_INFO", "{}")
            ),
        },
    ) -> Tuple[bool, Optional[Dict[str, Any]]]:

```

thanks for the suggestion!

cyyeh (approved):
LGTM

PR #463, opened 2024-06-30
remove query understanding

cyyeh:
since the functionality of the "query understanding" pipeline is unneccesary(it's only aimed to refuse user's questions using random words) at the moment, and I found some models may generate incorrect responses at this pipeline. In the future, we may add it back with much better design for user's query clarification.


paopa (commented):
overall LGTM. only have a question about should we also remove the error code for the misleading query? 

https://github.com/Canner/WrenAI/blob/e97a9de73a87881b500b4a0269eef11ade5bc04a/wren-ai-service/src/web/v1/services/ask.py#L94-L98

cyyeh:
> overall LGTM. only have a question about should we also remove the error code for the misleading query? 
> 
> 
> 
> https://github.com/Canner/WrenAI/blob/e97a9de73a87881b500b4a0269eef11ade5bc04a/wren-ai-service/src/web/v1/services/ask.py#L94-L98

Ok, I will remove it. Thanks

paopa (approved):
LGTM

PR #493, opened 2024-07-08
minor update for ai-service

cyyeh:
- automatically remove trailing / for urls in env vars
- fix azure open ai llm issue
- fix auto redeploy issue

### DIFF
--- a/wren-ai-service/src/force_deploy.py
+++ b/wren-ai-service/src/force_deploy.py
@@ -1,13 +1,12 @@
 import asyncio
 import os
-from pathlib import Path
 
 import aiohttp
 import backoff
-from dotenv import load_dotenv
 
-if Path(".env.dev").exists():
-    load_dotenv(".env.dev", override=True)
+from src.utils import load_env_vars
+
+load_env_vars()
 
 
 @backoff.on_exception(backoff.expo, aiohttp.ClientError, max_time=60, max_tries=3)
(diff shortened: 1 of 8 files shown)


wren-ai-service/src/force_deploy.py
```diff
@@ -1,12 +1,12 @@
...
+from pathlib import Path
 
 import aiohttp
 import backoff
 from dotenv import load_dotenv
 
-load_dotenv(override=True)
-if is_dev_env := os.getenv("ENV") and os.getenv("ENV").lower() == "dev":
+if Path(".env.dev").exists():
     load_dotenv(".env.dev", override=True)
```
paopa, line 10: how about using utils.load_env_vars func to instead of this.

wren-ai-service/src/providers/embedder/azure_openai.py
paopa, line 1: I saw many place need checking and removing the slash at the end of an endpoint. I think we could extract them a func in utils module.
```py
def endpoint_formatter(endpoint: str) -> str:
    return endpoint.rstrip("/") if endpoint.endswith("/") else endpoint

```

paopa (commented):
Thanks for your hard work! I have left two comments with my thoughts.

wren-ai-service/src/force_deploy.py
cyyeh: done

wren-ai-service/src/providers/embedder/azure_openai.py
cyyeh: done, and the function name is `remove_trailing_slash`

paopa (approved):
LGTM

PR #395, opened 2024-06-11
feature(wren-ai-service): integrate Langfuse SDK to represent the evaluation result

paopa:
This PR aims to integrate the decorator-based Langfuse SDK to collect data for the evaluation process. By default, the environment setup is disabled. To enable Langfuse, follow these steps:

```env
# .env.dev file
# Langfuse configuration
LANGFUSE_ENABLE=True
LANGFUSE_SECRET_KEY=
LANGFUSE_PUBLIC_KEY=
LANGFUSE_HOST=
```

Fill in the environment variables. You can generate the secret key, public key, and host by following the instructions at [Langfuse Documentation](https://langfuse.com/docs/get-started#create-new-project-in-langfuse).

Currently, we don't set the session ID. When implementing the main process for the evaluation, refer to the following code to set up the session ID.
```python
@observe()
async def main(user_id: str):
    langfuse_context.update_current_trace(
        user_id=user_id, 
        session_id=f"{user_id}_{uuid.uuid4()}"
    )
    return await story()


async def run():
    await asyncio.gather(main("foo"), main("bar"))
    langfuse_context.flush()


asyncio.run(run())
```

## Screenshots
image

image

image

image

image

PR #495, opened 2024-07-09
feature(wren-ai-service): integrate Langfuse SDK to represent the evaluation result (#395)

paopa:
Refer to #395, we believe this feature won't impact the current function for asking and other operations. Therefore, let's merge it into the main branch to avoid frequent conflict resolution.


wren-ai-service/src/pipelines/ask/followup_generation.py
cyyeh, line 1: how about we put the import of `langfuse_context` and `init_langfuse` to the __main__ function instead? We can follow the same pattern to other files as well

cyyeh (approved):
LGTM

PR #503, opened 2024-07-11
Fix(wren-ui): fix unit test and add some test cases for ibis adapter

onlyjackfrost:
fix import error and add some test cases for ibis server


fredalai (approved):
LGTM

andreashimin (approved):

PR #517, opened 2024-07-15
fix(wren-ai-service): modify the env ibis service in the dev env example

paopa:
This PR aims to fix an issue with the following format. The string after the equal sign is used as the value for the environment variable, causing a problem where Python imports the module and assumes the default value, regardless of whether the class is used. As a result, the connection info environment variable would be decoded and converted to JSON, leading to an error.

```
WREN_IBIS_MANIFEST= # this is a base64 encoded string of the MDL
WREN_IBIS_CONNECTION_INFO= # this is a base64 encode string of the connection info
```


cyyeh (commented):
I suggest LANGFUSE_ENABLE is false by default, and the LOGGING_LEVLE is DEBUG by default. What do you think?

paopa:
> I suggest LANGFUSE_ENABLE is false by default, and the LOGGING_LEVLE is DEBUG by default. What do you think?

it works to me. thanks for the suggestion.

cyyeh (approved):
LGTM

PR #541, opened 2024-07-18
feat(wren-ai-service): add gpt-4o-mini

cyyeh:


.github/workflows/ai-service-test.yaml
```diff
@@ -52,7 +52,7 @@ jobs:
           DOCUMENT_STORE_PROVIDER: qdrant
           LLM_OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
           EMBEDDER_OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
-          GENERATION_MODEL: gpt-3.5-turbo
+          GENERATION_MODEL: gpt-4o-mini
```
paopa, line 55: I'm not sure if we need to change the model to 4o-mini as well. What do you think?

paopa (approved):

.github/workflows/ai-service-test.yaml
paopa: discussed offline, the pricing of 4o-mini is cheaper than 3.5-turbo and the performance is much better, thus we use it instead.

Issue #520, opened 2024-07-16
Got incorrect SQL syntax when switching to database SQL


wwwy3y3:
**Describe the bug**
Got incorrect SQL syntax when switching to database SQL

**To Reproduce**
1. use BigQuery and ask a question
2. click show SQL
3. switch to use database SQL
4. copy to bigquery and run, it will fail

**Expected behavior**
The query should be successfully executed.

**Screenshots**
I use tpch data to ask a question ("list my customers") and copy the SQL from the answer:
![Screenshot 2024-07-16 at 10 49 17 AM](https://github.com/user-attachments/assets/b72c924d-acc9-4e70-9d53-ed6963ff87d2)

Copy to bigquery and run:
![Screenshot 2024-07-16 at 10 49 11 AM](https://github.com/user-attachments/assets/39510920-c64e-494d-9160-66a4e1ac3020)

Generated BigQuery SQL attached below:
```sql
WITH
  "customer" AS (
   SELECT
     "customer"."c_acctbal" "c_acctbal"
   , "customer"."c_address" "c_address"
   , "customer"."c_comment" "c_comment"
   , "customer"."c_custkey" "c_custkey"
   , "customer"."c_mktsegment" "c_mktsegment"
   , "customer"."c_name" "c_name"
   , "customer"."c_nationkey" "c_nationkey"
   , "customer"."c_phone" "c_phone"
   FROM
     (
      SELECT
        "customer"."c_acctbal" "c_acctbal"
      , "customer"."c_address" "c_address"
      , "customer"."c_comment" "c_comment"
      , "customer"."c_custkey" "c_custkey"
      , "customer"."c_mktsegment" "c_mktsegment"
      , "customer"."c_name" "c_name"
      , "customer"."c_nationkey" "c_nationkey"
      , "customer"."c_phone" "c_phone"
      FROM
        (
         SELECT
           "c_acctbal" "c_acctbal"
         , "c_address" "c_address"
         , "c_comment" "c_comment"
         , "c_custkey" "c_custkey"
         , "c_mktsegment" "c_mktsegment"
         , "c_name" "c_name"
         , "c_nationkey" "c_nationkey"
         , "c_phone" "c_phone"
         FROM
           "wrenai"."tpch_tiny"."customer" "customer"
      )  "customer"
   )  "customer"
) 
SELECT "c_name" "customer_name"
FROM
  "customer"
```

**Container Logs **
You can execute the following command to get the logs of containers and provide them here:
```
[...]

PR #550, opened 2024-07-24
Fix(wren-ui): Fix getting incorrect native sql issue

onlyjackfrost:
Fix issue https://github.com/Canner/WrenAI/issues/520 

We previously used Wren Engine to do dry-plan. Now, when the data source is not DuckDB, we use ibis server for dry-plan.


wwwy3y3 (approved):

wwwy3y3:
Kindly add some notes on what changed and why

PR #491, opened 2024-07-08
Multiple table selection enabled after search. #396

himanshu634:
I have fixed the issue of multiple table selection.

PR #551, opened 2024-07-24
fix(wren-ui): Fix select models duplicated keys issue

andreashimin:
## Description
In #491 using onChange to handle select all & deselect all click,
the if condition didn't bypass select one click successfully (it happens when the `selectedKeys` already had 1 key), causing onSelect & onChange to update `selectedKeys` at the same time, so there are duplicated keys inside.

image


fredalai (approved):
LGTM

PR #556, opened 2024-07-25
Chore: adjust telemetry and logger

onlyjackfrost:
- refine telemetry: rename event names and add postfix with success or not.
- hide dry-run error log due to happening too many times when ai service is doing SQL validation.


wwwy3y3 (approved):
LGTM, just don't send error object in logs. Too large and may contain sensitive information we don't want.

PR #558, opened 2024-07-25
feat(wren-ai-service): optimize qdrant for multi users setting

cyyeh:
- add [binary quantization](https://qdrant.tech/documentation/guides/quantization/?q=binary)
- allow [multitenancy](https://qdrant.tech/documentation/guides/multiple-partitions/)
- offload vectors to disk

### DIFF
--- a/wren-ai-service/src/pipelines/indexing/indexing.py
+++ b/wren-ai-service/src/pipelines/indexing/indexing.py
@@ -36,7 +36,7 @@ class DocumentCleaner:
         self._stores = stores
 
     @component.output_types(mdl=str)
-    async def run(self, mdl: str, user_id: str) -> str:
+    async def run(self, mdl: str, user_id: str = "") -> str:
         async def _clear_documents(store: DocumentStore, user_id: str) -> None:
             filters = (
                 {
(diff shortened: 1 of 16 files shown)


wren-ai-service/src/providers/document_store/qdrant.py
```diff
@@ -291,9 +355,26 @@ def get_store(
...
             index=dataset_name or "Document",
             recreate_index=recreate_index,
+            on_disk=True,
+            quantization_config=(
+                rest.BinaryQuantization(
+                    binary=rest.BinaryQuantizationConfig(
+                        always_ram=True,
+                    )
+                )
+                if embedding_model_dim >= 1024
```
paopa, line 369: I think we can mention why the setting enable if the dimension more than 1024.

wren-ai-service/src/pipelines/indexing/indexing.py
```diff
@@ -452,6 +464,7 @@ def visualize(self, mdl_str: str) -> None:
             output_file_path=f"{destination}/indexing.dot",
             inputs={
                 "mdl_str": mdl_str,
+                "id": id or "",
```
paopa, line 467: I'm curious what will it happen if we don't handle with the id is None.

paopa (commented):
overall LGTM. only have two questions for it.

wren-ai-service/src/pipelines/indexing/indexing.py
cyyeh: I've refactored the code

paopa: I noticed that all the `id` parameters have been handled properly. I think we can also remove the empty string.

paopa (approved):
LGTM👍

PR #561, opened 2024-07-26
fix(wren-ai-service): error handling for evaluation process

paopa:
This PR aims to add error handling to the evaluation process to prevent errors from affecting the results and also add a condition to avoid the error for Accuracy Metric. Additionally, it introduces a "failed count" metadata field into the summary trace, which will help us easily track the number of successful and failed traces.


cyyeh (approved):
LGTM

PR #589, opened 2024-08-09
chore(wren-ai-service): add script for mdl properties preparation

cyyeh:
add a script under the eval folder for mdl properties preparation
- generating csv file that is easier for non-technical users to update mdl properties
- filling in mdl properties defined in csv file to mdl json file

### DIFF
--- a/wren-ai-service/eval/preparation.py
+++ b/wren-ai-service/eval/preparation.py
@@ -98,9 +98,9 @@ def gen_new_mdl_from_csv(mdl_path: str, csv_path: str):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument(
-        "--mdl_path", type=str, help="Path to the MDL JSON file", required=True
+        "--mdl-path", type=str, help="Path to the MDL JSON file", required=True
     )
-    parser.add_argument("--csv_path", type=str, help="Path to the input CSV file")
+    parser.add_argument("--csv-path", type=str, help="Path to the input CSV file")
     args = parser.parse_args()
 
     if args.mdl_path and not args.csv_path:
--- a/wren-ai-service/eval/preparation.py
+++ b/wren-ai-service/eval/preparation.py
@@ -1,0 +1,107 @@
+import argparse
+from pathlib import Path
+
+import orjson
+
+
+def gen_eval_preparation_data_from_json_to_csv(mdl_path: str):
+    assert Path(mdl_path).exists(), f"File not found: {mdl_path}"
+
+    with open(mdl_path) as file:
+        mdl = orjson.loads(file.read())
+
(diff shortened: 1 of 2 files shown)


wren-ai-service/eval/preparation.py
```diff
@@ -0,0 +1,109 @@
...
+    with open(f"{Path(mdl_path).stem}_new.json", "w") as file:
+        file.write(orjson.dumps(mdl).decode())
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--mdl_path", type=str, help="Path to the MDL JSON file", required=True
+    )
+    parser.add_argument("--csv_path", type=str, help="Path to the input CSV file")
```
paopa, line 103: nitpick: I recommend avoiding the use of both hyphens and underscores in the option. Personally, I prefer using hyphens.

wren-ai-service/eval/preparation.py
```diff
@@ -0,0 +1,109 @@
+import argparse
+from pathlib import Path
+
+import orjson
+
+
+def gen_eval_preparation_data_from_json_to_csv(mdl_path: str):
+    assert Path(mdl_path).exists(), f"File not found: {mdl_path}"
```
paopa, line 8: assertions can be disabled with the -O option or by setting the PYTHONOPTIMIZE environment variable to a non-empty string. If we expect this behavior to extend beyond debugging purposes, using a conditional statement might be a better approach.

See also:
- https://docs.python.org/3/using/cmdline.html#cmdoption-O
- https://docs.python.org/3/using/cmdline.html#envvar-PYTHONOPTIMIZE

wren-ai-service/eval/preparation.py
```diff
@@ -0,0 +1,109 @@
...
+            )
+
+    with open(f"{Path(mdl_path).stem}.csv", "w") as file:
+        for row in csv_data:
+            file.write(",".join(row) + "\n")
+
+
+def gen_new_mdl_from_csv(mdl_path: str, csv_path: str):
+    assert Path(mdl_path).exists(), f"File not found: {mdl_path}"
+    assert Path(csv_path).exists(), f"File not found: {csv_path}"
```
paopa, line 43: A gentle mention, the same as the above comment

paopa (approved):
LGTM. have a few nitpick comments tho. not all need to be done. 😆

wren-ai-service/eval/preparation.py
cyyeh: adjusted

wren-ai-service/eval/preparation.py
cyyeh: I'll keep it as of now

PR #595, opened 2024-08-13
chore(wren-ai-service): handle add_quotes error

cyyeh:
sometimes generated sql may break sqlglot's transpile method, this PR fixes this issue

### DIFF
--- a/wren-ai-service/demo/utils.py
+++ b/wren-ai-service/demo/utils.py
@@ -5,7 +5,7 @@ import os
 import re
 import time
 from pathlib import Path
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Tuple
 
 import orjson
 import pandas as pd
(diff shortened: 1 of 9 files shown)


wren-ai-service/demo/utils.py
```diff
@@ -25,8 +25,12 @@
...
 
 
-def add_quotes(sql: str) -> str:
-    return sqlglot.transpile(sql, read="trino", identify=True)[0]
+def add_quotes(sql: str) -> Tuple[str, bool]:
+    try:
+        quoted_sql = sqlglot.transpile(sql, read="trino", identify=True)[0]
+        return quoted_sql, True
+    except Exception:
+        return sql, False
```
paopa, line 33: I found the same function in diff utils module, I think we can consider to remove the duplication.

paopa (approved):
LGTM. have a minor comment, which is not necessary to address here, perhaps we can consider to.

PR #599, opened 2024-08-14
chore(wren-ai-service): retrieval improvement

cyyeh:
indexing pipeline:
1. 3 collections: db_schema, table_descriptions, view_questions
2. to solve llm token window limit issue for indexing, we have a new env called `COLUMN_INDEXING_BATCH_SIZE` which  users can decide how many columns to index in one document at one time

retrieval pipeline:
1. select top 10(`TABLE_RETRIEVAL_SIZE`) tables based on table name and table descriptions (table_descriptions collection)
2. select top 1000(`TABLE_COLUMN_RETRIEVAL_SIZE`) tables and columns based on previous results (db_schma)
3. use llm to choose which tables and columns are needed to answer the question

we also expose two env vars for table and column selection: `TABLE_RETRIEVAL_SIZE` and `TABLE_COLUMN_RETRIEVAL_SIZE`


wren-ai-service/src/__main__.py
```diff
@@ -37,7 +37,23 @@ async def lifespan(app: FastAPI):
...
+        table_retrieval_size=(
+            int(os.getenv("TABLE_RETRIEVAL_SIZE"))
+            if os.getenv("TABLE_RETRIEVAL_SIZE")
+            else 10
+        ),
+        table_column_retrieval_size=(
+            int(os.getenv("TABLE_COLUMN_RETRIEVAL_SIZE"))
+            if os.getenv("TABLE_COLUMN_RETRIEVAL_SIZE")
+            else 1000
+        ),
```
paopa, line 56: just mention first, and I don't think it need to be done in this PR. we can consider to use a config class to initial all env and it will make the code more concise.

paopa (approved):
LGTM

PR #600, opened 2024-08-14
chore(wren-ai-service):update-launcher

cyyeh:
disable SHOULD_FORCE_DEPLOY by default and enabled only using launcher


wren-launcher/utils/docker.go
```diff
@@ -29,10 +29,14 @@ const (
...
 func replaceEnvFileContent(content string, projectDir string, openaiApiKey string, openAIGenerationModel string, hostPort int, aiPort int, userUUID string, telemetryEnabled bool) string {
-	// rplace PROJECT_DIR
+	// replace PROJECT_DIR
 	reg := regexp.MustCompile(`PROJECT_DIR=(.*)`)
 	str := reg.ReplaceAllString(content, "PROJECT_DIR="+projectDir)
 
+	// replace SHOULD_FORCE_DEPLOY
+	reg = regexp.MustCompile(`SHOULD_FORCE_DEPLOY=(.*)`)
+	str = reg.ReplaceAllString(str, "SHOULD_FORCE_DEPLOY=1")
+
```
paopa, line 39: curiously, do we expect always enabling this function? it seems whether we set any value(include the empty string), it will be force to enable.

the following is the example code.

```go
package main

import (
	"fmt"
	"regexp"
)

func main() {
	str := "SHOULD_FORCE_DEPLOY="

	reg := regexp.MustCompile(`SHOULD_FORCE_DEPLOY=(.+)`)
	str = reg.ReplaceAllString(str, "SHOULD_FORCE_DEPLOY=1")

	fmt.Println(str)
}
```

paopa: discussed offline, will force enable this config

paopa (approved):
LGTM

PR #601, opened 2024-08-14
fix(wren-ai-service): expend the timeout and move the waiting check mechanism when force deploying

paopa:
This PR aims to extend the timeout, as the AI service startup process sometimes requires additional time. It also moves the checking mechanism that activates when force deployment is enabled.


cyyeh (approved):
LGTM

PR #602, opened 2024-08-15
chore(wren-ai-service): Refine mdl preparation script

imAsterSun:
- Deal with models and columns without `properties` keys
- Use csv read/write functions can escape commas within quotes, prevent descriptions containing commas breaking the csv structure. 
  - Note: Still recommend to save the output csv file as xlsx file, if expecting users to edit descriptions with Excel.

### DIFF
--- a/wren-ai-service/eval/preparation.py
+++ b/wren-ai-service/eval/preparation.py
@@ -1,4 +1,5 @@
 import argparse
+import csv
 from pathlib import Path
 
 import orjson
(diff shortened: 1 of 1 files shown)


cyyeh (approved):
LGTM

wren-ai-service/eval/preparation.py
```diff
@@ -45,51 +48,55 @@ def gen_new_mdl_from_csv(mdl_path: str, csv_path: str):
     with open(mdl_path) as file:
         mdl = orjson.loads(file.read())
 
-    with open(csv_path) as file:
-        csv_data = [line.strip().split(",") for line in file.readlines()]
+    with open(csv_path, newline="\n") as file:
+        csv_data = csv.reader(file)
 
-    csv_data_by_table = {}
+        csv_data_by_table = {}
```
cyyeh, line 54: nitpick: L54-L99 don't need to be inside with block

imAsterSun: Fixed, thanks

PR #609, opened 2024-08-19
fix(wren-ui): UI layout is broken when selecting a sample dataset

fredalai:
## Description
 UI layout is broken when selecting a sample dataset

###  Video (Bug)
https://github.com/user-attachments/assets/7edff7d7-452d-4a79-bb8e-97431497f5e7


andreashimin (approved):
LGTM

PR #614, opened 2024-08-20
chore(wren-ai-service): fix eval

cyyeh:
- add some messages after running prediction and evaluation for better ux
image
- upgrade deepeval to 1.0.6
- some simple refactor
- fix accuracy metric

### DIFF
--- a/wren-ai-service/eval/metrics/column.py
+++ b/wren-ai-service/eval/metrics/column.py
@@ -76,6 +76,9 @@ class AccuracyMetric(BaseMetric):
             expected_dataset = await self._retrieve_data(test_case.expected_output)
             actual_dataset = await self._retrieve_data(test_case.actual_output)
 
+            print(f"expected columns: {set(expected_dataset.columns)}")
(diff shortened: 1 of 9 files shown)


wren-ai-service/eval/metrics/column.py
```diff
@@ -11,33 +11,60 @@
...
 
     def is_subset(self, expected: pd.DataFrame, actual: pd.DataFrame) -> bool:
-        if not set(actual.columns).issubset(set(expected.columns)):
+        if not set(expected.columns).issubset(set(actual.columns)):
             return False
 
-        common_columns = actual.columns
-        if common_columns.empty:
+        common_columns = sorted(expected.columns)
+        if not common_columns:
```
paopa, line 26: we use expected columns as the common. thus, we might consider changing the condition to check if the actual columns are empty. it seems unlikely that the expected columns will be empty. wdyt?

cyyeh: ok, removed

paopa (approved):

PR #625, opened 2024-08-29
feat(wren-ai-service): optional attribute user_id for Langfuse metadata

paopa:
Before this PR, we used the project ID as the user ID for Langfuse. In our team discussion, we decided to include the project ID in the metadata and introduce an optional user_id attribute for API requests. We also ensured that the logic remains consistent.


cyyeh (approved):
LGTM

PR #631, opened 2024-08-30
feat(wren-ai-service): sql-answer-pipeline

cyyeh:
- add a sql to answer pipeline that aims to answer user's question based on data retrieved by sql

### DIFF
--- a/wren-ai-service/src/providers/engine/wren.py
+++ b/wren-ai-service/src/providers/engine/wren.py
@@ -89,7 +89,7 @@ class WrenIbis(Engine):
                 return True, None
 
             res = await response.text()
-            return False, orjson.loads(res)
+            return False, res
 
 
 @provider("wren_engine")
@@ -121,4 +121,4 @@ class WrenEngine(Engine):
             if response.status == 200:
                 return True, None
             res = await response.text()
-            return False, orjson.loads(res)
+            return False, res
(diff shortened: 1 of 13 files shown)


wren-ai-service/src/providers/engine/wren.py
```diff
@@ -18,30 +18,32 @@ def __init__(self, endpoint: str = os.getenv("WREN_UI_ENDPOINT")):
...
-                        "sql": remove_limit_statement(sql),
-                        "projectId": project_id,
-                    }
-                },
+                "variables": {"data": data},
             },
         ) as response:
             res = await response.json()
-            if res.get("data"):
-                return True, None
```
paopa, line 44: how about keep the error and return one more value for data? I think using the same position to represent two meaning might be not a good choice.

e.g. 
```py
return status, data, error
```

wren-ai-service/src/web/v1/services/sql_answer.py
```diff
@@ -0,0 +1,149 @@
...
+    error: Optional[SqlAnswerError] = None
+
+
+class SqlAnswerService:
+    def __init__(
+        self,
+        pipelines: dict[str, BasicPipeline],
+    ):
+        self._pipelines = pipelines
+        self._sql_answer_results = {}
```
paopa, line 55: I think it should be instead by the TTL cache. you could refer to the function in ask service.

paopa (approved):
LGTM

PR #663, opened 2024-09-11
chore(wren-ai-service): refine retrieval in follow-up question

cyyeh:
- should include previous query summary in follow-up asking process
- add whitelist for model and column properties for indexing


paopa (approved):
LGTM

PR #682, opened 2024-09-25
feat(wren-ai-service): support multi compnents(llm, embedder, engine, document store) configuration

paopa:
This pull request introduces several updates and improvements to our codebase:

1. **Configuration Management**:
   - The system now supports fetching configuration from multiple sources, including environment variables and a new multi-LLM (Language Learning Model) configuration.
   - The existing method of fetching configuration from environment variables is retained to ensure backward compatibility.

2. **Deployment Enhancements**:
   - The multi-LLM configuration is currently supported only for local deployments.
   - Future plans include extending this support to Docker and Kubernetes deployments once the multi-LLM configuration is deemed stable and mature.

#### Detailed Description
- **Backward Compatibility**: We have ensured that the current method of fetching configuration from environment variables remains intact. This allows for a smooth transition and gives us the flexibility to switch to the new multi-LLM configuration gradually.
- **Local Deployment**: The new multi-LLM configuration is currently enabled only for local deployments. This allows us to test and refine the configuration before rolling it out to more complex deployment environments like Docker and Kubernetes.

#### Future Plans
- Once the multi-LLM configuration is stable and mature, we plan to:
  - Remove the old method of fetching configuration from environment variables.
  - Update the deployment approach for Docker and Kubernetes to support the new configuration.

### DIFF
--- a/wren-ai-service/config.example.yaml
+++ b/wren-ai-service/config.example.yaml
@@ -1,5 +1,5 @@
 type: llm
-vender: openai_llm
+provider: openai_llm
 models:
   - model: gpt-4o-mini
     kwargs: {}
@@ -9,17 +9,17 @@ api_base: https://api.openai.com/v1
 
 ---
 type: llm
-vender: azure_openai_llm
+provider: azure_openai_llm
 models:
   - model: gpt-4o
     kwargs: {}
 api_key: sk-xxx
 api_base: https://api.openai.com/v1
-version: "2024-05-13"
+api_version: "2024-05-13"
 
 ---
 type: embedder
-vender: openai_embedder
+provider: openai_embedder
 models:
   - model: text-embedding-3-small
     dimension: 3072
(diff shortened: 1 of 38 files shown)


wren-ai-service/config.example.yaml
```diff
@@ -0,0 +1,105 @@
...
+provider: wren_engine
+endpoint: http://localhost:8080
+manifest: ""
+
+---
+type: document_store
+provider: qdrant
+location: http://localhost:6333
+api_key: ""
+embedding_model_dim: 3072
```
cyyeh, line 65: we can simply use the dimension field from the embedder type?

wren-ai-service/src/providers/__init__.py
```diff
@@ -0,0 +1,197 @@
...
+            "llm": pipe.get("llm"),
+            "embedder": pipe.get("embedder"),
+            "document_store": pipe.get("document_store"),
+            "engine": pipe.get("engine"),
+        }
+        for pipe in entry["pipes"]
+    }
+
+
+def convert_data(config: list[dict]) -> dict:
```
cyyeh, line 90: I suppose we can have a more clear function name here?

wren-ai-service/src/providers/__init__.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def __iter__(self):
+        return iter(self.value)
+
+    def __len__(self):
+        return len(self.value)
+
+
+def generate_components() -> dict[str, PipelineComponent]:
+    raw = load_config()
```
cyyeh, line 160: I suppose we can have `logging.info` here to check whether we are loading from yaml or from env files here

wren-ai-service/src/providers/__init__.py
```diff
@@ -0,0 +1,197 @@
...
+) -> LLMProvider | EmbedderProvider | DocumentStoreProvider | Engine:
+    return loader.get_provider(config.get("provider"))(**config)
+
+
+def load_config(path: str = "config.yaml") -> list[dict]:
+    if not os.path.exists(path):
+        return []
+
+    with open(path, "r") as f:
+        return list(yaml.load_all(f, Loader=SafeLoader))
```
cyyeh, line 28: you could simply use `yaml.safe_load_all(f)`

cyyeh (approved):
LGTM, comments will be dealt in the near future cc @paopa

PR #721, opened 2024-10-04
chore(wren-ai-service): minor updates

cyyeh:
1. fix overriding config.yaml in demo
2. fix table column selection by incorporating struct as column; also we may need to use gpt-4o in this case


paopa (approved):
LGTM

PR #753, opened 2024-10-14
feature(wren-ui): inject description when import from ibis server

onlyjackfrost:
## Description
- inject description into properties when listing table from ibis server
    - table
    - column
    - nestedColumn

## Screenshots
BigQuery
- can import table & column & nested column's description from the data source
image


andreashimin (approved):
LGTM

PR #779, opened 2024-10-21
chore(wren-ai-service): fix eval bug

cyyeh:
exact match is used for spider dataset; however, in the original code, the init part needs to read the spider dataset content, which is not suitable if we don't use spider dataset for evalulation


wren-ai-service/eval/metrics/spider/exact_match.py
paopa, line 1: i think they passed kmaps from parameter as well, then there is no diff between two implementations.

https://github.com/taoyds/test-suite-sql-eval/blob/e97acc546ecbee8fa27fa8dbf025ef61493a876c/evaluation.py#L504

paopa:
Discussed the matter with @cyyeh offline. we'll create another PR to deal the issue that occurring when the meta file or db dir doesn't exist under the path of `/tools/dev/etc/spider1.0`.

PR #784, opened 2024-10-21
chore(wren-ai-service): refine sql generation

cyyeh:
- allow generating candidates in generating phase
- stablize generating multilingual output
- fix time issue

### DIFF
--- a/wren-ai-service/src/web/v1/services/ask.py
+++ b/wren-ai-service/src/web/v1/services/ask.py
@@ -248,17 +248,12 @@ class AskService:
                         configurations=ask_request.configurations,
                     )
 
-                valid_generation_results = []
-                valid_sql_summary_results = []
                 if sql_valid_results := text_to_sql_generation_results["post_process"][
                     "valid_generation_results"
                 ]:
-                    valid_generation_results += sql_valid_results
-
-                if valid_generation_results:
                     valid_sql_summary_results = (
                         await self._add_summary_to_sql_candidates(
-                            valid_generation_results,
+                            sql_valid_results,
                             ask_request.query,
                             ask_request.configurations.language,
                         )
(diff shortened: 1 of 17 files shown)


wren-ai-service/src/web/v1/services/ask.py
```diff
@@ -215,11 +248,23 @@ async def ask(
...
+                            ask_request.configurations.language,
+                        )
+                    )
+                    api_results += [
+                        AskResult(**result) for result in valid_sql_summary_results
+                    ]
+                    self._ask_results[query_id] = AskResultResponse(
+                        status="generating",
+                        response=api_results,
+                    )
```
paopa, line 267: i think the results might be more than three. Do we need to remove any after the third candidate?

wren-ai-service/src/web/v1/services/ask.py
```diff
@@ -233,25 +278,31 @@ async def ask(
...
+                        valid_sql_summary_results = (
+                            await self._add_summary_to_sql_candidates(
+                                valid_generation_results,
+                                ask_request.query,
+                                ask_request.configurations.language,
+                            )
+                        )
+                        api_results += [
+                            AskResult(**result) for result in valid_sql_summary_results
+                        ]
```
paopa, line 293: the reference is already pointed to the attribute of response object, thus add new results into this variable, it might be occurring more than three candidates back. i think we need to check how many candidates had before putting them into the array.

paopa: Here is the example to show the case

```python
from dataclasses import dataclass


@dataclass
class Demo:
    id: str


@dataclass
class Resp:
    demos: list[Demo]


results = []

results += [Demo(id="1")]

resp = Resp(demos=results)

print(resp)


results += [Demo(id="2")]

print(resp)

```

paopa (approved):
LGTM

PR #788, opened 2024-10-22
feat(wren-ui): Add candidate showing flow when in generating state

andreashimin:
## Description
This feature aims to show the candidate first while in the generating state.
Also, we provide skeletons to users to make it clear to know there is results will show after finishing the query.

image


fredalai (approved):
LGTM

PR #844, opened 2024-10-30
feat(wren-ai-service): intent classification and data assistance pipeline

cyyeh:


wren-ai-service/src/providers/llm/openai.py
```diff
@@ -188,16 +198,19 @@ def get_generator(
                 f"Creating OpenAI API-compatible generator {self._generation_model} with model kwargs: {self._model_kwargs}"
             )
 
+        print("")
```
paopa, line 201: i think we could remove this 😆

wren-ai-service/src/web/v1/services/ask.py
```diff
@@ -104,8 +104,10 @@ class AskResultResponse(BaseModel):
         "failed",
         "stopped",
     ]
+    type: Optional[Literal["MISLEADING_QUERY", "GENERAL", "TEXT_TO_SQL"]] = None
     response: Optional[List[AskResult]] = None
     error: Optional[AskError] = None
+    _data_assistance_task: Optional[asyncio.Task] = None
```
paopa, line 110: it seems not be used anywhere, could we remove this?

wren-ai-service/src/pipelines/generation/data_assistance.py
```diff
@@ -0,0 +1,198 @@
...
+DATA_ASSISTANCE_MODEL_KWARGS = {"response_format": {"type": "text"}}
+
+
+class DataAssistance(BasicPipeline):
+    def __init__(
+        self,
+        llm_provider: LLMProvider,
+        **kwargs,
+    ):
+        self._user_queues = {}
```
paopa, line 91: I think we could consider using a TTL cache to store it, which would help prevent memory usage from increasing over time.

paopa: another option is to store them in the cache of ask service, but we need to further consider the design.

paopa (commented):
overall LGTM. have some minor comments. 👍

paopa (approved):

PR #846, opened 2024-10-30
chore(wren-ai-service): add correlation ID to langfuse

grieve54706:
Add correlation ID from the ibis server to the Langfuse.

The AI team usually provides the issues from the Langfuse, if we want to find the ibis log for the issue, we need to find it by time, and the time is not completely correct.

The correlation ID is from the ibis-server. The AI service gets data and the correlation ID through the web API, and the AI service sends Langfuse logs. So, we must explain the correlation ID and return it to the AI service.

### DIFF
--- a/wren-ui/src/apollo/server/services/queryService.ts
+++ b/wren-ui/src/apollo/server/services/queryService.ts
@@ -25,11 +25,9 @@ export interface ColumnMetadata {
   type: string;
 }
 
-export interface PreviewDataResponse {
-  correlationId?: string;
-  processTime?: string;
-  columns?: ColumnMetadata[];
-  data?: any[][];
+export interface PreviewDataResponse extends IbisResponse {
+  columns: ColumnMetadata[];
+  data: any[][];
 }
 
 export interface DescribeStatementResponse {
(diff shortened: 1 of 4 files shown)


grieve54706:
Hi @onlyjackfrost, @cyyeh. Could you check this PR?

wren-ui/src/apollo/server/services/queryService.ts
```diff
@@ -28,8 +28,8 @@ export interface ColumnMetadata {
 export interface PreviewDataResponse {
   correlationId?: string;
   processTime?: string;
-  columns: ColumnMetadata[];
-  data: any[][];
+  columns?: ColumnMetadata[];
```
onlyjackfrost, line 31: I prefer not to share type definitions here, as preview and dry-run have distinct return formats, and they should be defined separately. Can we define their types individually?

onlyjackfrost (commented):
Could you please explain why the correlationId is being added and returned to the AI service?

wren-ui/src/apollo/server/services/queryService.ts
grieve54706: Sure

grieve54706:
> Could you please explain why the correlationId is being added and returned to the AI service?

Hi @onlyjackfrost. Because the AI team usually provides the issues from the Langfuse, if we want to find the ibis log for the issue, we need to find it by time, and the time is not completely correct. 
The correlation ID is from the ibis-server. The AI service gets data and the correlation ID through the web API, and the AI service sends Langfuse logs. So, you must explain the correlation ID and return it to the AI service.

onlyjackfrost (approved):

PR #850, opened 2024-10-31
chore(wren-ai-service): generate one sql only

cyyeh:


wren-ai-service/src/web/v1/services/ask.py
```diff
@@ -227,81 +227,67 @@ async def ask(
                         )
                         for result in historical_question_result
```
paopa, line 228: i'm considering if we should remove the historical pipeline from ask, due to we provide the one sql only. but we can move the function to draft sql pipeline for diversity

paopa (commented):
overall LGTM! have a comment only.

wren-ai-service/src/web/v1/services/ask.py
paopa: after discussed offline, we'll provide the historical question as the first-tier answer.

paopa (approved):
LGTM

PR #855, opened 2024-11-01
chore(wren-ui): Create thread directly when receiving one candidate

andreashimin:
## Description
Skip candidate selection and navigate directly to thread details when a single result is received.


fredalai (approved):
LGTM

PR #867, opened 2024-11-04
chore(wren-ai-service): Update retrieval prompt rule for GA4 struct type column

imAsterSun:
- Update retrieval prompt rule for GA4 struct type column


cyyeh (approved):
LGTM

PR #871, opened 2024-11-05
chore(wren-ai-service): optimize retrieval process

cyyeh:
1. add env `ALLOW_USING_DB_SCHEMAS_WITHOUT_PRUNING`, default: False
2. use tiktoken library to count tokens of db schemas ddl, and apply the rules of column pruning as follows:
- using gpt-4o or gpt-4o-mini, if token count of db schemas ddl is less than 100k -> column pruning
- using gpt-4o or gpt-4o-mini, if token count of db schemas ddl is greater than or equal to 100k -> no column pruning
- using other llms -> no column pruning, unless ALLOW_USING_DB_SCHEMAS_WITHOUT_PRUNING is True
- by default, ALLOW_USING_DB_SCHEMAS_WITHOUT_PRUNING is False
- in the near future after we accomodate multi-llms, users need to fill in context window size for more intricate control on whether we need column pruning
3. remove gpt-3.5-turbo and gpt-4-turbo

### DIFF
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -16,7 +16,7 @@ WREN_UI_ENDPOINT=http://docker.for.mac.localhost:3000
 # LLM
 LLM_OPENAI_API_KEY=
 EMBEDDER_OPENAI_API_KEY=
-# gpt-4o-mini, gpt-4o, gpt-4-turbo, gpt-3.5-turbo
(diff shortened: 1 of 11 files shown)


paopa (approved):
LGTM

PR #899, opened 2024-11-12
chore(wren-ui): Asking experience enhancement

andreashimin:
## Description
This PR aims to enhance the experience of create asking to create thread within home page & home/[id] page.

It has modified the following issues:
- Fill the loading timeframe when `CreateAsking` to `Searching` state & `CreateThread` to `Generating` state 
  (Currently state only appear while polling `AskingTask`, this changes make the state also appear while `CreateAsking` & `CreateThread`)
- Show thread title & question first while polling thread response
- Show fade-in animation after the answer polling done & give last answer a min-height to reduce the flicker caused by scroll top changed
- Remove global loading in home & home/[id] pages

## Recording

- Before 

https://github.com/user-attachments/assets/0b7154b2-527d-4c8a-9293-7f1771c1dff5

- After

https://github.com/user-attachments/assets/cb341584-ced6-4c2b-9c8d-d795437239f1

PR #904, opened 2024-11-13
fix(wren-ui): Separate view step sql & view full sql condition

andreashimin:
## Description
In PR #899, adding `isViewSQL` caused confusion within the `CollapseContent` template condition. This fix introduces `isViewStepSQL` to more clearly separate the conditions.


fredalai (approved):
LGTM

PR #908, opened 2024-11-13
chore(wren-ai-service): fix sql generation issues

cyyeh:
- update usecases testing
- fix sql generation issues


cyyeh:
merged here: https://github.com/Canner/WrenAI/pull/907

PR #916, opened 2024-11-15
fix(wren-ui): prompt result add wrapper after rebasing

andreashimin:
## Description
fix the prompt result style


fredalai (approved):
LGTM

PR #928, opened 2024-11-19
Ensure answers correctly indicate their associated view

AryanK1511:
**This PR fixes #880**

This update ensures that the `viewId` is correctly returned to the user. Previously, only the `sql` was passed with `onSelectResult`. Now, we include the `viewId` with:  

```javascript
onSelectResult && onSelectResult({ sql: result.sql, viewId: result.view?.id });
```  

A video is included to demonstrate that the functionality now works as expected.  

https://github.com/user-attachments/assets/0723833c-37e7-40b3-8c41-5de141bf6c63

Special thanks to @cyyeh for collaborating on this fix and helping bring it to completion!

PR #930, opened 2024-11-20
fix(wren-ui): Create thread from view summary value change to question

andreashimin:
## Description
This PR fixes the `summary` property to correctly use the input question when users create a thread from a view. It is related to #928, which addressed issues with creating threads from views.


fredalai (approved):
LGTM

PR #932, opened 2024-11-20
fix(wren-ui): adjust header key case for correlation id

grieve54706:
The header key from ibis-server is all lowercase.
In wren-ui, we use the wrong header key so we can not receive the current value.


grieve54706:
Hi @onlyjackfrost, could you review this PR? Thanks.

onlyjackfrost (approved):

PR #934, opened 2024-11-20
fix(wren-ui): await thread recommendation question generation

fredalai:
Add await for generateThreadRecommendationQuestions API


andreashimin (approved):
lgtm

PR #936, opened 2024-11-20
chore(wren-ai-service): refactor

cyyeh:
- simplify pipeline dry run
- use formal version of AsyncDriver instead of experimental version
- unify configuration for apis

### DIFF
--- a/wren-ai-service/src/pipelines/generation/followup_sql_generation.py
+++ b/wren-ai-service/src/pipelines/generation/followup_sql_generation.py
@@ -4,7 +4,7 @@ from pathlib import Path
 from typing import Any, List
 
 from hamilton import base
-from hamilton.experimental.h_async import AsyncDriver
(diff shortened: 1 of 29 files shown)


wren-ai-service/src/pipelines/generation/followup_sql_generation.py
```diff
@@ -20,7 +20,8 @@
     sql_generation_system_prompt,
 )
 from src.utils import async_timer, timer
-from src.web.v1.services.ask import AskConfigurations, AskHistory
+from src.web.v1.services import Configuration
+from src.web.v1.services.ask import AskHistory
```
paopa, line 24: nitpick: we could consider moving them into pipeline layer in the future. i think it can reduce the bi-dependency between pipeline and service layer

paopa (approved):
LGTM. having a nitpick comment, i think we could ship this PR first.

PR #946, opened 2024-11-22
feat(wren-ai-service): litellm integration

cyyeh:
- tested using OpenAI and Ollama
- for setup, could check tools/config/config.full.yaml
- in the future, users only need to update config and environment variable for api keys to use new llms
- this PR only implements LLM. we could implement embedder in the future

### DIFF
--- a/wren-ai-service/src/providers/__init__.py
+++ b/wren-ai-service/src/providers/__init__.py
@@ -78,6 +78,7 @@ def llm_processor(entry: dict) -> dict:
             "kwargs": model.get("kwargs"),
             "api_base": model.get("api_base"),
             "api_version": model.get("api_version"),
+            "api_key_name": model.get("api_key_name"),
             **others,
         }
     return returned
--- a/wren-ai-service/src/providers/__init__.py
+++ b/wren-ai-service/src/providers/__init__.py
@@ -70,12 +70,14 @@ def llm_processor(entry: dict) -> dict:
     """
     others = {k: v for k, v in entry.items() if k not in ["type", "provider", "models"]}
     returned = {}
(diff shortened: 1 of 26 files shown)


wren-ai-service/src/providers/__init__.py
```diff
@@ -70,12 +70,15 @@ def llm_processor(entry: dict) -> dict:
...
-        model_name = f"{entry['provider']}.{model['model']}"
+    for model in entry.get("models", []):
+        model_name = f"{entry.get('provider')}.{model.get('model')}"
         returned[model_name] = {
-            "provider": entry["provider"],
-            "model": model["model"],
-            "kwargs": model["kwargs"],
+            "provider": entry.get("provider"),
+            "model": model.get("model"),
+            "kwargs": model.get("kwargs"),
```
paopa, line 78: I think it’s better to keep direct access since these keys are required.

wren-ai-service/src/providers/__init__.py
```diff
@@ -70,12 +70,15 @@ def llm_processor(entry: dict) -> dict:
...
+            "api_base": model.get("api_base"),
+            "api_version": model.get("api_version"),
+            "api_key_name": model.get("api_key_name"),
```
paopa, line 81:
```python
addition = {k: v for k, v in model.items() if k not in ["model", "kwargs"]}
```

perhaps we could use an additional dict to store the optional items defined at the model level. it seems these keys arent for all llm provider.

paopa (commented):
thanks for the great work! have a few comments leave.

paopa (approved):
LGTM

PR #951, opened 2024-11-25
chore(wren-ai-service): update package deps

cyyeh:


wren-ai-service/src/providers/document_store/qdrant.py
```diff
@@ -265,7 +265,7 @@ async def write_documents(
 
         document_objects = self._handle_duplicate_documents(
             documents=documents,
-            index=self.index,
```
paopa, line 268: im curious why we comment this out.

cyyeh: new version of qdrant sdk doesn't need that anymore

paopa (approved):

PR #957, opened 2024-11-26
fix(wren-ai-service): wrong condition to avoid lacking some valid question

paopa:
This PR fixes an issue around handling question validation and category limits.

## Changes
- Added proper return values for validation steps for tracing
- Improved category and question limit handling


cyyeh (approved):
LGTM

PR #962, opened 2024-11-28
fix(wren-ui): adjust recommend question to reduce cost 

onlyjackfrost:
adjust recommend question to reduce cost 
- remove regenerate button
- reduce requested recommend question


fredalai (approved):
LGTM

PR #963, opened 2024-11-28
chore(wren-ui): Add text-truncate to recommend questions label

andreashimin:
## Description
- add text truncate to recommend questions label
- add HTML title to label


fredalai (approved):
LGTM

PR #991, opened 2024-12-12
chore(wren-ai-service): refine ai service

cyyeh:
- there should be no sql in GENERAL streaming response
- we should not accept sql in user input
- follow-up ask improvement


paopa (approved):
LGTM

PR #1000, opened 2024-12-15
1st i18n compatible

johnfelipe:
i want to contribute for have this nice and perfect web app with i18n compatible for spanish and italian

### DIFF
--- a/ejemplo_de_cambio_de_idioma
+++ b/ejemplo_de_cambio_de_idioma
@@ -1,0 +1,19 @@
+import { useTranslation } from 'react-i18next';
+
+function LanguageSwitcher() {
+  const { i18n } = useTranslation();
+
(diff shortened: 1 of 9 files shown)


wwwy3y3:
Hi @johnfelipe thanks for the contribution! We'll discuss some implementation details (ex: where we store the preferred language settings ? in db or local storage is fine at early stage?) and get back to you.

I'll add this PR to the open roadmap backlog for now.

wwwy3y3:
Some suggestions:
1. If you have any thoughts on how we store the preference, chime in as you want
2. some files (ex: `ejemplo_de_cambio_de_idioma`) don't make much sense to me. Seems like misplaced files ?
3. You may want to elaborate more how you want to implement this feature at higher level

Thanks!

PR #1016, opened 2024-12-18
chore(wren-ai-service): improve text2chart and text2answer

cyyeh:
- text2chart:
  - fix schema JSON decode issue
- text2answer:
  - remove reasoning process
  - remove technical jargon like SQL syntax
  - refine prompt wording

### DIFF
--- a/wren-ai-service/src/pipelines/generation/chart_adjustment.py
+++ b/wren-ai-service/src/pipelines/generation/chart_adjustment.py
@@ -93,6 +93,10 @@ class ChartAdjustmentPostProcessor:
             generation_result = orjson.loads(replies[0])
             reasoning = generation_result.get("reasoning", "")
             if chart_schema := generation_result.get("chart_schema", {}):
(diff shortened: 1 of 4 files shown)


wren-ai-service/src/pipelines/generation/chart_adjustment.py
```diff
@@ -93,6 +93,10 @@ def run(
             generation_result = orjson.loads(replies[0])
             reasoning = generation_result.get("reasoning", "")
             if chart_schema := generation_result.get("chart_schema", {}):
+                # sometimes the chart_schema is still in string format
```
paopa, line 96: Im curious about what is the string format be like. could you provide me an example? thanks

paopa (approved):
LGTM

wren-ai-service/src/pipelines/generation/chart_adjustment.py
paopa: have discussed this comment with Jimmy offline

PR #1019, opened 2024-12-19
feat(wren-ui): chart error handling & filter overflow data for nominal field

andreashimin:
## Description
- add AI service NO_CHART error code & UI
- filter out overflow data for chart, add title to chart

## Screenshot
image
image


fredalai (approved):
LGTM

PR #1029, opened 2024-12-19
chore(wren-ui): change base image and optimize docker build

cyyeh:
- change base image from `alpiine` to `bookworm-slim`
- use yarn version 4.5.3
  - for development, when you change the working directory to wren-ui, the yarn version would automatically change to 4.5.3.
  - in my local environment(Mac, Apple M2 pro 16GB), `yarn install` takes 30-40 seconds
- optimize docker build
  - only install `curl`, skipping other system packages
  - disable next telemetry
  - make node env production, so yarn installation would skip dev dependencies
  - use `yarn dlx` to execute migration without installing package `knex` first
  - separate two different platforms in ci: linux/amd64, linux/arm64
  - performance improvement result of CI release image
    - original: ~51 min
    - now: ~9 min
  - CI release build testing result: https://github.com/Canner/WrenAI/actions/runs/12420333063

CAVEAT: this pr doesn't change yaml for CI releasing stable image yet

## Summary by CodeRabbit

- **New Features**
  - Introduced a new environment variable for image location in the release process.
  - Added support for multiple architectures in the image build process.
  - Created a new job for merging image digests in the workflow.

- **Configuration Updates**
  - Updated Yarn configuration to specify dependency management methods and Yarn version.

- **Docker Enhancements**
  - Changed the base image for Docker to improve compatibility.
  - Streamlined dependency installation process in the Dockerfile.

- **Package Management**
  - Specified the version of the package manager in the project configuration.


cyyeh:
Thanks @grieve54706 's work for CI's multiple platform image build. It's a great help :)

.github/workflows/ui-release-image.yaml
```diff
@@ -35,13 +48,59 @@ jobs:
         uses: docker/setup-qemu-action@v3
```
grieve54706, line 48: You don't need `QEMU` anymore. It's for cross platform building.

grieve54706 (approved):
LGTM

grieve54706:
Here is the reference about multi-platform building from Docker official
https://docs.docker.com/build/ci/github-actions/multi-platform

And the GitHub beta arm runner
https://github.blog/changelog/2024-06-24-github-actions-ubuntu-24-04-image-now-available-for-arm64-runners/

.github/workflows/ui-release-image.yaml
cyyeh: thanks for explanation!

wren-ui/Dockerfile
```diff
@@ -1,14 +1,28 @@
...
+FROM node:18-bookworm-slim AS base
+
+# Install required packages
+RUN apt-get update && apt-get install -y curl && rm -rf /var/lib/apt/lists/*
+
+# Next.js collects completely anonymous telemetry data about general usage.
+# Learn more here: https://nextjs.org/telemetry
+# Uncomment the following line in case you want to disable telemetry during the build.
+ENV NEXT_TELEMETRY_DISABLED 1
+ENV NODE_ENV production
```
wwwy3y3, line 10: I think we could remove this line in base stage image. Setting NODE_ENV to prod will have a larger impact than we thought. If you want yarn installation to skip dev dependencies, simply do this in `deps` stage, or add to `yarn install` (`NODE_ENV=production yarn install --immutable`) would be better, IMO.

seeking advice from @onlyjackfrost  as well.

wwwy3y3: I'm thinking you could just keep `apt-get update ...` commands in base stage image. Others commands that are related to `yarn` installation could be moved to `deps` stage.

wwwy3y3: ohh, we did have `ENV NODE_ENV production` at runner stage before, I guess it's fine. Let's keep it the same way.

wren-ui/Dockerfile
```diff
@@ -50,12 +53,12 @@ COPY --from=builder /app/.next/standalone ./
...
 
 EXPOSE 3000
 
 ENV PORT 3000
 
 # server.js is created by next build from the standalone output
 # https://nextjs.org/docs/pages/api-reference/next-config-js/output
-CMD knex migrate:latest && HOSTNAME="0.0.0.0" node server.js
+# use yarn dlx to run package without installing it
+CMD yarn dlx knex migrate:latest && HOSTNAME="0.0.0.0" node server.js
```
wwwy3y3, line 64: `yarn dlx` will download package, hence will make server startup more unpredictable & time consuming. Could we still put `knex` installation in build steps?

cyyeh: sth like this?
image

wwwy3y3: `RUN yarn global add knex`

cyyeh: fixed

wren-ui/Dockerfile
```diff
@@ -58,4 +61,4 @@ ENV PORT 3000
 
 # server.js is created by next build from the standalone output
```
onlyjackfrost, line 62: https://yarnpkg.com/migration/guide#use-yarn-dlx-instead-of-yarn-global

should we use `yarn dlx` on line 56? (There is no diff so I comment on this line)

onlyjackfrost: I ran into this error while building an image on my Mac (m1).
image

onlyjackfrost (commented):
@cyyeh I think it's helpful to do the yarn migration. Thanks!
I followed the yarn migration [instruction](https://yarnpkg.com/migration/guide#use-yarn-dlx-instead-of-yarn-global) and had one question.

wren-ui/Dockerfile
```diff
@@ -50,12 +53,12 @@ COPY --from=builder /app/.next/standalone ./
 COPY --from=builder /app/.next/static ./.next/static
 COPY migrations ./migrations
 COPY knexfile.js ./knexfile.js
-RUN yarn global add knex
+RUN yarn add knex
```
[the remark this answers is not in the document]
cyyeh: thanks for your input. However, we prefer stability since `yarn dlx knex` may fail if the package runs into some issues

wwwy3y3 (approved):

PR #1045, opened 2024-12-23
chore(wren-ai-service): make openai llm version fixed

cyyeh:
According to OpenAI docs, in production applications, it is a best practice to use dated model snapshot IDs instead of aliases, which may change periodically.

ref: https://platform.openai.com/docs/models#current-model-aliases


paopa (approved):
LGTM

PR #1051, opened 2024-12-23
fix(wren-ui): Calculated fields in model preview Issue & improvements

andreashimin:
## Description
This PR addresses an issue where some calculated fields (CFs) fail to display in the model preview.

## Root Cause:
The engine requires all columns, including calculated fields, to be explicitly listed in the query when previewing a model. Using a wildcard (*) does not include calculated fields in the result.

## Solution:
Update the model preview query to explicitly list all columns and calculated fields, ensuring accurate previews. 
For example:
```
SELECT "col"..., "cf"... FROM "model"
```

## Additional context
- Add functionality to disable calculated fields in the same model
- Add some column types


fredalai (approved):
LGTM

onlyjackfrost (approved):

PR #1062, opened 2024-12-26
fix(wren-ui): remove custom scale & add adjustment flag for adjust chart scenario

andreashimin:
## Description
- remove custom scale
- add adjustment flag for adjust chart scenario
- chart buttons deeper color

## Summary by CodeRabbit

- **New Features**
	- Added an `adjustment` property to chart-related data structures, enhancing the representation of chart adjustments.
	- Updated the `ChartAnswer` component to conditionally render a regeneration button based on the `adjustment` state.

- **Bug Fixes**
	- Improved error handling in the `ChartAnswer` component with a reusable regeneration button.

- **Style**
	- Enhanced button and summary element visibility in the chart UI by adjusting opacity and color properties.


wren-ui/src/apollo/server/schema.ts
```diff
@@ -684,6 +684,7 @@ export const typeDefs = gql`
     error: Error
     description: String
     chartSchema: JSON
+    adjustment: Boolean
```
wwwy3y3, line 687: @andreashimin where do you do a fallback ? I mean for response not adjusted or older responses not having this value, will this value default to `false` ?

fredalai (approved):
LGTM

wren-ui/src/apollo/server/schema.ts
andreashimin: there's no fallback for the value since we cannot know the existed chart schemas were adjusted or not.

The value will only appear `true` or `null`.

so if null the regenerate button will not show. But for new asking, it should be appear if the encounter the scenario.

PR #1063, opened 2024-12-26
feat(wren-ui): remove unnecessary sql column of thread table

fredalai:
## Description
The sql column in the thread table is no longer in use, so we will remove it.

## Summary by CodeRabbit

- **New Features**
	- Streamlined thread data structure by removing the `sql` field from various types and operations.
  
- **Bug Fixes**
	- Adjusted GraphQL queries and mutations to ensure they no longer reference the deprecated `sql` field.

- **Refactor**
	- Updated methods in the `AskingService` to eliminate SQL handling during thread creation.

- **Documentation**
	- Enhanced clarity in thread representation by removing unnecessary SQL fields from the schema and types.

### DIFF
--- a/wren-ui/migrations/20241226135712_remove_thread_sql.js
+++ b/wren-ui/migrations/20241226135712_remove_thread_sql.js
@@ -1,0 +1,26 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function (knex) {
+  // drop foreign key constraint before altering column type to prevent data loss
+  await knex.schema.alterTable('thread_response', (table) => {
+    table.dropForeign('thread_id');
+  });
+  await knex.schema.alterTable('thread', (table) => {
+    table.dropColumn('sql');
+  });
+  await knex.schema.alterTable('thread_response', (table) => {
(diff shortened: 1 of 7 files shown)


wren-ui/migrations/20241226135712_remove_thread_sql.js
```diff
@@ -0,0 +1,32 @@
...
+exports.down = async function (knex) {
+  await knex.schema.alterTable('thread_response', (table) => {
+    table.dropForeign('thread_id');
+  });
+  await knex.schema.alterTable('thread', (table) => {
+    table.text('sql').nullable();
+  });
+  await knex.schema.alterTable('thread_response', (table) => {
+    table.foreign('thread_id').references('thread.id').onDelete('CASCADE');
+  });
```
wwwy3y3, line 31: Isn't it the same as the statements in `up` migration ?

```
ON DELETE { NO ACTION | CASCADE | SET NULL | SET DEFAULT }
```
I think the default is NO ACTION. You could alter back to default.

wwwy3y3 (commented):
others LGTM

wwwy3y3 (approved):

PR #1082, opened 2025-01-02
feat(wren-ai-service): Embed the SQL pairs in MDL

paopa:
## Changes
1. **Pipeline Simplification**
   - Removed SQL intention generation using LLM
   - Simplified document storage to use direct question-SQL pairs
   - Updated pipeline configuration to remove unnecessary LLM dependency

2. **Configuration Changes**
   - Added `SQL_PAIRS_PATH` environment variable
   - Added `pairs.json` configuration file support
   - Updated deployment configurations to include new file mounts

3. **Sample Management**
   - Added structured JSON format for storing SQL query examples
   - Updated sample format to use question-SQL pairs
   - Improved sample integration in SQL generation prompts

4. **Testing**
   - Updated existing tests to reflect new pipeline structure
   - Added test data for pairs.json
   - Temporarily skipped outdated tests pending updates

## Related Changes
- Updated docker compose configurations
- Modified kustomization files for deployment
- Updated configuration examples
- Added new utility tool for MDL string conversion

## Release Notes

- **New Features**
  - Added a new SQL pairs indexing pipeline for processing and storing SQL query examples
  - Introduced a tool to convert Model Definition Language (MDL) to string format
  - Enhanced SQL generation with support for sample query examples

- **Changes**
  - Renamed `sql_pairs_preparation` pipeline to `sql_pairs_indexing`
  - Updated configuration files to reflect pipeline changes
  - Removed LLM-specific configuration from SQL pairs pipeline

- **Improvements**
  - Streamlined SQL pairs handling across multiple service components
  - Added more flexible configuration for SQL pair processing

- **Testing**
  - Updated test cases to align with new SQL pairs indexing pipeline
  - Added sample SQL pairs test data

PR #1109, opened 2025-01-14
chore(wren-ai-service): add missing pipeline def

cyyeh:


paopa (commented):
we renamed the pipeline name to `sql_pairs_indexing` in #1082.

PR #1324, opened 2025-02-22
chore(wren-ai-service): allow regenerate sql using retrieved tables

cyyeh:
`POST /v1/ask-feedbacks`: now users need to pass parameter `tables` (type: [string], which means table names) to regenerate sql.

## Summary by CodeRabbit

- **New Features**
  - Introduced a comma-separated text input for specifying table names, giving users direct control over SQL query regeneration.
  - Integrated database schema context into the SQL generation process for more accurate query outputs.
  - Enabled feedback requests to include specific table details for improved context.

- **Improvements**
  - Updated the feedback status indicator to display “searching” for clearer progress communication.
  - Enhanced document retrieval with improved filtering, leading to a more reliable experience.

### DIFF
--- a/wren-ai-service/src/providers/document_store/qdrant.py
+++ b/wren-ai-service/src/providers/document_store/qdrant.py
@@ -209,6 +209,36 @@ class AsyncQdrantDocumentStore(QdrantDocumentStore):
                 document.score = score
         return results
 
+    async def _query_by_filters(
+        self,
+        filters: Optional[Dict[str, Any]] = None,
+        top_k: Optional[int] = None,
+    ) -> List[Document]:
+        qdrant_filters = convert_filters_to_qdrant(filters)
+        points_list = []
+        offset = None
+        while True:
+            points = await self.async_client.scroll(
+                collection_name=self.index,
+                offset=offset,
+                scroll_filter=qdrant_filters,
+                limit=top_k,
+            )
+            points_list.extend(points[0])
+            if points[1] is None:
+                break
+            offset = points[1]
+
+        if points_list:
+            return [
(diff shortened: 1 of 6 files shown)


wren-ai-service/src/providers/document_store/qdrant.py
```diff
@@ -209,6 +209,36 @@ async def _query_by_embedding(
...
+
+        if points_list:
+            return [
+                convert_qdrant_point_to_haystack_document(
+                    point, use_sparse_embeddings=self.use_sparse_embeddings
+                )
+                for point in points_list
+            ]
+        else:
+            return []
```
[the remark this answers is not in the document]
cyyeh: skipped

wren-ai-service/src/providers/document_store/qdrant.py
```diff
@@ -209,6 +209,36 @@ async def _query_by_embedding(
...
+            points = await self.async_client.scroll(
+                collection_name=self.index,
+                offset=offset,
+                scroll_filter=qdrant_filters,
+                limit=top_k,
+            )
+            points_list.extend(points[0])
+            if points[1] is None:
+                break
+            offset = points[1]
```
paopa, line 230: I executed the code and found that None is present at index 1, allowing it to satisfy the condition and exit the while loop. However, I suggest we clarify this in the docstring for future reference.

paopa: Additionally, the while loop w/ an always True condition poses a potential risk of an infinite loop. we are better to change the condition from my opinion.

paopa (approved):
overall lgtm. having comments for advice.

PR #1338, opened 2025-02-25
fix(wren-ui): sql server ssl property default value replace with initial value

andreashimin:
## Description
This PR fixes the issue where the default value of SQL Server SSL properties does not work because Antd Form treats initialValue as the actual value.

## Summary by CodeRabbit

- **Bug Fixes**
  - Adjusted the default behavior of the "Enable Trust Server Certificate" option so it now initializes as enabled consistently, ensuring that the toggle accurately reflects its intended starting state for a clearer user experience.


fredalai (approved):
LGTM. Thanks for fixing this issue.

PR #1407, opened 2025-03-13
chore(wren-ai-service): add followup sql generation reasoning

cyyeh:
## Summary by CodeRabbit

- **New Features**
	- Introduced a follow-up SQL generation capability that enhances query responses when historical context is provided.
	- Updated configurations across various environments to support the new follow-up SQL generation reasoning, enabling more contextually aware SQL query generation.


paopa (approved):
Overall lgtm. Can you also add the pipeline setup into other example configs like Ollama, Groq, etc.?

PR #1488, opened 2025-03-28
fix(wren-ui): separate instant recommend questions to new effect

andreashimin:
## Description
fix generate instant recommended question call multiple times triggered by incorrect dependency

## How to reproduce
- ask `Show me database schema` in prompt input
- wait for generate instant recomended question

## Summary by CodeRabbit

- **Refactor**
	- Optimized internal update processes to reduce unnecessary reloads, which can lead to a more efficient and responsive experience.


fredalai (approved):
LGTM

PR #1553, opened 2025-04-15
feat(wren-ui): Import from data source SQL

andreashimin:
## Description
This PR introduces a SQL dialect conversion feature that allows users to import SQL written in their data source’s dialect and manually convert it to Wren SQL when adding a question-SQL pair.

It also adds autocomplete support to the SQL editor when editing Wren SQL, improving the editing experience with keyword and schema suggestions.

Support autocomplete:
- Fix SQL editor
- Adjust SQL editor
- Question-SQL pair editor

Not support autocomplete:
- Import from data source SQL editor (it's not writing in Wren SQL)

### Remarks
Importing from data source SQL currently does not support the DuckDB data source, including sample datasets based on DuckDB.

## Screenshots
- Show toolbar in question-SQL pair
image

- Import from data source SQL modal
image

- Error in data source SQL
image

- Autocomplete in SQL editor
image


wren-ui/src/hooks/useAutoComplete.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+// For ace completer usage
+export const convertCompleter = (
+  item: (Model | Field) & { parent?: Model },
+) => {
+  return {
+    caption: item.parent
+      ? `${item.parent.displayName}.${item.displayName}`
+      : item.displayName,
+    value: item.referenceName,
```
fredalai, line 58: I'm thinking...do we need to consider wrapping column or table names in double quotes if they start with a number, contain special characters, rare reserved keywords?

fredalai (approved):
LGTM, except one question.

wren-ui/src/hooks/useAutoComplete.tsx
andreashimin: added in https://github.com/Canner/WrenAI/pull/1553/commits/73fa0882c50407b208dd4b8e02be283307370d80

fredalai (approved):

PR #1570, opened 2025-04-18
Fix: Correct provider name in config from openai to lm_studio

XavierMoon:
PR Description
This PR fixes a typo in the config.lm_studio.yaml file where the provider was incorrectly set to openai instead of lm_studio. This caused local LLM services (such as LM Studio) to be unrecognized and unusable. Changes
Corrected the provider field from openai to lm_studio in the config file Ensured compatibility with local LLM services

## Summary by CodeRabbit

- **Documentation**
  - Updated configuration examples to use the "lm_studio/" model provider prefix instead of "openai/" for both language and embedding models. Comments were revised to match the new format.


cyyeh:
@XavierMoon  this is not type. openai prefix means we use openai compatible api for lm studio

PR #1599, opened 2025-04-25
feat(wren-ui): Integrate dashboard cache APIs

andreashimin:
## Description
This PR integrates dashboard cache-related APIs with the UI.


fredalai (commented):
I have a question: I set a cache refresh time, but after clicking Refresh cache, the last refresh time is not displayed on the chart. What am I missing?

andreashimin:
> I have a question: I set a cache refresh time, but after clicking Refresh cache, the last refresh time is not displayed on the chart. What am I missing?

Hi @fredalai 
As previously discussed,
The cache settings are hidden when using a sample dataset or DuckDB. As you seen, if caching is not supported, the backend will not return the last refresh time. Check this https://github.com/Canner/WrenAI/pull/1599/commits/73bb8e88570bed289c550e7b556f9d7866c1bf5d, thanks

fredalai (approved):
LGTM

PR #1693, opened 2025-05-27
feat(wren-ai-service): add context window size handling in LLMProvider and related components

yichieh-lu:
This PR integrates the `context_window_size` parameter from `LLMProvider` into the token limit checking logic. This ensures that prompt + context do not exceed the model’s actual maximum context length.


cyyeh (approved):
lgtm, thanks for contribution

PR #1984, opened 2025-10-03
fix(dbt): add quote for the condition of relationship

goldmedal:
- Always add the quotes for the identifier to avoid the case-sensitive issue and Unique character.
- To avoid returning `nil`, create an empty slice for the relationship.


douenergy (approved):

douenergy:
Thanks @goldmedal

PR #2201, opened 2026-04-25
fix(wren-ai-service): expose batch_size in LitellmEmbedderProvider to cap embedding API call size

### LINKED ISSUE #2031: Bug Report: column_indexing_batch_size Not Taking Effect with OpenAI-Compatible Embedding Model
#### Environment:
- WrenAI version: [canner/wren-ai-service:0.27.14]
- Embedding model: OpenAI-compatible model with a maximum batch size limit of 10.
- Configuration: Set `column_indexing_batch_size: 10` in `config.yaml`.

#### Description:
When using an OpenAI-compatible Embedding model that enforces a maximum batch size of 10, the `column_indexing_batch_size` parameter set to 10 in `config.yaml` does not appear to take effect. During operation, the container logs the following error, indicating that the batch size being sent exceeds the model's limit of 10, despite the explicit configuration:

```
litellm.llms.openai.common_utils.OpenAIError: Error code: 400 - {'error': {'message': '<400> InternalError.Algo.InvalidParameter: Value error, batch size is invalid, it should not be larger than 10.: input.contents', 'type': 'InvalidParameter', 'param': None, 'code': 'InvalidParameter'}, 'id': '2354f194-0475-45e1-ada8-95757467f96a', 'request_id': '2354f194-0475-45e1-ada8-95757467f96a'}
```

This suggests that the batch size configuration is either being ignored or overridden internally, leading to invalid requests to the embedding model.

#### Steps to Reproduce:
1. Configure `column_indexing_batch_size: 10` in `config.yaml`.
2. Use an OpenAI-compatible Embedding model with a max batch size of 10.
3. Run the indexing process or relevant operation.
4. Observe the error in container logs.

#### Expected Behavior:
The batch size should respect the configured value of 10 and not exceed the model's limit, preventing the 400 error.

#### Actual Behavior:
The error occurs as if the batch size is larger than 10, ignoring the config setting.

#### Additional Notes:
- This issue persists even with the exact max batch size configured.
- Please investigate if there's an internal default or override affecting this parameter.

octo-patch:
Fixes #2031

## Problem

Users whose embedding providers enforce a maximum batch size (e.g. max 10 documents per API call) had no way to configure the embedding API batch size from `config.yaml`.

The internal `AsyncDocumentEmbedder.batch_size` (which controls how many texts are sent in each call to `aembedding()`) defaulted to 32 and was not exposed by `LitellmEmbedderProvider`. Setting `column_indexing_batch_size` in the `settings` section has a different effect — it controls how many columns are grouped into a single DDL schema chunk, not how many documents are batched for the embedding API call.

As a result, users received `400` errors like:
```
litellm.llms.openai.common_utils.OpenAIError: Error code: 400 - batch size is invalid, it should not be larger than 10
```

## Solution

Add `batch_size` as an explicit named parameter to `LitellmEmbedderProvider.__init__()` (defaulting to `32`) and forward it explicitly to `AsyncDocumentEmbedder` in `get_document_embedder()`.

Users can now set it per-model in the `embedder` section of `config.yaml`:

```yaml
type: embedder
provider: litellm_embedder
models:
  - model: openai/text-embedding-v4
    alias: default
    batch_size: 10   # limit API call size for providers with batch restrictions
```

Also added a commented example in `config.qwen3.yaml` to document the option.

## Testing

- Existing tests pass unchanged (the default behavior is preserved).
- The new parameter propagates through the provider → document embedder chain, capping each `aembedding()` call to at most `batch_size` documents.

## Release Notes

* **New Features**
  - Embedder providers now support a configurable batch size parameter for controlling per-call document batching, useful when embedding providers enforce batch size limits.

* **Documentation**
  - Updated embedder configuration examples with documentation for the optional batch size setting and guidance on adjusting it per your provider's requirements.

### DIFF
--- a/wren-ai-service/docs/config_examples/config.qwen3.yaml
+++ b/wren-ai-service/docs/config_examples/config.qwen3.yaml
@@ -66,6 +66,9 @@ models:
     alias: default
     api_base: https://api.openai.com/v1  # change this according to your embedding model
     timeout: 120
+    # batch_size controls how many documents are sent per embedding API call (default: 32).
+    # Lower this value if your embedding provider enforces a maximum batch size.
+    # batch_size: 32
 
 ---
 type: engine
--- a/wren-ai-service/src/pipelines/indexing/db_schema.py
+++ b/wren-ai-service/src/pipelines/indexing/db_schema.py
@@ -1,4 +1,3 @@
-import asyncio
 import logging
 import sys
 import uuid
@@ -29,7 +28,7 @@ logger = logging.getLogger("wren-ai-service")
 @component
 class DDLChunker:
     @component.output_types(documents=List[Document])
-    async def run(
+    def run(
         self,
         mdl: Dict[str, Any],
         column_batch_size: int,
@@ -48,7 +47,7 @@ class DDLChunker:
                 },
                 "content": chunk["payload"],
             }
-            for chunk in await self._get_ddl_commands(
+            for chunk in self._get_ddl_commands(
                 **mdl, column_batch_size=column_batch_size
             )
         ]
@@ -63,7 +62,7 @@ class DDLChunker:
             ]
         }
 
-    async def _model_preprocessor(
+    def _model_preprocessor(
         self, models: List[Dict[str, Any]], **kwargs
     ) -> List[Dict[str, Any]]:
         def _column_preprocessor(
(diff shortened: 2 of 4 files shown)


AmirF194 (commented):
I checked this PR's current mergeability against main.

wren-ai-service, wren-ui, wren-launcher, docker and deployment were archived off main to the legacy/v1 branch in the 2026-05-04 repo consolidation (commit e5b25eec, PR #2210, five days after this PR was opened). None of the three files this PR touches (wren-ai-service/src/providers/embedder/litellm.py, wren-ai-service/src/pipelines/indexing/db_schema.py, wren-ai-service/docs/config_examples/config.qwen3.yaml) exist on main anymore.

```
$ gh api repos/Canner/WrenAI/pulls/2201 --jq '{mergeable,mergeable_state}'
{"mergeable":false,"mergeable_state":"dirty"}
$ gh api repos/Canner/WrenAI/contents/wren-ai-service
{"message":"Not Found", ...}
```

Rebasing onto main would not resolve this the way a normal conflict rebase would, since the target directory is gone rather than just changed. If the batch_size fix is still wanted, it would need to move to whatever repo/branch now owns the GenBI service (legacy/v1, or a successor if one exists).

Separately, this branch also carries two commits (166e4af7, 33999d1a) that are identical to PR #2194, still open. Worth knowing if #2201 is meant to supersede it.

PR #2251, opened 2026-05-08
feat(context): bind a connection profile to a project

PaulChen79:
## Summary

- New `wren context set-profile <name>` command writes both `profile:` and `data_source:` into `wren_project.yml`, locking a project to a specific connection profile
- `dry-plan` / `dry-run` now consult the project's pinned profile before falling back to the global active one — multi-project setups stop accidentally hitting the wrong DB
- `wren context validate` hints when no profile is bound and warns when the bound profile has been removed
- onboarding skill + connect / quickstart / installation docs reordered to profile-first so users leave onboarding with a fully bound project

## Why

Today, the CLI's only profile mechanism is the global `active:` field in `~/.wren/profiles.yml`. That's fine for one-project-per-machine, but it breaks the moment you have two Wren projects pointing at different databases — `wren profile switch` becomes a footgun and the SDK already supports `profile:` in `wren_project.yml` while the CLI ignores it. This PR closes that gap with a single new command (`set-profile`) and a small resolver helper, so the project↔profile mapping is now an explicit first-class fact in the project file.

The SDK's existing 3-tier resolution (kwarg → project pin → active) is already aligned with the new CLI behavior — no SDK changes needed.

## Design decisions (locked during planning)

- **Minimal surface area**: only `set-profile` is new. `init`, `profile add`, and `profile switch` are unchanged.
- **`set-profile` semantics**: always overwrites `data_source` from the bound profile (single source of truth at bind time). Mismatched MDL is surfaced later by `validate` / `build` rather than blocking at bind.
- **Onboarding ordering**: `mkdir → fill .env → profile add → context init → set-profile → MDL gen` — profile-first so the new step slots in cleanly after profile validation.
- **`validate` hint tone**: info (not warning) when no profile bound, so error output stays uncluttered for projects with real problems.

## Commit breakdown

| Commit | Purpose |
|---|---|
| `feat(profile): add resolve_profile_for_project helper` | Pure addition — 3-tier resolver, no call sites |
| `refactor(cli): honor project-pinned profile in dry-plan and dry-run` | Wire the resolver into existing CLI sites; behavior-equivalent for unpinned projects |
| `feat(context): add wren context set-profile command` | The new command + `save_project_config()` helper |
| `feat(context): hint about set-profile in validate when no profile bound` | UX nudge + missing-pin warning |
| `docs(skill): reorder wren-onboarding to profile-first + add set-profile step` | Skill version 2.0 → 2.1 |
| `docs: document project↔profile binding in connect.md` | New Step 5 in the connect guide |
| `docs: align installation and quickstart with profile-first onboarding` | Loose ends in the other onboarding-adjacent docs |

## Test plan

- [x] 22 new unit tests covering: resolver fallback, pinned-profile-missing error, set-profile first-bind, set-profile rebind, set-profile preserves other fields, validate hints, validate strict on missing pin
- [x] `pytest core/wren/tests/` — 364 passed, 1 pre-existing failure (`test_validate_strict_warns`) unrelated to this PR
- [x] `ruff check` clean on every changed source/test file
- [x] Manually walked the new onboarding flow in an isolated `WREN_HOME` (set-profile binds, validate goes silent, dry-plan honors the pin)
- [ ] Reviewer: skim the connect.md / quickstart.md diffs for tone/clarity
- [x] Reviewer: confirm the SKILL.md ordering reads well end-to-end (Step 2 → 3 → 3.5 → 3.6 → 4)

## Out of scope (deliberate)

- `init --profile X` flag — onboarding flow now creates the profile *before* init runs, so the flag has no use case
- `profile add` cwd detection — `set-profile` already covers post-add binding without polluting the global command
- `profile switch` warning — keeps the global command pure
- SDK changes — the SDK was already reading `profile:` correctly
- Eager connection validation in `WrenToolkit.from_project()` — separate SDK improvement, tracked separately

## Summary by CodeRabbit

* **New Features**
  * Added `wren context set-profile` to bind a connection profile to a project; CLI now prefers a project-bound profile for project-aware commands.

* **Bug Fixes / Improvements**
  * CLI commands (including `dry-plan`) use project-bound profiles when present.
  * Validation now warns on missing/mismatched pinned profiles.
  * Stable, round-trip-friendly saving of project config.

* **Documentation**
  * Updated connect, installation, and quickstart docs; onboarding skill bumped to v2.1 and expanded to cover .env and binding.

* **Tests**
  * Added unit and integration tests for profile resolution and binding behaviors.

### DIFF
--- a/core/wren/src/wren/cli.py
+++ b/core/wren/src/wren/cli.py
@@ -464,7 +464,11 @@ def dry_plan(
         if prof_dict:
             prof_ds = prof_dict.pop("datasource", None)
             if prof_ds is None:
-                typer.echo("Error: no datasource in active profile.", err=True)
+                typer.echo(
+                    "Error: no datasource in resolved profile "
+                    "(project-pinned or active).",
+                    err=True,
+                )
                 raise typer.Exit(1)
             try:
                 ds = DataSource(prof_ds.lower())
--- a/core/wren/src/wren/cli.py
+++ b/core/wren/src/wren/cli.py
@@ -127,6 +127,44 @@ def _load_conn(
     return {}
 
 
+def _resolve_engine_profile(mdl: str | None) -> tuple[str | None, dict]:
+    """Resolve the connection profile for project-context CLI commands.
+
+    Tries to find the project root from ``--mdl`` (when it points to a real
+    file) or the current working directory; if a project is found, defers
+    to :func:`wren.profile.resolve_profile_for_project` (which honors any
+    ``profile:`` pin in ``wren_project.yml``). Otherwise falls back to the
+    global active profile.
+
+    The fall-through preserves backward compatibility: callers that pass
+    ``--mdl <base64-string>`` or run outside a project see the same
+    ``get_active_profile()`` behavior as before.
+    """
+    from wren.profile import (  # noqa: PLC0415
+        get_active_profile,
+        resolve_profile_for_project,
+    )
+
+    project_path: Path | None = None
+    if mdl is not None:
+        mdl_path = Path(mdl).expanduser()
+        if mdl_path.exists() and mdl_path.is_file():
+            candidate = mdl_path.resolve().parent.parent
+            if (candidate / "wren_project.yml").exists():
+                project_path = candidate
+    else:
+        try:
+            from wren.context import discover_project_path  # noqa: PLC0415
+
+            project_path = discover_project_path()
+        except SystemExit:
+            project_path = None
+
+    if project_path is None:
+        return get_active_profile()
+    return resolve_profile_for_project(project_path)
+
+
 def _resolve_datasource(conn_dict: dict, explicit: str | None = None) -> str:
     """Return datasource from explicit arg or connection dict.
 
@@ -160,15 +198,15 @@ def _build_engine(
 
     manifest_str = _load_manifest(_require_mdl(mdl))
 
-    # Try active profile when no explicit connection flags given
+    # Try project-pinned profile (or fall back to active) when no explicit
+    # connection flags given.
     if not connection_info and not connection_file:
         from wren.profile import (  # noqa: PLC0415
             MissingSecretError,
             expand_profile_secrets,
-            get_active_profile,
         )
 
-        prof_name, prof_dict = get_active_profile()
+        prof_name, prof_dict = _resolve_engine_profile(mdl)
         if prof_dict:
             prof_ds = prof_dict.pop("datasource", None)
             ds_str = datasource or prof_ds
@@ -419,11 +457,10 @@ def dry_plan(
 
     manifest_str = _load_manifest(_require_mdl(mdl))
 
-    # Try active profile when no explicit flags given
+    # Try project-pinned profile (or fall back to active) when no explicit
+    # connection flags given.
     if datasource is None and connection_file is None:
-        from wren.profile import get_active_profile  # noqa: PLC0415
-
-        _prof_name, prof_dict = get_active_profile()
+        _prof_name, prof_dict = _resolve_engine_profile(mdl)
         if prof_dict:
             prof_ds = prof_dict.pop("datasource", None)
             if prof_ds is None:
--- a/core/wren/src/wren/cli.py
+++ b/core/wren/src/wren/cli.py
@@ -130,39 +130,50 @@ def _load_conn(
 def _resolve_engine_profile(mdl: str | None) -> tuple[str | None, dict]:
     """Resolve the connection profile for project-context CLI commands.
 
-    Tries to find the project root from ``--mdl`` (when it points to a real
-    file) or the current working directory; if a project is found, defers
-    to :func:`wren.profile.resolve_profile_for_project` (which honors any
-    ``profile:`` pin in ``wren_project.yml``). Otherwise falls back to the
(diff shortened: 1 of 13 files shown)


goldmedal (commented):
Solid PR overall — the project-pin design closes a real footgun in multi-project setups, and CodeRabbit's earlier findings are all addressed in commits 9–15. CI is green across lint/unit/memory/mysql/postgres/skill-version-parity. A few notes before merge, one of which I'd consider a follow-up worth pursuing.

## 1. `_resolve_engine_profile` couples `--mdl` parsing with project detection (medium)

`core/wren/src/wren/cli.py:148-161`: the resolver uses `--mdl`'s path as the primary project-detection signal and only walks cwd when `--mdl` is absent. This re-introduces the same class of footgun the PR is trying to close.

**Current behavior matrix:**

| `--mdl` | cwd in project A | Result |
|---|---|---|
| `/tmp/external.json` (not in any project) | yes | global active ⚠️ — A's pin silently bypassed |
| `<base64-string>` | yes | global active ⚠️ — same bypass |
| `/other-proj/target/mdl.json` | yes | other-proj pin |
| (omitted) | yes | A's pin |

The first two rows are exactly the kind of "wrong DB silently chosen" problem this PR is designed to eliminate — just triggered by `--mdl` shape instead of by a stale global active.

**Suggested rewrite — decouple, then fall back only when no project context exists at all:**

```python
def _resolve_engine_profile(mdl: str | None) -> tuple[str | None, dict]:
    project_path = _discover_project_for_engine(mdl)
    if project_path is None:
        return get_active_profile()
    return resolve_profile_for_project(project_path)


def _discover_project_for_engine(mdl: str | None) -> Path | None:
    # 1. If --mdl is a real file, walk up looking for wren_project.yml.
    #    Don't hard-code parent.parent — `target/mdl.json` is the build
    #    default, not a contract. Users may run `--mdl <project>/build/mdl.json`.
    if mdl is not None:
        mdl_path = Path(mdl).expanduser()
        if mdl_path.exists() and mdl_path.is_file():
            for parent in mdl_path.resolve().parents:
                if (parent / "wren_project.yml").exists():
                    return parent
                if parent == Path.home() or parent == parent.parent:
                    break
    # 2. Always try cwd discovery, regardless of whether --mdl was given.
    try:
        return discover_project_path()
    except SystemExit:
        return None
```

Two changes in one:

- **Walk up from `--mdl`** instead of hard-coding `parent.parent` — the `<project>/target/mdl.json` layout is a build default, not a contract.
- **Always fall through to cwd discovery** when `--mdl` doesn't carry project context. `--mdl` becomes a pure "which MDL artifact to load" override; project (and therefore profile) is determined independently. Active profile is reserved for "no project context anywhere" — its actual fallback role.

This also enables a coherent use case the current code blocks: running `wren dry-plan --mdl /tmp/experiment.json` from inside project A to test an external MDL against A's connection, without `wren profile switch` first.

If you'd rather keep the current "self-contained `--mdl`" semantics, please at least print a stderr line when falling back to active despite cwd being in a project — silent bypass is the worst of both worlds.

## 2. No bind-time hint when re-binding overwrites `data_source` and MDL is already built (low/medium)

`core/wren/src/wren/context_cli.py:619-630`: when `old_ds != new_ds`, the success line shows `data_source: postgres -> duckdb`, but there's no warning that an existing `target/mdl.json` was built for the old dialect. `docs/core/get_started/connect.md` does call this out, but the user only sees the doc if they read it; the CLI is silent.

Cheap UX win: when `old_ds != new_ds` AND `(project_path / "target" / "mdl.json").exists()`, append:

```
⚠ MDL was built for <old_ds>. Run `wren context build` to regenerate before querying.
```

## 3. `set_profile` hardcodes the project filename (nit)

`core/wren/src/wren/context_cli.py:524 / :589`: the literal `"wren_project.yml"` is duplicated instead of using `wren.context._PROJECT_FILE`. Either expose the constant (drop the underscore) or accept the duplication — current state is mildly inconsistent with the rest of `context.py`.

## 4. `save_project_config` round-trip not tested with custom fields (nit)

`core/wren/src/wren/context.py:432-449`: the field-reorder logic correctly appends unknown keys at the end, but `test_set_profile_preserves_other_fields` only verifies the standard fields (`name`, `catalog`, `schema`, `schema_version`). One round-trip with a custom field (e.g. `tags: [foo, bar]`) would lock the "extra keys are preserved" contract — cheap insurance against a future re-shuffle of `_PROJECT_FIELD_ORDER` accidentally dropping fields.

## 5. No-pin info hint suppressed when warnings exist (intentional, but worth confirming)

`core/wren/src/wren/context_cli.py:344-359`: the *"no profile bound"* hint only fires when validation has zero errors AND zero warnings. The comment says this is to keep error output uncluttered — fine if the goal is "don't pile noise on warnings," but it means a project with description warnings + no profile pin never sees the hint. Tests don't catch this because they assert on clean projects.

Worth a deliberate call: should the hint be unconditional once the structural check passes? Or always after the warning summary, on its own line so it's clearly scoped? Whichever you pick, please add a one-line comment so the next reader doesn't think it's an oversight.

---

Items #1 and #2 I'd address before merge; #3–#5 are fine as follow-ups.

goldmedal (approved):
Thanks @PaulChen79 Look good 👍

PR #2258, opened 2026-05-12
feat(wren): add YTsaurus (CHYT) connector

### LINKED ISSUE #2257: feat(wren): add YTsaurus (CHYT) connector
## Feature request

Add a first-class data source for [YTsaurus](https://ytsaurus.tech/en), an open-source distributed storage and execution system, via its **CHYT** (ClickHouse-over-YT) clique.

## Why

YTsaurus is the storage and compute backbone for petabyte-scale analytics at several organizations and is a common destination for Wren-style semantic SQL workloads. Today, users running on YT have no first-class Wren connector and have to either:

- Bypass Wren entirely and query CHYT directly (losing MDL semantics, dry-run validation, denied-functions policy, memory module, text-to-SQL surface), or
- Misconfigure the `clickhouse` connector against the YT HTTP proxy — which fails because CHYT mounts at a non-root URL path, routes by a URL parameter (`chyt.clique_alias`) rather than the ClickHouse `database` field, and accepts only `Authorization: OAuth <token>` (it rejects Basic and Bearer).

The CHYT clique exposes a ClickHouse-compatible HTTP protocol, so the bulk of Wren's existing ClickHouse / Ibis / sqlglot machinery applies unchanged — only the auth/routing shim is new.

## Proposed scope

- New `DataSource.ytsaurus` enum value and factory entry
- New `YTsaurusConnectionInfo` Pydantic model: `proxy`, `clique`, `token` (with `YT_TOKEN` env fallback), `secure`, `port`, `query_path`, `settings`, `kwargs`
- New `wren.connector.ytsaurus` module (subclass of the Ibis ClickHouse connector with YT-specific auth header injection, clique-alias URL param patching, and a CHYT-friendly `query` / `dry_run` path that bypasses ibis' `CREATE VIEW`-based introspection)
- New pip extra: `wren-engine[ytsaurus]`
- sqlglot dialect map: `ytsaurus -> clickhouse`
- Optional MDL-level rewrite: when `data_source == ytsaurus`, replace `<schema>.<table>` references in the physical SQL with backticked YT paths sourced from each model's `properties.ytPath`
- Docs: `core/wren/docs/connectors/ytsaurus.md`, `README` install row, `docs/connections.md` JSON example

## Out of scope

- Native YQL (Query Tracker) — CHYT only; users who need raw YQL can fork the connector
- SPYT (Spark-on-YT)

## Implementation

PR forthcoming, will reference this issue.

nar3k:
Closes #2257.

## Summary

Adds a first-class `ytsaurus` data source that talks to a [YTsaurus](https://ytsaurus.tech/en) cluster through its CHYT (ClickHouse-over-YT) clique. CHYT exposes a ClickHouse-compatible HTTP protocol on the YT proxy, so the bulk of Wren's existing ClickHouse / Ibis / sqlglot path is reused — only the auth and routing shim is new.

## What's in scope

- `DataSource.ytsaurus` enum + factory entry + sqlglot dialect map (`ytsaurus -> clickhouse`)
- `YTsaurusConnectionInfo` Pydantic model (`proxy`, `clique`, `token` w/ `YT_TOKEN` env fallback, `secure`, `port`, `query_path`, `settings`, `kwargs`)
- `wren.connector.ytsaurus.YTsaurusConnector` — subclass of the Ibis ClickHouse connector
- `wren-engine[ytsaurus]` pip extra (pulls `ibis-framework[clickhouse]`)
- Optional MDL-level physical-SQL rewrite in `engine.py`: when `data_source == ytsaurus`, `<schema>.<table>` references are replaced with backticked YT paths sourced from each model's `properties.ytPath`. Path map is cached; gated entirely on `data_source == ytsaurus`, so other connectors are unaffected
- Docs: `core/wren/docs/connectors/ytsaurus.md`, `README` install row, `docs/connections.md` JSON example

## CHYT-specific details (why this is more than a ClickHouse alias)

1. `clickhouse_connect.get_client()` does not accept `http_headers` as a kwarg → headers are injected post-construction, with a belt-and-braces fallback for clients that re-snapshot state.
2. CHYT mounts at URL path `/query`, not `/` → new `query_path` field (default `/query`) maps to `clickhouse_connect`'s `proxy_path`.
3. YT routes by `chyt.clique_alias=<*alias>` URL param, not by `database=` → the connector patches `HttpClient.params` at construction so the alias is present before the startup `SELECT version()` runs.
4. YT proxy accepts only `Authorization: OAuth <token>` (rejects Basic / Bearer) → the connector subclasses `HttpClient` with a one-line override of `_init_common_settings` that injects the OAuth header before the parent's startup query.
5. ibis `backend.sql()` introspects via `CREATE VIEW`, which CHYT rejects with `std::out_of_range` → `YTsaurusConnector.query` / `dry_run` bypass ibis and call `clickhouse_connect` directly (native protocol → `pyarrow.Table`; `EXPLAIN AST` for dry-runs).

## Test plan

- [x] `wren context build` and `wren --sql ...` against a live YT cluster with a CHYT clique
- [x] `COUNT(*)`, JOIN-and-aggregate, and dialect-specific functions (`toUnixTimestamp`, `startsWith`, `INTERVAL` arithmetic, `COUNT(DISTINCT ...)`) all return matching results vs. the same SQL run against the cluster directly
- [x] `connector.dry_run(sql)` returns the CHYT `EXPLAIN AST` without executing
- [x] `cargo` / `ruff` / existing `wren` test suite untouched — no other connector paths altered
- [ ] Reviewers: please flag any preferred test fixtures for a CHYT-against-mock test; happy to add a `clickhouse_connect`-mock-based unit test if useful

## Out of scope

- Native YQL (Query Tracker) — CHYT only
- SPYT (Spark-on-YT)

## Summary by CodeRabbit

* **New Features**
  * YTsaurus (CHYT) added as a supported data source for queries, including connector and engine integration.
  * OAuth token support with optional YT_TOKEN environment fallback.
  * Statement-timeout handling for YTsaurus queries.
  * SQL rewriting to resolve YTsaurus-backed tables to YT paths at runtime.

* **Documentation**
  * New installation, connector, and connection-format docs with examples and field references.

* **Chores**
  * New optional install extra to ease YTsaurus dependency installation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Canner/WrenAI/pull/2258)

### DIFF
--- a/core/wren/src/wren/model/data_source.py
+++ b/core/wren/src/wren/model/data_source.py
@@ -409,8 +409,10 @@ class DataSourceExtension(Enum):
         # injects the OAuth header into `self.headers` right before the
         # parent `Client.__init__` runs the startup queries.
         try:
-            import clickhouse_connect.driver as _ch_driver
-            import clickhouse_connect.driver.httpclient as _ch_http
+            # Deferred: clickhouse_connect is an optional dependency installed
+            # only with the [ytsaurus] / [clickhouse] extras.
+            import clickhouse_connect.driver as _ch_driver  # noqa: PLC0415
+            import clickhouse_connect.driver.httpclient as _ch_http  # noqa: PLC0415
         except ImportError as e:
             raise WrenError(
                 ErrorCode.INVALID_CONNECTION_INFO,
--- a/core/wren/src/wren/model/data_source.py
+++ b/core/wren/src/wren/model/data_source.py
@@ -130,6 +130,12 @@ class DataSource(StrEnum):
 
     def _build_connection_info(self, data: dict) -> ConnectionInfo:
         if "connectionUrl" in data or "connection_url" in data:
+            if self == DataSource.ytsaurus:
+                raise WrenError(
+                    ErrorCode.INVALID_CONNECTION_INFO,
+                    "YTsaurus does not support connectionUrl; "
+                    "use proxy/clique/token fields.",
+                )
             if self == DataSource.clickhouse:
                 return self._handle_clickhouse_url(
                     urllib.parse.urlparse(
--- a/core/wren/src/wren/model/data_source.py
+++ b/core/wren/src/wren/model/data_source.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import base64
+import os
 import ssl
 import urllib
 from enum import Enum, StrEnum, auto
@@ -39,6 +40,7 @@ from wren.model import (
     SparkConnectionInfo,
     SSLMode,
     TrinoConnectionInfo,
+    YTsaurusConnectionInfo,
 )
 from wren.model.error import ErrorCode, WrenError
 
@@ -66,6 +68,7 @@ class DataSource(StrEnum):
     duckdb = auto()
     spark = auto()
     databricks = auto()
+    ytsaurus = auto()
 
     def get_connection(self, info: ConnectionInfo) -> BaseBackend:
         try:
@@ -95,7 +98,7 @@ class DataSource(StrEnum):
                     options += f"-c statement_timeout={headers.get(X_WREN_DB_STATEMENT_TIMEOUT, 180)}s"
                     kwargs["options"] = options
                 info.kwargs = kwargs
-            case DataSource.clickhouse:
+            case DataSource.clickhouse | DataSource.ytsaurus:
                 session_timeout = headers.get(X_WREN_DB_STATEMENT_TIMEOUT, 180)
                 if info.settings is None:
                     info.settings = {}
@@ -175,6 +178,8 @@ class DataSource(StrEnum):
                 ):
                     return DatabricksServicePrincipalConnectionInfo.model_validate(data)
                 return DatabricksTokenConnectionInfo.model_validate(data)
+            case DataSource.ytsaurus:
+                return YTsaurusConnectionInfo.model_validate(data)
             case _:
                 raise NotImplementedError(f"Unsupported data source: {self}")
 
@@ -229,6 +234,7 @@ class DataSourceExtension(Enum):
     gcs_file = "gcs_file"
     databricks = "databricks"
     spark = "spark"
+    ytsaurus = "ytsaurus"
 
     def get_connection(self, info: ConnectionInfo) -> BaseBackend:
         try:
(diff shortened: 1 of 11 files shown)


nar3k:
Addressed all five CodeRabbit findings in b234b45:

| # | File | Issue | Fix |
|---|---|---|---|
| 1 | `connector/ytsaurus.py:53-55` | `limit` interpolated into f-string without runtime type check | Reject anything that isn't a non-negative `int` (also rejects `bool`) before building the wrapper SQL |
| 2 | `connector/ytsaurus.py:60-63` | `zip()` could silently truncate if `column_names` / `result_columns` disagree | Raise `WrenError(INVALID_SQL)` when lengths differ |
| 3 | `engine.py:172-189` | `_rewrite` could match an empty-key path map entry | Early-return the node when `tbl_name` is empty |
| 4 | `engine.py:209-221` | Path map could grow an empty-string key when a model has no resolvable `table` | `continue` on empty `table`; also drop the redundant `if table:` guard around `setdefault` |
| 5 | `model/data_source.py:372-423` | Class-level monkey-patching of `HttpClient.params` and `_wren_yt_token` raced under concurrency — could leak one caller's token into another's connection | Added module-level `_YTSAURUS_PATCH_LOCK` (`threading.Lock`) and wrapped the patch/connect/finally block in `with _YTSAURUS_PATCH_LOCK:` |

The lock is at module scope rather than on `DataSourceExtension` because the latter is an `Enum` and a plain `threading.Lock()` attribute would be treated as an enum member.

nar3k:
@coderabbitai review

nar3k:
@coderabbitai review

nar3k:
@coderabbitai review

goldmedal:
Thanks @nar3k, there are some CI failures.  Could you fix them before I review this PR?

core/wren/src/wren/model/data_source.py
```diff
@@ -316,15 +344,136 @@ def get_canner_connection(info: CannerConnectionInfo) -> BaseBackend:
...
+        with _YTSAURUS_PATCH_LOCK:
+            return ibis.clickhouse.connect(
+                host=info.host,
+                port=int(info.port),
+                database=info.database,
+                user=info.user,
+                password=(info.password and info.password.get_secret_value()),
+                settings=info.settings if info.settings else {},
+                **info.kwargs if info.kwargs else {},
+            )
```
goldmedal, line 362: I'm not familer with `ytsaurus`. Is it fully compatible with ClickHouse? Because we're planning to remove the `ibis-framework` dependency and use the native `clickhouse_connect` driver instead.
I think you can connect with `ytsaurus` through `clickhouse_connect` directly in this PR.

nar3k: Hi, i will check it out and get back!

goldmedal: refer to https://github.com/Canner/WrenAI/pull/2275
You can also wait for me to finish #2275, then follow the way to connect with `ytsaurus`.

goldmedal:
@nar3k, could you add some tests or document how to test YTSaurus?  It will be helpful to protect the follow-up change so that it won't break the YTsaurus connector.

core/wren/src/wren/model/data_source.py
goldmedal: BTW, #2313 is merged, which includes the change for the clickhouse (#2275).  I think you can check how we are using clickhouse_connect to connect with Clickhouse.

PR #2290, opened 2026-05-15
docs(wren-core-wasm): refresh onboarding for inline-mode default

goldmedal:
## Summary

- Bumps CDN version refs in README + AGENT_GUIDE from `@0.1.0` → `@0.3.0` (matches the currently published `@wrenai/wren-core-wasm` package).
- Reorders **Inline Mode first** and labels it as the recommended default for local dev and bundled dashboards. URL mode is now framed as "useful when data is large or already on a CDN" — surfaces the HTTP `Range:` requirement up front.
- Adds a **Choosing a local dev server** section with a range-support comparison table (built-in `python -m http.server` ❌, `npx serve` ✅, `RangeHTTPServer` ✅, etc.) plus a `curl -I -H "Range: ..."` smoke check.
- Adds a **Node.js usage** section showing the `wasmUrl: BufferSource` pattern — `WrenEngine.init()` without options fails immediately under Node because `file://` fetch isn't supported. Required for any `node --test` smoke checks against the published package.
- Expands AGENT_GUIDE **Common Pitfalls** with three entries: range-support requirement, Node `BufferSource`, and "register sequentially" (the engine is single-threaded; concurrent `registerParquet` corrupts state).

## Motivation

These changes were gathered from an end-to-end onboarding session where the agent and user hit four predictable, time-costing failure modes documented in this PR:

1. Following the docs literally pulls a 4-versions-stale package (`@0.1.0`).
2. URL mode is presented as "recommended" but the most common dev server (`python -m http.server`) silently breaks it — fetches stall after the Parquet footer with no error.
3. Smoke testing with Node fails on `init()` before any user code runs.
4. Parallel `registerParquet` calls corrupt the engine state.

Each item above maps to a concrete friction point that previously cost time during real onboarding.

## Test plan

- [x] `grep` shows no remaining `@0.1.0` references in the touched files.
- [ ] Manual: copy-paste the README inline-mode Quick Start into a new HTML file, serve with `npx serve`, and confirm the query returns rows.
- [ ] Manual: copy-paste the Node.js usage snippet into `node --test` and confirm `WrenEngine.init()` resolves with the BufferSource.
- [ ] Manual: run `curl -I -H "Range: bytes=0-1023"` against a Parquet served by `python -m http.server` to confirm the docs match observed behavior (no `206`).

## Out of scope

A companion plan (`IMPROVEMENT_PLAN_skill_and_docs.md` in the contributor's workspace) also called for:
- Adding a `cubeQuery`/`listCubes` API reference. Skipped here — those APIs are not yet on `main`. They should land with the feature, not ahead of it.
- Mentioning `registerCsv`. Same reason.
- Skill-side changes (in `~/.claude/skills/wren-dlt-connector/`). Those live outside this repo.

## Summary by CodeRabbit

* **Documentation**
  * Updated CDN examples to @wrenai/wren-core-wasm@0.3.0.
  * Made Inline Mode the recommended local/bundled workflow; clarified URL Mode for large datasets.
  * Specified HTTP Range and CORS requirements for URL Mode, added a range-support table and a curl check for 206 Partial Content.
  * Added Node.js guidance to pass WASM as a BufferSource to init().
  * Documented sequential registration pattern, load order, handling of empty source, and expanded common pitfalls.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Canner/WrenAI/pull/2290)


goldmedal:
Addressed @coderabbitai's review in 846702a3:

1. **🔴 Revert CDN refs to `@0.3.0`** — `@0.4.0` is on `main` (release-please bump in #2287) but isn't published to npm yet. Reverted all three CDN refs in README + AGENT_GUIDE to `@0.3.0`. Will bump again once `0.4.0` is actually live on unpkg.
2. **🟡 Dev-server table caveats** — expanded the Notes column in both README and AGENT_GUIDE:
   - `npx serve` and Vite → ⚠️ Single-range; both use `sirv` and can return `416` when the requested range extends past EOF.
   - `webpack-dev-server` → ⚠️ Single-range; multipart range requests fall back to returning the whole resource.
3. **💤 README `200` formatting** — backticked `\`200\`` to match AGENT_GUIDE.
4. **💤 `file://` wording** — rephrased the URL-mode CORS pitfall to "URL mode needs HTTP(S) + CORS — opening the dashboard from a `file://` URL won't work as the data source…", which is clearer than the original "won't work for fetching remote Parquet".

PaulChen79 (approved):

PR #2329, opened 2026-05-28
feat(wren): serve agent skills and reference docs from the CLI

PaulChen79:
## Summary

Move the source of truth for agent workflow guides, reference docs, and prompt templates **into the `wren` CLI**. Replace the previous "five fat skills installed into the agent's skill directory" model with a single ~50-line **discovery stub** plus three new content-delivery commands: `wren skills get/list`, `wren docs get/list`, `wren ask --guided/--direct`.

**Why:** The old model installed `~/.claude/skills/wren-*` markdown directly. That bundle drifted from the installed CLI (the markdown referenced commands and flags that didn't always match the user's `pip install`-ed version), and the agent loaded all five skills at session start whether it needed them or not. The new model ships the content inside the `wrenai` wheel itself, so the version an agent reads always matches the installed CLI, and the agent only pays for what it fetches.

Design discussion: `.tmp/prds/2026-05-17-agent-cli-additive-refactor-zh.md` (and English mirror). Modeled after `vercel-labs/agent-browser`'s single-stub pattern.

### CLI surfaces

```bash
# Workflow guides (5 bundled skills)
wren skills list
wren skills get onboarding              # entry point: setup + first query
wren skills get usage                   # day-to-day querying
wren skills get generate-mdl            # generate MDL from a DB schema
wren skills get dlt-connector           # SaaS via dlt → DuckDB → Wren project
wren skills get enrich-context          # business context (units, enums, cubes)
wren skills get <name> --full           # inline the skill's references/
wren skills get dlt-connector --script introspect_dlt   # fetch bundled script

# Reference docs (15 mirrored from docs/core/)
wren docs list
wren docs get <reference>               # connect, mdl, cubes, installation, quickstart, ...

# Prompt shaping (no execution — just renders to stdout)
wren ask "<question>" --guided          # strict task flow for weaker LLMs
wren ask "<question>" --direct          # minimal wrapping for stronger LLMs
```

`wren ask` has no default mode — the user must pick `--guided` or `--direct` (silently changing a default would alter agent behavior across an upgrade).

### Discovery stub

`skills/wren/SKILL.md` is now a ~50-line stub that teaches the agent to fetch everything else from the CLI. `npx skills add Canner/WrenAI` installs it.

[...]

PR #2348, opened 2026-06-08
feat(wren): GenBI app build & deploy — semantic layer → shareable web app

PaulChen79:
## What

Adds a **GenBI app build & deploy** pipeline to the `wren` CLI: turn a Wren project's semantic layer into a shareable, browser-side GenBI web app and deploy it to the user's Vercel or Cloudflare account — from a natural-language request to a public URL in one conversation.

## Design: CLI ↔ agent split

- **The CLI owns all deterministic state** — the authoritative build instruction, the app index, verify, and deploy.
- **The agent authors the app code**, following the CLI's build instruction. `.wren/apps.yml` is only ever written by the CLI, never hand-edited.

## New CLI surface (`wren genbi`)

| Command | Purpose |
| --- | --- |
| `wren genbi build <name> --prompt "…" [--data-mode snapshot\|live]` | Print a project-hydrated build instruction (wasm wiring with pinned `wren-core-wasm` version, the project's model/column inventory, data-mode guidance, acceptance criteria, target folder). Writes nothing to disk. |
| `wren genbi register / list / remove` | Machine-written app index. |
| `wren genbi verify <name>` | Deterministic deploy preflight. |
| `wren genbi open <name>` | Local preview. |
| `wren genbi deploy <name> --provider vercel\|cloudflare [--prod]` | Deploy and return a shareable URL. |

## Data modes

- **snapshot** (default) — data is bundled with the app as parquet/duckdb and queried client-side. Fully serverless.
- **live** — the app calls back to the user's warehouse/API at view time. Requires a CORS-enabled endpoint.

## Security

- **Secret-leak gate** — `verify` fails the app if credentials are inlined into app files, and `deploy` refuses to ship it (the deploy target is a public static host).
- **No tokens on the command line** — `VERCEL_TOKEN` / `CLOUDFLARE_API_TOKEN` (and `CLOUDFLARE_ACCOUNT_ID`) are discovered from the environment or `.env`, never passed as flags.
- Production deploys are preview-by-default; `--prod` is explicit.

## Skill

Ships a `genbi-app` agent workflow guide. Per main's new model (skills served from the CLI, with `skills/wren/` as the only external discovery stub), the guide lives at `core/wren/src/wren/skills_content/genbi-app/SKILL.md` and is served via `wren skills get genbi-app` — no separate `skills/` directory, no version drift.

## Tests

- Unit tests for build, deploy, index, and verify (`core/wren/tests/unit/test_genbi_*.py`).
- End-to-end walkthrough script `core/wren/scripts/genbi-e2e.sh`.
- The existing served-content guard and skill-stub guards pass with `genbi-app` added to the bundled set.

## Notes

This branch also merges `main` (0.9.0) and aligns the GenBI skill with main's "skills served from the CLI" architecture (#2329).

## Summary by CodeRabbit

* **New Features**
  * Added a full `wren genbi` command suite: build (composes agent instructions), register/list/remove, verify, open (local HTTP preview), and deploy to Vercel or Cloudflare. Token lookup supports env/project .env fallbacks.

* **Bug Fixes / Safety**
  * Strict app-name validation and structural/security preflight checks to prevent path traversal and inline-secret leakage.

* **Documentation**
  * New GenBI workflow guide and updated skill/manifest entries.

* **Tests**
  * Extensive unit tests covering build, index, verify, deploy, and provider flows.

https://github.com/user-attachments/assets/73004673-e07e-4d30-9879-dfeac2e9afcd

https://github.com/user-attachments/assets/2c64af4c-0350-4a58-9f6e-4dca9fd408b5


skills/wren/SKILL.md
```diff
@@ -1,6 +1,6 @@
...
...ases (Postgres, MySQL, BigQuery, Snowflake, Spark, …). The actual workflow guides live inside the `wren` CLI itself; this is just a discovery stub. Use whenever the user asks a data question (how many, show me, top N, compare, trend, breakdown, metric, revenue, customers, orders), wants to install / set up Wren Engine, connect a new database, connect SaaS data via dlt (HubSpot, Stripe, Salesforce, GitHub, Slack), generate or regenerate an MDL project from a database schema, enrich a project with business context (enum meanings, units, cubes like ARR / DAU / churn), or turn a project's semantic layer into a shareable GenBI web app / dashboard and deploy it to Vercel or Cloudflare. Triggers: 'install wren', 'set up wren engine', 'connect database to wren', 'connect SaaS to wren', 'load hubspot / stripe / salesforce data', 'generate mdl', 'scaffold wren project', 'enrich wren context', 'augment my project', 'add cubes', 'build a dashboard', 'make a shareable analytics app', 'deploy my semantic layer as a web app', 'genbi app', '把這個語意層變成 app', '做一個 dashboard 分享出去', '部署成網頁', 'wren onboarding', 'wren usage', 'wren generate mdl', 'wren dlt connector', 'wren enrich context', 'wren genbi'."
```
goldmedal: Is it necessary to add the Chinese description here? Or we can use english only?

goldmedal:
## Review — GenBI app build & deploy

**Verdict: Approve with one security follow-up.** Clean CLI↔agent split, deterministic state owned by the CLI, and strong behavior-test coverage across build/index/verify/deploy. The prior review round already hardened path traversal, symlink exfiltration, URL fabrication, malformed-index handling, and fail-closed verify — each with a regression test. One gap below is worth fixing before `verify` is leaned on as a real safety boundary.

### Scope & CI
- Entirely `core/wren/**` + `skills/` — no Rust/PyO3/connector/MDL surfaces, so no connector-regression or schema risk.
- `wren-ci.yml` covers `core/wren/**` and runs `pytest tests/unit/ -v -m unit`; all new tests carry `pytestmark = pytest.mark.unit`, so they run.
- Verified `@wrenai/wren-core-wasm@0.4.1` matches `core/wren-core-wasm/package.json` exactly (name + version) — no CDN drift.

### 🟠 Medium — secret-leak gate has a suffix blind spot (reproduced)
`verify._scan_for_secrets` scans an **allowlist** of suffixes only: `.css .htm .html .js .json .md .mjs .txt`. Everything else is skipped — including the two most likely secret carriers an LLM will create:
- **`.env`** — the canonical secret file; agents write these reflexively.
- **`.ts` / `.tsx` / `.jsx` / `.vue`** — very plausible for a "web app"; none are scanned.

Meanwhile `vercel._collect_files` uploads **every** regular file under the app dir (only symlinks / outside-root skipped), and `wrangler pages deploy .` ships the whole folder. I reproduced this: a `live` app carrying a `.env` (postgres connection string) **and** a `config.ts` with the same secret **passes `verify` cleanly** and would be uploaded to the public host — defeating the gate's stated purpose ("a credential-leaking app can never reach a public URL").

**Recommendation:** invert to **default-deny** — scan every file *except* a small set of known-binary/data suffixes (`.parquet`, `.duckdb`, images, fonts, `.wasm`) — rather than allowlisting text types. As belt-and-suspenders, also refuse to upload dotfiles like `.env*` in `_collect_files`. At minimum add `.env*`, `.ts`, `.tsx`, `.jsx`, `.vue`, `.yaml`, `.yml`, `.toml` — but the allowlist will keep losing this race against whatever extension the agent picks next; default-deny is the durable fix.

### 🟡 Low — don't oversell the gate
The patterns are deliberately narrow (good for false positives) but miss GCP service-account JSON (`private_key`), bare bearer tokens, and generic high-entropy keys. The composer/SKILL text presents `verify` as *the* thing that catches inlined credentials, which invites over-trust. Worth one sentence noting it's best-effort defense-in-depth, not a guarantee — the real rule (already stated) is "never inline secrets."

### Nits
- CI lint runs `ruff … src/` only, so the new test files aren't lint-gated (pre-existing behavior).
- `open`/`remove` take `name` without routing through `_resolve_app_dir`, but they only touch index dict keys / serve an already-validated dir, and `register` (the only index writer) validates the slug — no traversal risk. Fine as-is.

### Done well
- Token discipline is textbook: env → `.env` tiers, never argv, never persisted — asserted on both providers and on `apps.yml`.
- The autouse `_isolate_env_loading` fixture is the right call — without it the missing-token tests could pass spuriously or fire a real API call on a dev machine with a live `VERCEL_TOKEN`.
- "Never fabricate a URL" + fail-closed-on-unknown-mode + `MalformedIndexError` are each small, exactly-right robustness choices with tests.

The Medium finding is the only thing I'd want addressed before merge, given security is the headline feature.

goldmedal (commented):
We should also write the doc for the genBi command.

goldmedal:
### Follow-up note: update the project-structure doc

The default-deny secret-scan fix looks good (verified: the `.env` + `config.ts` case that previously slipped now fails verify, with regression tests). One minor doc-completeness gap worth folding into this PR's docs commit:

GenBI introduces two new **project-local** artifacts — `<project>/apps/<name>/` (the generated app) and `<project>/.wren/apps.yml` (the app index) — but the canonical "what lives in the project vs. globally in `~/.wren/`" reference, [`docs/core/guides/manage_project.md`](https://github.com/Canner/WrenAI/blob/f0d41537bea610232157622c59d99b7900f35b2f/docs/core/guides/manage_project.md#L9-L12), isn't updated and lists neither. It already documents `target/mdl.json` and `.wren/memory/`, so adding `apps/` and placing `.wren/apps.yml` next to `.wren/memory/` keeps the overview complete. The genbi guide/cli reference mention `apps/<name>/` inline, but the structure overview doesn't.

Not a blocker — `genbi.md` covers the layout for genbi users; this is just to keep the central structure doc in sync with the directories this PR adds.

(Unrelated, pre-existing: `docs/core/README.md` links to `guides/modeling/wren_project.md`, which doesn't exist yet — leave for a separate PR.)

goldmedal (approved):

PR #2349, opened 2026-06-08
docs: add China mainland mirror tips for pip and HuggingFace

HBH0713:
Add a tip for users in mainland China about using Tsinghua PyPI mirror and HuggingFace mirror.
  When installing from China, both `pip install` and HuggingFace model downloads can fail due to network restrictions. This tip saves new users from the troubleshooting we went through.

## Summary by CodeRabbit

* **Documentation**
  * Updated Quickstart installation section with guidance for users in mainland China, including alternative PyPI mirror configuration and HuggingFace mirror endpoint settings to improve installation reliability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,13 @@ pip install wrenai                      # core (DuckDB included)
 pip install "wrenai[postgres,memory]"   # add per-datasource and memory extras as needed
 ```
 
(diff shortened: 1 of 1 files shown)


goldmedal (approved):
Thanks @HBH0713 👍

PR #2352, opened 2026-06-08
perf(cli): use find_spec instead of eager import to detect memory extra

ttw225:
Non-memory CLI commands (`--version`, `--help`, `context`, `dry-run`, etc.) currently pay the startup cost of the optional memory stack because `wren/cli.py` eagerly imports `lancedb` and `sentence_transformers` at module load time just to detect whether the memory extra is installed.

That eager import pulls in the heavy ML stack (`lancedb -> sentence_transformers -> transformers/torch ...`) even when the user is not running a `wren memory` command. On macOS, a fresh install can also trigger XProtect scanning of the large native binaries on first run.

This PR replaces the try/import detection block with `importlib.util.find_spec`, so Wren checks package availability without executing package `__init__` files. `from wren.memory.cli import memory_app` remains cheap because the heavy memory imports are already lazy inside the memory command handlers.

## Measured impact

Warm startup, `wrenai[main,memory]`, invoked directly via `.venv/bin/wren` (N=15, mean ± σ):

| Command | Before (eager import) | After (`find_spec`) | Speedup |
| --- | ---: | ---: | ---: |
| `wren --version` | 2.82s ± 0.13 | 0.16s ± 0.002 | **~17.5x faster** |
| `wren --help` | 2.76s ± 0.10 | 0.19s ± 0.004 | **~14.3x faster** |

`torch`, `lancedb`, and `sentence_transformers` no longer enter `sys.modules` on non-memory startup — confirmed via `python -X importtime` and the regression test added below. (Invoked through `uv run` instead of the binary directly, both versions also carry uv's fixed ~0.25s wrapper overhead, so the "after" reads ~0.41s / ~6x there — the import-cost reduction is the same.)

### macOS first run

On a clean install, the first `wren --version` drops from **~48s to ~5s**. The eager import loads the ~800MB of ML native libraries (`libtorch_cpu.dylib` 240MB, lancedb, …), which macOS XProtect then scans on first execution — process sampling shows `XprotectService` pinned at ~95% CPU for the whole window. With `find_spec`, non-memory startup no longer loads those libraries, so the scan is not triggered. The one-time scan is *deferred* to the first real `wren memory` command, not eliminated.

## Test plan

- Added regression coverage that importing `wren.cli` does not load `torch` or `lancedb`
- Added coverage that `wren memory` is registered when the memory extra is installed
- Added coverage that `wren memory` is absent when the memory extra is not installed
- Ran `uv run pytest tests/test_cli_memory_detection.py tests/unit/test_version.py`

## Summary by CodeRabbit

* **Performance**
  * Deferred registration of the optional memory subcommand so CLI startup is lighter when optional ML dependencies are not present.

* **Tests**
  * Added tests ensuring the CLI does not load heavy ML dependencies on import and that the memory subcommand is present only when its extras are available.

### DIFF
--- a/core/wren/src/wren/cli.py
+++ b/core/wren/src/wren/cli.py
@@ -607,17 +607,14 @@ app.add_typer(cube_app)
 app.add_typer(utils_app)
 app.add_typer(skills_app)
 
-try:
-    import lancedb  # noqa: PLC0415, F401
-    import sentence_transformers  # noqa: PLC0415, F401
+from importlib.util import find_spec  # noqa: E402
 
+# Detect the optional `memory` extra without importing it: a real import would
+# eagerly pull lancedb (and the heavy ML stack it loads) into every CLI startup.
+if find_spec("lancedb") and find_spec("sentence_transformers"):
     from wren.memory.cli import memory_app  # noqa: PLC0415
 
     app.add_typer(memory_app)
-except ImportError:
-    # `memory` is installed on demand via `pip install "wrenai[memory]"`;
-    # until then the subcommand group simply isn't registered.
-    pass
 
 from wren.profile_cli import profile_app  # noqa: PLC0415, E402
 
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
Thanks @ttw225  look great 👍

PR #2359, opened 2026-06-10
docs(wren): fix cube quickstart and align YAML/CLI examples with implementation

ttw225:
## Problem

While following the quickstart literally, [Step 8 on main](https://docs.getwren.ai/oss/get_started/quickstart#step-8--query-a-cube-optional) does not have a cube available to query yet.

This PR updates the quickstart to include a **cube creation step** before querying it, so the flow becomes complete from setup to query execution.

While making this adjustment, I also noticed a few small mismatches in the current cube-related examples and aligned them with the actual schema and CLI behavior:

1. Example fields and cube names were not fully aligned with jaffle_shop.  
   The examples used `created_at` as the time dimension name, while jaffle_shop / the generated `orders` model uses `order_date`. The cube name `order_metrics` and the measure naming also did not match the cube that users would create in the updated quickstart (`revenue`, with measures `total` and `order_count`).

2. The YAML examples included a `grain:` field that is not part of the current schema.  
   `mdl.schema.json` defines `timeDimension` with only `name`, `expression`, and `type` (`additionalProperties: false`). Granularity is applied at query time via the CLI (`--time-dimension name:granularity`), not in YAML, so the existing `grain: month` examples could be confusing.

3. Some CLI examples did not match the current argument format.  
   `--time-dimension` expects `name:granularity`, and `--filter` expects `dim:op:value`.

If I misunderstood anything or missed some context here, please feel free to correct me. Thank you!

### Quickstart

- Added a cube creation step before querying a cube.
- Split the original Step 8 flow into:
  - **8a** — add a cube using Claude Code and the `/wren` skill.
  - **8b** — query the cube after it exists.
- Updated the examples to use a consistent jaffle_shop `revenue` cube:
  - `SUM(amount)` as `total`
  - `order_count`
  - grouped by `status`
  - monthly by `order_date`

### Cube guide and CLI reference

- Updated CLI examples so they match the current parser behavior:
  - `--time-dimension "order_date:month"`
  - `--filter "status:eq:completed"`
- Removed the `wren.cube.query(...)` Python example, since that API does not currently exist.
- Aligned the CLI reference example with the same `revenue` cube used in the quickstart.

### MDL reference and skill template

- Removed `grain: month` from:
  - `docs/core/reference/mdl.md`
  - `docs/core/guides/cubes.md`
  - `enrich-context` skill template (`cube_proposals.md`)
- Updated the `time_dimensions[]` table description in `mdl.md` to clarify that granularity is applied at query time via `--time-dimension name:granularity`.
- Replaced the `LAST_7_DAYS` placeholder in the skill example with a concrete `start,end` date range format, since the CLI expects two concrete dates.

### README and usage skill

- Aligned `core/wren/README.md` and `usage/SKILL.md` cube examples with the same `revenue` / `order_date` / `total` naming used in the quickstart and cube guide.

### Tests

- Extended the `_make_mdl` fixture in `test_cube_cli.py` with:
  - an `o_orderdate` column
  - a `timeDimensions` entry
- Added `test_cube_query_sql_only_time_dimension` to verify that:
  - `--time-dimension order_date:month`
  - produces `DATE_TRUNC('month', o_orderdate)` in the generated SQL.

## Scope

This PR updates docs, skill content (`enrich-context` and `usage`), `core/wren/README.md`, and one test file.

It does not change engine logic, CLI argument parsing, or the YAML loader.

The updates were checked against the current source of truth:

- CLI flag behavior: `cube_cli.py` (`--time-dimension`, `--dimensions`, `--filter`)
- YAML schema: `mdl.schema.json` (`timeDimension` definition, `additionalProperties: false`)
- Rust struct: `TimeDimension` in `manifest-macro/src/lib.rs`
- Cube loading path: `context.py` `_load_cubes_v2` (`schema_version 3` → `cubes/<name>/metadata.yml`)

## Summary by CodeRabbit

* **Documentation**
  * Updated cube query examples to use `revenue` cube with refined time dimension syntax
  * Clarified that time granularity is applied at query time via CLI flag
  * Updated filter syntax examples to align with current conventions

* **Tests**
  * Added unit test coverage for cube query functionality

### DIFF
--- a/core/wren/README.md
+++ b/core/wren/README.md
@@ -109,8 +109,8 @@ then query them with a structured input instead of writing `GROUP BY` SQL by han
 
 ```bash
 wren cube list
-wren cube describe order_metrics
-wren cube query --cube order_metrics --measures revenue --time-dimension "created_at:month"
+wren cube describe revenue
+wren cube query --cube revenue --measures total --time-dimension "order_date:month"
 ```
 
 The translator produces `DATE_TRUNC` / `GROUP BY` / `WHERE` clauses for you and
--- a/core/wren/src/wren/skills_content/enrich-context/references/cube_proposals.md
+++ b/core/wren/src/wren/skills_content/enrich-context/references/cube_proposals.md
@@ -72,10 +72,9 @@ dimensions:
     expression: <column>      # column on base_object
     type: VARCHAR             # use the column's actual type
 time_dimensions:
-  - name: <td_name>           # e.g. "month"
+  - name: <td_name>           # e.g. "order_date" — semantic column name, not the grain
     expression: <ts_column>   # e.g. "created_at"
-    grain: month              # one of: day | week | month | quarter | year
-    type: TIMESTAMP
+    type: TIMESTAMP           # granularity (day/week/month/…) is set at query time via --time-dimension
 hierarchies:
   - name: time
     levels: [year, quarter, month]   # only if multiple time grains are declared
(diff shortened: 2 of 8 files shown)


goldmedal (approved):
Thanks @ttw225, nice catch 👍

PR #2363, opened 2026-06-12
chore(wren-core-py): migrate from Poetry to uv

ttw225:
## What

`core/wren-core-py` is the only Python module still using Poetry; every
other module (`core/wren`, SDK CI) already uses uv. This migrates it to uv
so the whole repo shares one dependency manager. The build backend stays
maturin — only the dependency manager changes.

## Why

Inherited from the wren-engine import: `pyproject.toml` already had a PEP 621
`[project]` table, but the legacy `[tool.poetry]` block was never removed
(stale version mismatch: `[project]` 0.6.0 vs `[tool.poetry]` 0.1.0). Poetry
here only managed 3 dev deps (maturin, pytest, ruff) — there's no technical
reason for it. Unifying on uv removes the repeated `pipx install poetry` steps
in CI and lets contributors build the whole Rust→bindings→CLI chain with just
uv + Rust + just.

## Changes

- **`pyproject.toml`** — remove `[tool.poetry]` / `[tool.poetry.*]`; add
  `[dependency-groups] dev` with the same pins (`maturin==1.9.4`,
  `pytest==8.4.2`, `ruff==0.13.1`).
- **Lockfile** — replace `poetry.lock` with `uv.lock`.
- **`justfile` & CI** — `uv sync --no-install-project` + `uv run --no-sync …`
  for maturin/pytest/ruff. This mirrors `poetry install --no-root`: a plain
  `uv sync` builds the editable maturin extension during dependency install
  (redundant with the explicit `maturin develop`/`build` step), so
  `--no-install-project` skips that and `--no-sync` stops uv re-syncing before
  each tool run. CI uses `--locked` to enforce `uv.lock`.
- **`core-py-ci.yml` / `wren-ci.yml`** — add `astral-sh/setup-uv`, drop
  `pipx install poetry`.
- **Docs** — `README.md`, `CONTRIBUTING.md`, `CLAUDE.md`; remove the now-dead
  `**/poetry.lock` glob from `.github/labeler.yml`.

## Notes

- Direct pins (maturin/pytest/ruff) are unchanged. Only pytest's transitive
  test-only deps drift in the new lock (`iniconfig`, `packaging`, `pygments`) —
  harmless for a migration.

## Release Notes

* **Chores**
  * Migrated dependency management from Poetry to `uv` across development workflows, CI/CD pipelines, and build configurations
  * Updated auto-labeling configuration for dependency tracking

* **Documentation**
  * Updated development setup instructions and environment configuration guides to reflect new tooling


goldmedal (approved):
Thanks @ttw225 👍

PR #2368, opened 2026-06-12
fix: disable transformers progress bar on Windows

### LINKED ISSUE #2366: Windows: wren memory fetch fails with OSError: [WinError 1] Incorrect function while loading sentence-transformers model
## Summary

On Windows, `wren memory fetch` fails during memory store initialization when the local sentence-transformers embedding model is loaded.

The failure happens inside `transformers -> tqdm -> sys.stderr/sys.stdout.flush()` while loading model weights.

## Environment

* OS: Windows
* Python: 3.11
* Environment: conda
* Shells tested:

  * Git Bash / MINGW64
  * Windows terminal / CMD style
* Wren command:

```bash
wren memory fetch -q "successful transaction volume in may"
```

## Error

```text
OSError: [WinError 1] Incorrect function
```

Traceback points to:

```text
wren.memory.cli.fetch
wren.memory.store.MemoryStore.__init__
wren.memory.embeddings.warm_up
lancedb.embeddings.sentence_transformers
sentence_transformers.SentenceTransformer
transformers.modeling_utils
tqdm.std.status_printer
sys.stderr.flush / sys.stdout.flush
```

There is also this warning before the crash:

```text
Warning: You are sending unauthenticated requests to the HF Hub.
Please set a HF_TOKEN to enable higher rate limits and faster downloads.
```

But the HF warning does not appear to be the root cause. The crash seems related to the transformers/tqdm progress bar while loading the local embedding model.

## Expected behavior

`wren memory fetch` should initialize the local memory store and embedding model successfully on Windows.

## Actual behavior

The command crashes with:

```text
OSError: [WinError 1] Incorrect function
```

## Workaround

This workaround works without modifying Wren source code:

```bash
python -c "from transformers.utils import logging; logging.disable_progress_bar(); logging.set_verbosity_error(); from wren.cli import app; app()" memory fetch -q "successful transaction volume in may"
```

This suggests the issue is caused by the transformers/tqdm progress bar being enabled before Wren loads the sentence-transformers embedding model.

## Possible fix

Disable transformers progress bars before warming up the embedding model, especially on Windows, or provide an official CLI option / environment-variable based workaround for Wren memory commands.

For example, Wren could call something equivalent to:

```python
from transformers.utils import logging
logging.disable_progress_bar()
```

before initializing the sentence-transformers embedding model.

Federicorao:
Fixes #2366.

Disable transformers progress bar before loading sentence-transformers to prevent OSError on Windows due to sys.stderr.flush() failure. On Windows, the tqdm progress bar used by transformers/huggingface internally calls sys.stderr.flush() which can raise OSError: [WinError 1] Incorrect function in certain terminals (e.g., Git Bash). Calling transformers.logging.disable_progress_bar() before importing sentence-transformers or lancedb.embeddings eliminates this crash without changing any other behavior.

I checked that the changed Python files still parse correctly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced noisy console output by disabling Transformers/Hugging Face progress bar display during embedding function initialization and warm-up.
* **Chores**
  * Kept embedding startup behavior the same while ensuring initialization runs without extra progress logging.

### DIFF
--- a/core/wren/src/wren/memory/embeddings.py
+++ b/core/wren/src/wren/memory/embeddings.py
@@ -8,8 +8,6 @@ from __future__ import annotations
 import contextlib
 import os
 
-from transformers.utils import logging as transformers_logging
-
 _DEFAULT_MODEL = os.getenv(
     "WREN_EMBEDDING_MODEL", "paraphrase-multilingual-MiniLM-L12-v2"
 )
@@ -17,6 +15,10 @@ _DEFAULT_DIM = 384
 
 
 def _disable_transformers_progress_bar() -> None:
+    # Imported lazily: transformers ships with the optional `memory` extra,
+    # so this module must stay importable when that extra is not installed.
+    from transformers.utils import logging as transformers_logging  # noqa: PLC0415
+
     transformers_logging.disable_progress_bar()
 
 
--- a/core/wren/src/wren/memory/embeddings.py
+++ b/core/wren/src/wren/memory/embeddings.py
@@ -1,39 +1,65 @@
-"""
+"""Embedding function abstraction for Wren Memory.
+
 Uses LanceDB's embedding registry with sentence-transformers (local, no API key).
 """
 
+from __future__ import annotations
+
+import contextlib
+import os
+
 from transformers.utils import logging as transformers_logging
 
-transformers_logging.disable_progress_bar()
-
-from lancedb.embeddings import get_registry
-
-
-# Register the sentence-transformers embedding function so LanceDB can use it.
-try:
-    registry = get_registry()
-    # The "sentence-transformers" embedding function is registered by LanceDB
-    # when sentence_transformers is importable.
-    _SENTENCE_TRANSFORMERS_AVAILABLE = "sentence-transformers" in registry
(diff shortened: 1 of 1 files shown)


goldmedal:
Hi @Federicorao, thanks for you working on this. There are some CI fails. Could you fix them?

Federicorao:
Thanks @goldmedal — went back through the change and fixed a real issue: the new `from transformers.utils import logging` was at module top level, which makes `wren.memory.embeddings` fail to import whenever the optional `memory` extra (transformers) isn't installed. That's a likely cause of the import-time CI failures.

I moved it into `_disable_transformers_progress_bar` as a lazy import, matching the existing lazy `lancedb` import in this module, and verified the module now imports cleanly with `transformers` absent. The `model_name` parameter and defaults are unchanged.

The only check that ran here is `labeler` (green); the test workflows look like they're waiting on maintainer approval for this fork PR — could you re-trigger them so I can confirm against the full CI? Thanks!

goldmedal (approved):
Thanks @Federicorao 👍

PR #2404, opened 2026-06-27
fix(trino): treat DECIMAL(p) as scale 0, not a default non-zero scale

Bartok9:
## Summary

- The Trino cursor type parser maps `DECIMAL(p)` (precision only) and bare `DECIMAL` to a PyArrow decimal with a **default scale of 9** instead of scale `0`.
- This mistypes every precision-only `DECIMAL` column, and for small precisions (e.g. `DECIMAL(5)`) produces `scale > precision`, which PyArrow rejects outright.

## Root Cause

**Symptom** — Columns declared as `DECIMAL(p)` (or bare `DECIMAL`) in Trino come back from `wren` with a fractional scale (e.g. `DECIMAL(10)` → `decimal128(10, 9)`). For small precisions like `DECIMAL(5)`, materialising the result table raises `pyarrow.lib.ArrowInvalid` because the requested scale (9) exceeds the precision (5).

**Root cause** — In `_trino_data_type_to_arrow` (`core/wren/src/wren/connector/trino.py`), the `T.DECIMAL` branch initialises `precision, scale = 38, 9`. The scale is only overwritten when an *explicit* second parameter is present. But in Trino `DECIMAL(p)` means precision `p` with **scale 0**, and bare `DECIMAL` is `DECIMAL(38, 0)`. The non-zero default scale is wrong.

**Evidence** — Before the fix, on current `main`:

```
decimal(10)   -> decimal128(10, 9)   # should be (10, 0)
decimal       -> decimal128(38, 9)   # should be (38, 0)
decimal(5)    -> decimal128(5, 9)    # scale > precision; ArrowInvalid when building the array
decimal(10,2) -> decimal128(10, 2)   # explicit scale unaffected (correct)
```

The existing test even codified the wrong default (`("decimal", pa.decimal128(38, 9))`); it is corrected here to `(38, 0)`.

**Fix + why this level** — Change the default to `precision, scale = 38, 0`. This is the correct layer: the cursor-type → Arrow mapping is the single place Trino DECIMAL types are interpreted, so fixing the default here corrects schema typing, value coercion, and avoids the downstream ArrowInvalid without touching any caller. The explicit `DECIMAL(p, s)` path is unchanged.

**Scope / risk** — Touches only the `T.DECIMAL` branch in the Trino connector (and the corresponding test expectations). `DECIMAL(p, s)` behaviour is identical. This mirrors the same fix in the Athena connector (#2403); the two connectors share the Trino-flavoured type grammar.

## Verification

- `uv run --no-sync pytest tests/connectors/test_trino.py -k data_type` — **40 passed** (was 37; +3 new cases).
- `ruff format --check src/` + `ruff check src/` — clean.

## Real behavior proof

Captured from a real run on this branch (`uv run python3`):

```
decimal(10)   -> decimal128(10, 0)
decimal       -> decimal128(38, 0)
decimal(5)    -> decimal128(5, 0)
decimal(10,2) -> decimal128(10, 2)
```

**Regression cases** (added to the parametrized `test_parse_trino_data_type`) assert the new shape and FAIL on current `main`:

```
FAILED test_parse_trino_data_type[decimal-expected18]      assert decimal128(38, 9) == decimal128(38, 0)
FAILED test_parse_trino_data_type[decimal(10)-expected19]  assert decimal128(10, 9) == decimal128(10, 0)
FAILED test_parse_trino_data_type[decimal(5)-expected20]   assert decimal128(5, 9)  == decimal128(5, 0)
```

All three pass with the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected parsing of Trino and Athena `DECIMAL` types so bare `DECIMAL` and precision-only `DECIMAL(p)` now default to scale `0`.
  * Improved handling of small-precision decimals (e.g., `DECIMAL(5)`) to prevent invalid decimal definitions during conversion.
  * Updated and expanded unit-test expectations to match the revised `DECIMAL` parsing behavior.

### DIFF
--- a/core/wren/src/wren/connector/trino.py
+++ b/core/wren/src/wren/connector/trino.py
@@ -101,7 +101,12 @@ def _trino_data_type_to_arrow(node) -> pa.DataType:
         return _TRINO_DATA_TYPE_TO_ARROW[kind]
 
     if kind == T.DECIMAL:
-        precision, scale = 38, 9
+        # Trino DECIMAL semantics: DECIMAL(p) is precision p with scale 0, and
+        # bare DECIMAL is DECIMAL(38, 0). Defaulting the scale to a non-zero
+        # value mistyped precision-only DECIMAL columns and, for small
+        # precisions (e.g. DECIMAL(5)), produced scale > precision, which
+        # PyArrow rejects with an ArrowInvalid.
+        precision, scale = 38, 0
         params = node.expressions
         if len(params) >= 1:
             with contextlib.suppress(AttributeError, ValueError):
--- a/core/wren/tests/connectors/test_trino.py
+++ b/core/wren/tests/connectors/test_trino.py
@@ -64,7 +64,10 @@ _SCHEMA = "default"
         # Decimal
         ("decimal(10,2)", pa.decimal128(10, 2)),
         ("decimal(38,9)", pa.decimal128(38, 9)),
-        ("decimal", pa.decimal128(38, 9)),
+        # DECIMAL(p) is scale 0 and bare DECIMAL is DECIMAL(38, 0) in Trino.
+        ("decimal", pa.decimal128(38, 0)),
+        ("decimal(10)", pa.decimal128(10, 0)),
+        ("decimal(5)", pa.decimal128(5, 0)),
         # Time / timestamp
         ("time", pa.time64("us")),
         ("time(3)", pa.time64("us")),
--- a/core/wren/tests/unit/test_athena_connector.py
+++ b/core/wren/tests/unit/test_athena_connector.py
@@ -74,8 +74,20 @@ def test_parse_athena_type_primitives(type_str, expected):
     assert _parse_athena_type(type_str) == expected
 
 
-def test_parse_athena_type_decimal():
-    assert _parse_athena_type("decimal(12,4)") == pa.decimal128(12, 4)
+@pytest.mark.parametrize(
+    ("type_str", "expected"),
+    [
+        ("decimal(12,4)", pa.decimal128(12, 4)),
+        # bare DECIMAL and precision-only DECIMAL(p) default to scale 0
+        # (kept aligned with the Trino parser).
+        ("decimal", pa.decimal128(38, 0)),
+        ("decimal(10)", pa.decimal128(10, 0)),
+        # small precision must not produce scale > precision (ArrowInvalid).
+        ("decimal(5)", pa.decimal128(5, 0)),
+    ],
+)
+def test_parse_athena_type_decimal(type_str, expected):
+    assert _parse_athena_type(type_str) == expected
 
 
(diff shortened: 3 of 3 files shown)


Bartok9:
Good catch — the Athena connector shares the same Trino-flavoured DECIMAL parsing and had the identical bug. Pushed 3d078e47 applying the same fix there: bare `DECIMAL` → `(38, 0)` and `DECIMAL(p)` → scale 0, with scale capped at precision. Both connectors are now consistent.

Bartok9:
Added the requested Athena DECIMAL regression coverage in `test_athena_connector.py` — bare `decimal` → `(38, 0)`, `decimal(10)` → `(10, 0)`, and the small-precision `decimal(5)` → `(5, 0)` case that previously risked scale > precision. `pytest -k decimal` → 4 passed. Both connectors now have matching coverage.

Bartok9:
Addressed the CodeRabbit suggestion in 2ed54d6: added `test_parse_athena_type_decimal_bare_defaults_to_scale_0`, `...precision_only_has_scale_0` (decimal(10)), and `...small_precision_only` (decimal(5)) to test_athena_connector.py. All 4 decimal cases pass locally, keeping Athena aligned with the Trino parser.

goldmedal:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

goldmedal:
@Bartok9, there is some conflict. Could you resolve it?

Bartok9:
Thanks @goldmedal — and thanks for the review. Conflict resolved: merged latest `main` (the Athena fix #2403 had landed independently, so I merged it and kept the identical decimal logic). The 9 Trino/Athena DECIMAL tests still pass locally (`pytest -k decimal`). Branch is back to mergeable.

Bartok9:
Thanks @goldmedal — conflict resolved. ✅

The branch had picked up a tangled history (duplicate commits + merge commits) after the ibis-drop refactor (#2313) landed on main. I've rebased it onto current `main` as a single clean commit — the net change is unchanged (3 files, +39/−4):

- `core/wren/src/wren/connector/trino.py` — DECIMAL(p) → scale 0; bare DECIMAL → (38, 0)
- `core/wren/src/wren/connector/athena.py` — same semantics, kept aligned with the Trino parser
- regression tests for bare DECIMAL, DECIMAL(p), and the small-precision case (DECIMAL(5)) that previously produced scale > precision → ArrowInvalid

CI is re-running on the clean commit; lint already green. Should be a clean merge now — let me know if you'd like anything else adjusted. 🙏

goldmedal (approved):
Thanks @Bartok9  look good 👍

PR #2405, opened 2026-06-28
fix(policy): block table-valued functions reached via JOIN in strict mode

Bartok9:
## Summary

- Strict-mode SQL validation blocks table-valued / set-returning functions used as a query source, but only checked the `FROM` source — never `JOIN` sources.
- A TVF that parses to an `exp.Func` subclass with **no** `exp.Table` node (notably `UNNEST` → `exp.Unnest`) slipped through strict mode whenever it was reached via a `JOIN`.

## Motivation

`_check_tables` blocks two things: (1) `exp.Table` references not in the manifest, and (2) `exp.Func` sources under `exp.From` (which is how `UNNEST`/`generate_series` in the FROM clause are caught). But a `JOIN` source was never scanned:

```sql
-- strict_mode=True, manifest = {orders}
SELECT * FROM orders CROSS JOIN UNNEST(orders.items) AS t(item)   -- silently ALLOWED (bug)
SELECT * FROM UNNEST(orders.items) AS t(item)                     -- correctly blocked
```

`generate_series`/`read_csv` in a JOIN happen to parse as an empty-name `exp.Table` and were caught, but `UNNEST` parses to `exp.Unnest` (an `exp.Func`, not a `Table`), so it bypassed the guard entirely.

## Fix

Iterate `ast.find_all(exp.From, exp.Join)` instead of `exp.From` alone, unwrapping `exp.Alias` and rejecting `exp.Func` sources. One-line scope widening; the FROM behavior is unchanged.

## Real behavior proof

**Real environment:** Python 3.12, `core/wren` via `uv run --no-sync`, repo `main`.

**Commands:**
```
uv run --no-sync pytest tests/unit/test_policy.py -q
```

**AFTER fix:**
```
25 passed in 0.21s            # test_policy.py
779 passed, 42 skipped         # full unit suite
```

**BEFORE fix (new test stashed onto unpatched policy.py):**
```
1 failed, 24 passed
FAILED tests/unit/test_policy.py::test_tvf_unnest_in_join_blocked
```

## Verification / tests

`tests/unit/test_policy.py` gains:
- `test_tvf_unnest_in_join_blocked` — fails without the fix.
- `test_tvf_generate_series_in_join_blocked` — JOIN-source TVF coverage.
- `test_join_between_two_mdl_models_allowed` — guards against over-blocking plain model JOINs.

## Scope / risk

Apache-2.0 path (`core/wren`). Only widens the existing func-source scan from FROM to FROM+JOIN. Plain joins between MDL models and all prior FROM behavior are unaffected (covered by the new allow-test + the existing 22 policy tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened strict SQL table policy validation to correctly handle row-expansion operators (such as `UNNEST`/`FLATTEN`) as non-table sources.
  * Expanded strict-mode scanning to cover both `FROM` and `JOIN`, including `LATERAL`-wrapped cases, preventing table-valued functions introduced via `JOIN` from being missed.
  * Ensured data-generating table-valued functions (e.g., `generate_series`) are still blocked when reached through join/lateral syntaxes.
* **Tests**
  * Updated and extended strict-mode tests for `UNNEST`/`FLATTEN`, including `LATERAL` scenarios and join-path regressions.

### DIFF
--- a/core/wren/src/wren/policy.py
+++ b/core/wren/src/wren/policy.py
@@ -15,6 +15,21 @@ from sqlglot.errors import SqlglotError
 from wren.config import WrenConfig
 from wren.model.error import ErrorCode, ErrorPhase, WrenError
 
+# Row-expansion operators (UNNEST / FLATTEN / EXPLODE) are not data sources —
+# they restructure an array/struct expression that is already in query scope
+# (e.g. a governed model column like ``orders.items``), so they never read data
+# outside the manifest and must not be treated as a disallowed table-valued
+# function. The class mapping is stable across dialects: ``UNNEST`` ->
+# ``exp.Unnest`` (trino/postgres/duckdb), Snowflake ``FLATTEN`` -> ``exp.Explode``,
+# so no per-dialect name matching is needed.
+#
+# Note: ``UNNEST(read_csv(...))`` is therefore also allowed, but that is an
+# instance of the broader pre-existing gap (data-reading TVFs in non-source
+# positions — already reachable today via projection/WHERE subqueries, since
+# strict mode only governs the top-level source position) and is tracked
+# separately. It is not a regression introduced here.
+_ROW_EXPANSION_FUNCS: tuple[type[exp.Func], ...] = (exp.Unnest, exp.Explode)
+
 # Dialects we probe when canonicalising the user's denylist. sqlglot can map
 # the same function name (e.g. ``version()``) onto different concrete AST
 # subclasses depending on the dialect — postgres/mysql/duckdb/trino/clickhouse
@@ -155,6 +170,10 @@ def _check_tables(
             source = source.this
             if isinstance(source, exp.Alias):
                 source = source.this
+        if isinstance(source, _ROW_EXPANSION_FUNCS):
+            # Row-expansion over an in-scope expression (e.g. a governed model
+            # column) reads nothing outside the manifest — allow it.
+            continue
         if isinstance(source, exp.Func):
             raise WrenError(
                 ErrorCode.MODEL_NOT_FOUND,
--- a/core/wren/src/wren/policy.py
+++ b/core/wren/src/wren/policy.py
@@ -146,6 +146,15 @@ def _check_tables(
         source = clause.this
         if isinstance(source, exp.Alias):
             source = source.this
+        # LATERAL-wrapped TVFs (e.g. ``LATERAL FLATTEN(...)`` /
+        # ``LATERAL generate_series(...)``) parse to an exp.Lateral node that
+        # *wraps* the function rather than being an exp.Func itself, so the
+        # bare-Func check below would miss them. Unwrap to inspect the inner
+        # source — this is the same "TVF reached via JOIN" bug class.
+        if isinstance(source, exp.Lateral):
+            source = source.this
+            if isinstance(source, exp.Alias):
+                source = source.this
         if isinstance(source, exp.Func):
             raise WrenError(
                 ErrorCode.MODEL_NOT_FOUND,
--- a/core/wren/src/wren/policy.py
+++ b/core/wren/src/wren/policy.py
@@ -137,10 +137,13 @@ def _check_tables(
             phase=ErrorPhase.SQL_POLICY_CHECK,
         )
 
-    # Func subclasses used as FROM sources (e.g. UNNEST) produce no exp.Table
-    # node at all. Scan for Func nodes inside From clauses.
-    for from_clause in ast.find_all(exp.From):
-        source = from_clause.this
+    # Func subclasses used as query sources (e.g. UNNEST, generate_series)
+    # produce no exp.Table node at all. They can appear both as the FROM
+    # source AND as a JOIN source (e.g. ``orders CROSS JOIN UNNEST(items)``),
+    # so scan both — checking only FROM let a table-valued function slip
+    # through strict mode whenever it was reached via a JOIN.
+    for clause in ast.find_all(exp.From, exp.Join):
+        source = clause.this
         if isinstance(source, exp.Alias):
             source = source.this
         if isinstance(source, exp.Func):
--- a/core/wren/tests/unit/test_policy.py
+++ b/core/wren/tests/unit/test_policy.py
@@ -242,6 +242,27 @@ def test_unnest_model_column_allowed():
     validate_sql_policy(ast, _MODELS, config)
 
 
+def test_unnest_over_reader_tvf_allowed_known_gap():
+    # Regression anchor for CodeRabbit's nested-reader question.
+    #
+    # ``UNNEST(read_csv(...))`` is currently ALLOWED, and intentionally so for
+    # this PR. The row-expansion allow-list keys on the wrapper class
+    # (exp.Unnest/exp.Explode), and the inner reader contributes no exp.Table
+    # node, so the data-source guard never sees ``read_csv``. This is NOT a
+    # regression introduced here: strict mode only governs the *top-level*
+    # source position, so readers in non-source positions (projection, WHERE
+    # subqueries, and now a row-expansion argument) were already reachable on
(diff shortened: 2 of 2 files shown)


goldmedal:
Nice, well-scoped fix — verified the mechanism directly against sqlglot (`UNNEST` via JOIN parses to `exp.Unnest` under `Join.this` with no `exp.Table` node, so the FROM-only scan genuinely missed it), and confirmed plain model joins and derived-table joins (`exp.Subquery`) are not over-blocked.

One residual gap worth a follow-up (not a blocker for this PR): **`LATERAL`-wrapped TVFs still slip through**, because they parse to an `exp.Lateral` node *wrapping* the func rather than to a bare `exp.Func`:

```sql
-- snowflake, strict_mode=True, manifest = {orders}
SELECT * FROM orders, LATERAL FLATTEN(input => orders.items) f
```
```
Join.this    -> exp.Lateral   (is_Func=False)   <- not caught by the FROM/JOIN func scan
  Lateral.this -> exp.Explode (is_Func=True)    <- the actual TVF, unseen
```

This is the same class of bug as the one fixed here (a TVF "reached via JOIN"), and it's **pre-existing** — the old FROM-only code also missed `LATERAL` in the FROM clause, so it's not a regression introduced by this PR. Given the title says "block table-valued functions reached via JOIN," it'd be worth either unwrapping `exp.Lateral` to inspect its `.this`, or explicitly scoping it out with a comment + follow-up issue.

Happy to merge as-is and track the LATERAL case separately.

Bartok9:
Thanks for the careful review and the sqlglot-verified mechanism check — much appreciated.

I went ahead and closed the `LATERAL` gap in this PR rather than deferring it, since it's the same "TVF reached via JOIN" bug class. The JOIN scan now unwraps `exp.Lateral` (its `.this` is the func directly — `LATERAL FLATTEN(...)` → `exp.Lateral` wrapping `exp.Explode`, `LATERAL generate_series(...)` → wrapping `exp.GenerateSeries`) before the bare-Func check, with an alias-unwrap inside in case the lateral source is aliased.

Added two regression tests (`6548c4c2`):
- `LATERAL FLATTEN(input => orders.items)` via comma-join (snowflake) → blocked
- `CROSS JOIN LATERAL generate_series(1, 3)` (snowflake) → blocked

Verified all four cases locally (both LATERAL forms + the existing UNNEST-via-JOIN block + the two-model JOIN still passes, no over-blocking). Ready for another look whenever you have a moment.

goldmedal:
Follow-up after the LATERAL fix — one case I think this PR should also handle: **`UNNEST` over a governed model column is currently false-blocked.**

```sql
-- strict_mode=True, manifest = {orders}
SELECT * FROM orders CROSS JOIN UNNEST(orders.items) AS t(item)   -- currently BLOCKED
```
(Verified on the PR branch: raises `MODEL_NOT_FOUND`.)

**Why it should be allowed:** `UNNEST` / `FLATTEN` / `EXPLODE` aren't *data sources* — they restructure an array/struct expression that's already in query scope. Here `orders.items` is a column of a governed model (RLAC/CLAC already apply to it); wrapping it in `UNNEST(...)` shouldn't make it a "non-model source." It reads nothing outside the manifest.

**Security check:** allowing this does not widen the attack surface. Data-reading TVFs like `read_csv` are *already* reachable today via projection / WHERE subqueries — strict mode only governs the **top-level source** position, so it never blocked `SELECT read_csv(...)` or `WHERE id IN (SELECT read_csv(...))` (confirmed on both `main` and this branch). The real exploit — `read_csv` as a top-level table source — stays blocked.

**Suggested minimal change** — a class-based allow before the bare-Func block:

```python
# Row-expansion operators (UNNEST / FLATTEN / EXPLODE) are not data sources —
# they restructure an array/struct already in query scope (e.g. a governed
# model column), so they never read data outside the manifest.
_ROW_EXPANSION_FUNCS = (exp.Unnest, exp.Explode)   # snowflake FLATTEN -> Explode

    ...
    if isinstance(source, exp.Lateral):
        source = source.this
        if isinstance(source, exp.Alias):
            source = source.this
    if isinstance(source, _ROW_EXPANSION_FUNCS):
        continue
    if isinstance(source, exp.Func):
        raise WrenError(...)
```

I verified the class mapping is stable across dialects: `UNNEST` -> `exp.Unnest` (trino/postgres/duckdb), Snowflake `FLATTEN` -> `exp.Explode` — so no per-dialect name matching is needed. Worth adding a `test_unnest_model_column_allowed` alongside the existing block tests.

Note this does mean `UNNEST(read_csv(...))` would also be allowed — but as above, that's an instance of the broader pre-existing gap (readers in non-source positions), which I've filed as a separate issue for proper treatment. It's not a regression introduced here.

Bartok9:
Done — fixed the false-block in this PR (c901a2ff).

`UNNEST` / `FLATTEN` / `EXPLODE` over a governed model column are now allowed, since they're row-expansion of an expression already in query scope (RLAC/CLAC already apply to e.g. `orders.items`) rather than a data source. Implemented exactly as you suggested — a class-based allow before the bare-`exp.Func` block:

```python
_ROW_EXPANSION_FUNCS = (exp.Unnest, exp.Explode)  # snowflake FLATTEN -> Explode
...
if isinstance(source, _ROW_EXPANSION_FUNCS):
    continue
if isinstance(source, exp.Func):
    raise WrenError(...)
```

The class mapping checked out across dialects as you noted: `UNNEST` -> `exp.Unnest` (trino/postgres/duckdb), Snowflake `FLATTEN` -> `exp.Explode`, so no per-dialect name matching. The `exp.Lateral` unwrap from the previous commit still runs first, so `LATERAL FLATTEN(...)` over a model column is allowed too.

What stays blocked (verified): top-level `read_csv(...)`, `generate_series` as a FROM/JOIN source, and `CROSS JOIN LATERAL generate_series(...)` — those are real data-generating sources, not row-expansion of an in-scope column. As you flagged, `UNNEST(read_csv(...))` now also passes, but that's the broader pre-existing non-source-reader gap you've filed separately, not a regression here.

Test changes:
- `test_unnest_row_expansion_allowed`, `test_unnest_model_column_allowed` (UNNEST via JOIN), `test_lateral_flatten_model_column_allowed` — now assert allowed.
- `generate_series` / `read_csv` block tests + the LATERAL generate_series block + the two-model JOIN allow remain.

`test_policy.py`: 27 passed. Full unit suite: 781 passed, 42 skipped. Ready for another look.

Bartok9:
@coderabbitai Good catch on the nested-reader case — I added coverage for it in `c54aaa3a`, but as an **allow** assertion rather than a block, because that's the behavior @goldmedal and I deliberately agreed on for this PR.

`test_unnest_over_reader_tvf_allowed_known_gap` parses `SELECT * FROM orders CROSS JOIN UNNEST(read_csv('s3://b/f.csv'))` in strict mode and asserts it passes. Rationale (matching the discussion above):

- The row-expansion allow-list keys on the wrapper class (`exp.Unnest`/`exp.Explode`). The inner `read_csv` contributes **no `exp.Table` node**, so the top-level source guard never sees it.
- Strict mode only governs the **top-level source position** — readers in non-source positions (projection, `WHERE` subqueries, and now a row-expansion argument) are already reachable on `main`, so this isn't a regression introduced here.
- @goldmedal confirmed this and filed the broader non-source-reader gap as a **separate issue** rather than blocking it in this PR.

So rather than assert `MODEL_NOT_FOUND` (which would contradict that agreement and force a wider change), the test pins the gap explicitly with a comment, so any future tightening is a deliberate, reviewed flip rather than silent. The real exploit — `read_csv` as a top-level source — stays blocked (`test_tvf_read_csv_blocked`, plus the FROM/JOIN/LATERAL generate_series block tests). `test_policy.py`: 28 passed.

goldmedal (approved):
Thanks @Bartok9 👍

PR #2406, opened 2026-06-28
fix(genbi): normalise null apps/schema_version in app index

Bartok9:
## Summary

- `load_index` normalised a missing `apps` key with `data.setdefault("apps", {})`, but `setdefault` is a no-op when the key already exists.
- An `apps.yml` with an explicit-but-empty `apps:` (a hand-edit or a truncated write) is valid YAML that parses to `{"apps": None}` — the key is present, so `setdefault` leaves `None` in place. Every accessor (`get_app`, `register_app`, `remove_app`, `update_app`) then crashes with an opaque `AttributeError: 'NoneType' object has no attribute 'get'`.

This defeats the whole point of `MalformedIndexError`, which exists precisely so a bad index file fails clearly instead of crashing deep in a command.

## Reproduction

```python
# .wren/apps.yml:
#   schema_version: 1
#   apps:
load_index(project)["apps"]   # -> None
get_app(project, "foo")       # -> AttributeError: 'NoneType' object has no attribute 'get'
```

## Fix

Normalise null values with `or` instead of `setdefault`, so an explicit null `apps:` / `schema_version:` is coerced to its default. Also raise `MalformedIndexError` (not a later opaque crash) when `apps` is present but is a non-mapping (e.g. a list).

## Real behavior proof

**Real environment:** Python 3.12, `core/wren` via `uv run --no-sync`, repo `main`.

**Commands:**
```
uv run --no-sync pytest tests/unit/test_genbi_index.py -q
```

**AFTER fix:**
```
13 passed                      # test_genbi_index.py
779 passed, 42 skipped         # full unit suite
```

**BEFORE fix (new tests stashed onto unpatched index.py):**
```
3 failed, 10 passed
FAILED test_load_index_normalises_null_apps
FAILED test_load_index_normalises_null_schema_version
FAILED test_load_index_raises_on_non_mapping_apps
```

## Tests

`tests/unit/test_genbi_index.py` gains coverage for null `apps`, null `schema_version`, and a non-mapping `apps` (raises `MalformedIndexError`). The null-apps test fails without the fix.

## Scope / risk

Apache-2.0 path (`core/wren`). Only changes index normalisation; well-formed index files behave identically (covered by the existing 10 index tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved normalization of incomplete index data, safely handling explicit null top-level fields.
  * Added clearer error reporting when the `apps` field isn’t a mapping, preventing downstream crashes.
  * Ensured schema version defaults apply for null/absent values while preserving explicit `0`.
* **Tests**
  * Added unit tests covering null/empty `apps`, null `schema_version`, invalid (non-mapping) `apps`, and preservation of `schema_version: 0`.

### DIFF
--- a/core/wren/src/wren/genbi/index.py
+++ b/core/wren/src/wren/genbi/index.py
@@ -50,8 +50,19 @@ def load_index(project_path: Path) -> dict:
             f"{path} is malformed — expected a mapping at the top level, "
             f"got {type(data).__name__}."
         )
-    data.setdefault("schema_version", INDEX_SCHEMA_VERSION)
-    data.setdefault("apps", {})
+    # Use ``or`` rather than ``setdefault`` so an explicit null value
+    # (``apps:`` / ``schema_version:`` with no value, e.g. from a hand-edit or
+    # a truncated write) is normalised too. ``setdefault`` is a no-op when the
+    # key already exists, which left ``apps: None`` in place and made every
+    # accessor crash with an opaque ``AttributeError`` deep in a command.
+    data["schema_version"] = data.get("schema_version") or INDEX_SCHEMA_VERSION
+    apps = data.get("apps") or {}
+    if not isinstance(apps, dict):
+        raise MalformedIndexError(
+            f"{path} is malformed — 'apps' must be a mapping, "
+            f"got {type(apps).__name__}."
+        )
+    data["apps"] = apps
     return data
 
 
--- a/core/wren/src/wren/genbi/index.py
+++ b/core/wren/src/wren/genbi/index.py
@@ -50,13 +50,18 @@ def load_index(project_path: Path) -> dict:
             f"{path} is malformed — expected a mapping at the top level, "
             f"got {type(data).__name__}."
         )
-    # Use ``or`` rather than ``setdefault`` so an explicit null value
-    # (``apps:`` / ``schema_version:`` with no value, e.g. from a hand-edit or
-    # a truncated write) is normalised too. ``setdefault`` is a no-op when the
-    # key already exists, which left ``apps: None`` in place and made every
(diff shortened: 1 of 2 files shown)


Bartok9:
Good catch, @coderabbitai — fixed in 75007cb5. Switched both fields from `or` to an explicit `is None` check so valid falsy values (`0`, `False`, `[]`, `""`) are preserved and surfaced for validation rather than silently coerced to a default. The non-mapping `apps` guard is unchanged. Added `test_load_index_preserves_falsy_schema_version` (schema_version: 0). Full file: 14 passed.

goldmedal:
@Bartok9, please check the CI fails. Thanks.

Bartok9:
Thanks @goldmedal — looked into it. The `unit tests` failure isn't from this PR's change; it's a transient network error in the Rust build step:

```
error: failed to get `jiff` as a dependency of package `env_logger v0.11.10`
  download of ji/ff/jiff failed
  curl failed
  [56] OpenSSL SSL_read: unexpected eof while reading
```

crates.io dropped the connection mid-download while building `wren-core-py`. My change is pure-Python in the genbi app index and the actual test suites that touch it pass (test_genbi_index.py 14 passed, full unit suite green locally). I've pushed an empty commit to re-trigger the run. If it flakes again it's purely a crates.io fetch hiccup on the runner, not a code issue — happy to rebase if you'd prefer.

Bartok9:
@goldmedal CI is green now ✅ — all checks pass (unit, postgres, mysql, memory, ui tests, lint, CodeRabbit). The earlier failure has cleared on the latest run. Ready for review whenever you have a moment. Thanks! 🙏

goldmedal (approved):
Thanks @Bartok9 👍

PR #2410, opened 2026-06-29
feat(wren): add cross-dialect type translation to type_mapping

Bartok9:
## What & why

`wren.type_mapping` already gives us `parse_type()` / `parse_types()` to **normalize** a raw DB type string into sqlglot's canonical form for a single dialect. A common adjacent need in cross-engine modeling (mirroring a Postgres schema into BigQuery, generating MDL for a different target engine, etc.) is to **translate** a type from one engine's spelling into another's.

This PR adds that as a small, focused companion:

- `translate_type(type_str, source_dialect, target_dialect)` — parse in source dialect, render in target dialect.
- `translate_types(columns, source_dialect, target_dialect, *, type_field="raw_type")` — batch variant mirroring `parse_types` (non-mutating, adds a `type` key).
- `wren utils translate-type` and `wren utils translate-types` CLI commands, matching the existing `parse-type` / `parse-types` UX (single via flags, batch via stdin/`--input` JSON).

Examples:

| input | source → target | output |
|-------|-----------------|--------|
| `int8` | postgres → bigquery | `INT64` |
| `character varying(255)` | postgres → clickhouse | `Nullable(String)` |
| `INT64` | bigquery → postgres | `BIGINT` |
| `DECIMAL(10,2)` | mysql → snowflake | `DECIMAL(10, 2)` |

Parse failures fall back to the original string, matching `parse_type`'s existing contract.

## Implementation notes

- Built on **sqlglot only** — no connector drivers — so it's consistent with the existing module and the new tests run in the lightweight `unit tests` CI job (no `pytest.importorskip` guard needed).
- Original input dicts are never mutated; helpers return new lists.
- No new dependencies; no behavior change to existing `parse_*` paths.

## Tests

Added to `tests/unit/test_type_mapping.py`:
- parametrized `translate_type` cases (cross-dialect, identity-normalization, unknown-type fallback, empty passthrough)
- `translate_types` batch tests (adds field, no mutation, custom field, empty list)
- CLI integration tests for `translate-type` / `translate-types`

Local run (`core/wren`, sqlglot 30.12.0):
```
$ python -m pytest tests/unit/test_type_mapping.py -q
43 passed in 2.27s
$ ruff check src/wren/type_mapping.py src/wren/utils_cli.py tests/unit/test_type_mapping.py
All checks passed!
```

## Summary by CodeRabbit

* **New Features**
  * Added SQL type translation between dialects for single values and batch inputs.
  * Added CLI commands to translate one type or a JSON list (from a file or stdin).
* **Bug Fixes**
  * Improved error handling for unreadable input files and invalid JSON, returning a clean nonzero exit with an error message.
  * Unknown/invalid types now safely fall back to the original value.
  * Batch translation avoids mutating the original records.
* **Tests**
  * Expanded unit and CLI integration coverage for translation, custom type fields, empty inputs, and failure scenarios.

### DIFF
--- a/core/wren/src/wren/type_mapping.py
+++ b/core/wren/src/wren/type_mapping.py
@@ -14,6 +14,8 @@ Use as a library:
     #     {"column": "id", "raw_type": "int8", "type": "BIGINT"},
     #     {"column": "name", "raw_type": "character varying", "type": "VARCHAR"},
     # ]
+
+    translate_type("int8", "postgres", "bigquery")  # → "INT64"
 """
 
 from __future__ import annotations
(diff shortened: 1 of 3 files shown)


Bartok9:
Good catch — fixed in `82234814`. Wrapped `path.read_text(...)` in both `translate-types` and `parse-types` so `OSError`/`UnicodeDecodeError` now hit the clean `typer.Exit(1)` path with a `could not read file` message instead of escaping as a traceback, matching the existing not-found and invalid-JSON handling. Added two regression tests (missing file → exit 1 + no traceback; unreadable path → `could not read file` + no traceback). Full suite: 45 passed.

Bartok9:
The one actionable finding (catch `OSError`/`UnicodeDecodeError` alongside `json.JSONDecodeError` on `--input` file reads) is already addressed in commit 82234814 for both `translate-types` and `parse-types` — the `path.read_text(...)` call is wrapped in a try/except that routes `OSError`/`UnicodeDecodeError` through `typer.echo(..., err=True)` + `typer.Exit(1)`. No further change needed.

goldmedal (approved):
Thanks @Bartok9 look good 👍

PR #2411, opened 2026-06-29
fix(duckdb): case-insensitive .duckdb discovery, deterministic attach aliases, and single-statement SQL hardening

Bartok9:
## Summary

Harden the DuckDB connector around database-file discovery and single-statement SQL handling:

1. **Case-insensitive `.duckdb` discovery** (the original fix) — `_list_duckdb_files` matches the extension case-insensitively, so valid databases with upper/mixed-case extensions (e.g. `WAREHOUSE.DUCKDB`) are no longer silently skipped and no longer surface a spurious `DUCKDB_FILE_NOT_FOUND`.
2. **Deterministic alias de-duplication** — case-insensitive discovery can now surface `warehouse.duckdb` and `warehouse.DUCKDB`, which previously derived the same attach alias. `_attach_database` sorts the file list and derives unique aliases deterministically (`warehouse`, `warehouse_1`, …), tracked case-insensitively.
3. **`query()` / `dry_run()` semicolon handling** — both now use the shared `_strip_trailing_semicolon` regex helper (matching `canner`/`clickhouse`/`trino`) so a trailing `;` doesn't produce invalid wrapped SQL, while semicolons inside string literals are preserved. `dry_run()` neutralizes multi-statement input by wrapping in a `LIMIT 0` subquery (same as the other connectors) rather than pre-rejecting it — no side effects, no false positives.

## Motivation

`str.endswith(".duckdb")` is case-sensitive, dropping mixed-case database files on case-preserving filesystems/object stores and raising `DUCKDB_FILE_NOT_FOUND` even when a valid database is present. The follow-on changes address edge cases that fix newly exposes (alias collisions) and align the LIMIT-wrap / dry-run semicolon handling with the rest of the codebase.

## Fix

- `_list_duckdb_files`: `file.path.lower().endswith(".duckdb")`.
- `_attach_database`: `sorted()` + deterministic, case-insensitively tracked unique aliases.
- `query()`/`dry_run()`: reuse `_strip_trailing_semicolon`; `dry_run` wraps in `LIMIT 0` subquery.

## Verification

`uv run --no-sync pytest tests/unit/test_duckdb_file_listing.py tests/unit/test_engine.py -q` — 15 passed. `ruff check` clean.

Regression tests cover: mixed-case discovery, case-colliding alias uniquification, `query()` trailing-semicolon strip, `dry_run()` LIMIT-0 wrap, and `dry_run()` preserving a semicolon inside a string literal.

## Scope / risk

Changes confined to `core/wren/src/wren/connector/duckdb.py` (Apache-2.0 path) plus its unit tests.

### DIFF
--- a/core/wren/src/wren/connector/duckdb.py
+++ b/core/wren/src/wren/connector/duckdb.py
@@ -135,9 +135,7 @@ class DuckDBConnector(ConnectorABC):
             for file in op.list("/"):
                 if file.path != "/":
                     stat = op.stat(file.path)
-                    if not stat.mode.is_dir() and file.path.lower().endswith(
-                        ".duckdb"
-                    ):
+                    if not stat.mode.is_dir() and file.path.lower().endswith(".duckdb"):
                         files.append(f"{connection_info.url}/{file.path}")
         except Exception as e:
             raise WrenError(
--- a/core/wren/src/wren/connector/duckdb.py
+++ b/core/wren/src/wren/connector/duckdb.py
@@ -98,6 +98,14 @@ class DuckDBConnector(ConnectorABC):
                 )
 
     def _list_duckdb_files(self, connection_info) -> list[str]:
+        """List DuckDB database files in the configured directory.
+
+        Walks the connection's root directory and returns the full paths of
+        all non-directory entries whose name ends with ``.duckdb``. The
+        extension comparison is case-insensitive so files exported or renamed
+        with upper/mixed-case extensions (e.g. ``WAREHOUSE.DUCKDB``) are still
+        discovered. Raises ``WrenError`` if the directory cannot be listed.
+        """
         op = opendal.Operator("fs", root=connection_info.url)
         files = []
         try:
--- a/core/wren/src/wren/connector/duckdb.py
+++ b/core/wren/src/wren/connector/duckdb.py
@@ -96,8 +96,11 @@ class DuckDBConnector(ConnectorABC):
         if not db_files:
             raise WrenError(ErrorCode.DUCKDB_FILE_NOT_FOUND, "No DuckDB files found.")
 
+        # Sort for deterministic alias assignment: OpenDAL listing order is not
+        # guaranteed, so without this the bare alias could attach to a different
+        # file across runs when case-colliding names are present.
         used_aliases: set[str] = set()
-        for file in db_files:
+        for file in sorted(db_files):
             try:
                 escaped_file = file.replace("'", "''")
                 base_alias = os.path.splitext(os.path.basename(file))[0]
--- a/core/wren/src/wren/connector/duckdb.py
+++ b/core/wren/src/wren/connector/duckdb.py
@@ -96,10 +96,22 @@ class DuckDBConnector(ConnectorABC):
         if not db_files:
             raise WrenError(ErrorCode.DUCKDB_FILE_NOT_FOUND, "No DuckDB files found.")
 
+        used_aliases: set[str] = set()
         for file in db_files:
             try:
                 escaped_file = file.replace("'", "''")
-                alias = os.path.splitext(os.path.basename(file))[0].replace('"', '""')
+                base_alias = os.path.splitext(os.path.basename(file))[0]
+                # Case-insensitive discovery can surface files whose names differ
+                # only by case (e.g. warehouse.duckdb / warehouse.DUCKDB), which
+                # would otherwise derive the same attach alias and collide. Make
+                # each alias unique deterministically.
+                unique_alias = base_alias
+                suffix = 1
+                while unique_alias.lower() in used_aliases:
+                    unique_alias = f"{base_alias}_{suffix}"
+                    suffix += 1
+                used_aliases.add(unique_alias.lower())
+                alias = unique_alias.replace('"', '""')
                 self.connection.execute(
                     f"ATTACH DATABASE '{escaped_file}' AS \"{alias}\" (READ_ONLY);"
                 )
--- a/core/wren/src/wren/connector/duckdb.py
+++ b/core/wren/src/wren/connector/duckdb.py
@@ -104,7 +104,9 @@ class DuckDBConnector(ConnectorABC):
             for file in op.list("/"):
                 if file.path != "/":
                     stat = op.stat(file.path)
-                    if not stat.mode.is_dir() and file.path.endswith(".duckdb"):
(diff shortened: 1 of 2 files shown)


Bartok9:
Thanks @coderabbitai — good catch on the alias collision. That edge case is real: case-insensitive discovery can now surface `warehouse.duckdb` and `warehouse.DUCKDB`, both of which previously derived the same attach alias (`warehouse`), so the second `ATTACH` would fail on a case-sensitive filesystem/object store.

Pushed a fix: `_attach_database` now derives a unique alias per file deterministically (`warehouse`, then `warehouse_1`, …), tracked case-insensitively to match the discovery semantics. Added a regression test (`test_attach_database_uniquifies_case_colliding_aliases`) asserting distinct aliases. Full suite green.

On the two inline LIMIT/`dry_run` semicolon comments: those touch pre-existing code paths unrelated to this one-line case-insensitivity change, so I'm leaving them out of scope for this PR to keep it focused.

Bartok9:
Thanks @coderabbitai — addressed the nondeterministic alias ordering. `db_files` is now `sorted()` before the attach loop, so the bare alias deterministically maps to the same file across runs regardless of OpenDAL listing order, while the existing `used_aliases` collision handling (warehouse, warehouse_1, …) is preserved. Pushed in 5cbbc49.

Bartok9:
Addressed in cb3bcb97:

- **query() LIMIT-wrap with trailing semicolon** (valid): now strips a trailing `;` before wrapping, so `SELECT 1;` becomes `SELECT * FROM (SELECT 1) AS _q LIMIT n`.
- **dry_run() multi-statement guard** (valid): rejects multi-statement input via WrenError before EXPLAIN, preserving the no-side-effects contract (single trailing `;` still allowed).
- **Alias collision on case-insensitive discovery** (already resolved): `_attach_database` deterministically uniquifies case-colliding aliases (`warehouse` / `warehouse_1`) over a sorted file list.

Added 3 regression tests; full `test_duckdb_file_listing.py` (5) and `test_engine.py` (9) pass, ruff clean.

goldmedal:
## Review

The core fix is correct and well-tested, but the PR bundles more than the title/description says, and the `dry_run` change has a correctness bug. Two items to address before merge (no need to split the PR — updating the title/description is fine).

### 🔴 1. Title & description don't match the diff

The body says:

> One-line change (`file.path` -> `file.path.lower()`) in the suffix check only. Directory exclusion and path construction are untouched.

But the diff actually changes four things:

| Change | Disclosed? |
|---|---|
| `_list_duckdb_files` — `.lower().endswith(".duckdb")` | ✅ the stated fix |
| `_attach_database` — `sorted()` + alias uniquification | ⚠️ only in the CodeRabbit blurb (a valid follow-on: the case fix newly makes case-colliding aliases possible) |
| `query()` — strip trailing `;` before the LIMIT wrap | ❌ not mentioned |
| `dry_run()` — reject multi-statement SQL | ❌ not mentioned |

The `query`/`dry_run` changes are legitimate hardening, so keeping them here is fine — but please **update the title and description** so they reflect the real scope (case-insensitive discovery + alias dedup + `query`/`dry_run` semicolon/multi-statement handling). A reviewer reading "one-line change" would not expect `dry_run` semantics to change.

### 🔴 2. `dry_run` multi-statement check has false positives

```python
stripped = sql.rstrip().rstrip(";")
if ";" in stripped:
    raise WrenError(ErrorCode.INVALID_SQL, "dry_run only accepts a single SQL statement.")
```

`";" in stripped` rejects **valid single statements** that contain a semicolon inside a string literal:

```
SELECT ';' AS x   →  rejected as "multi-statement"
```

This contradicts the codebase's own stance — the sibling `_strip_trailing_semicolon` helpers (`clickhouse.py:315`, `trino.py:354`, `canner.py:124`) deliberately *preserve* semicolons inside string literals. The new tests don't cover this case, so it stays green while the bug is live.

Suggested fix: the other connectors don't reject multi-statement at all — they neutralize it by wrapping in a subquery (`SELECT * FROM ({stripped}) AS _q LIMIT 0`), where any trailing statement becomes a natural syntax error. DuckDB's `EXPLAIN {sql}` is the odd one out because `EXPLAIN` doesn't wrap. Applying the same subquery-wrap to `dry_run` removes the side-effect risk without false positives and matches the rest of the codebase.

### 🟡 3. Reuse the shared `_strip_trailing_semicolon` helper

`query()` uses inline `sql.rstrip().rstrip(";")`, but three connectors already have a regex-based `_strip_trailing_semicolon` for exactly this LIMIT-wrap problem. The inline version is also weaker — it doesn't fully clean interleaved separators:

```
"SELECT 1 ; ; "  →  "SELECT 1 ; "   # stray ; survives → invalid wrapped subquery
```

Prefer promoting/reusing the existing helper over a fourth, less-robust inline variant.

### 🟢 What's good
- The core `.lower().endswith` fix is correct and minimal.
- Alias uniquification is sound: lowercased-alias tracking gives `warehouse.duckdb` / `warehouse.DUCKDB` → `warehouse` / `warehouse_1`, and it handles a real `warehouse_1.duckdb` colliding with the generated suffix. `sorted()` makes assignment deterministic — a genuine improvement.
- Tests are clean (bypass the real connection via `__new__`) and cover the happy paths for all four changes.

Bartok9:
Thanks for the thorough review @goldmedal — all three addressed in 89bb6cbd.

**1. Title & description scope** — Updated both to reflect the real diff: case-insensitive discovery + deterministic alias dedup + `query`/`dry_run` semicolon/multi-statement handling. No more "one-line change" framing.

**2. `dry_run` false positives (🔴)** — Good catch, `";" in stripped` did reject valid statements like `SELECT ';' AS x`. Dropped the multi-statement pre-check entirely and switched to the same neutralization the other connectors use: wrap in `SELECT * FROM ({stripped}) AS _q LIMIT 0`. Any trailing statement becomes a natural syntax error inside the subquery, no rows materialize, and semicolons inside string literals are preserved. Added a regression test for the string-literal case.

**3. Shared helper (🟡)** — `query()` and `dry_run()` now both call a `_strip_trailing_semicolon` helper using the same `re.compile(r"[;\s]+\Z")` pattern as `canner`/`clickhouse`, which also fixes the interleaved-separator weakness (`"SELECT 1 ; ; "` → `"SELECT 1"`).

Full suite green: `test_duckdb_file_listing.py` + `test_engine.py` = 15 passed, ruff clean.

goldmedal (approved):
@Bartok9 LGTM

PR #2412, opened 2026-06-30
fix(clickhouse): default port-less clickhouse+https URLs to 8443, not 8123

Bartok9:
## What

A `clickhouse+https://` connection URL **without an explicit port** is parsed into client kwargs with `secure=True` but `port=8123` — the plaintext HTTP port. ClickHouse serves HTTPS on **8443** and plaintext HTTP on 8123, so the TLS client dials the non-TLS listener and the handshake fails.

This fixes `_build_clickhouse_client_kwargs` in the URL branch to derive the port-less default from the scheme: `8443` for HTTPS, `8123` for plaintext HTTP. An explicit port in the URL is still respected.

## Why it's a bug

The scheme says "use TLS" but the default port says "plaintext" — they contradict each other, so a perfectly valid `clickhouse+https://user:pw@host/db` URL (no port) is unconnectable.

## Evidence

**Before (unpatched)** — new test fails:
```
>       assert out["port"] == 8443
E       assert 8123 == 8443
tests/connectors/test_clickhouse.py:319: AssertionError
1 failed, 2 passed
```

**After (patched)** — all pass:
```
$ pytest tests/connectors/test_clickhouse.py -k "Kwargs or Url"
6 passed, 46 deselected
```

## Scope

- `core/wren/src/wren/connector/clickhouse.py` — derive default port from scheme (1 change)
- `core/wren/tests/connectors/test_clickhouse.py` — `TestClickHouseUrlKwargs` covering https→8443, http→8123, and explicit-port-respected

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ClickHouse URL handling so secure connections are detected correctly from the URL scheme.
  * Default ports now match the connection type: secure URLs use 8443, non-secure URLs use 8123.
  * Explicit ports provided in the URL are preserved while still applying the correct secure setting.

### DIFF
--- a/core/wren/src/wren/connector/clickhouse.py
+++ b/core/wren/src/wren/connector/clickhouse.py
@@ -261,9 +261,15 @@ def _build_clickhouse_client_kwargs(connection_info: Any) -> dict:
         # urlparse leaves percent-encoded characters in userinfo, so decode
         # them before clickhouse-connect sees the credentials. Matches the
         # mssql / postgres URL handling elsewhere in this package.
+        secure = parsed.scheme == "clickhouse+https"
+        # Pick the port-less default from the scheme: ClickHouse listens for
+        # HTTPS on 8443 and plaintext HTTP on 8123. Defaulting an https URL to
+        # 8123 silently dialed the plaintext port with ``secure=True``, so the
+        # TLS handshake hit a non-TLS listener and the connection failed.
+        default_port = 8443 if secure else 8123
         out: dict = {
             "host": parsed.hostname,
-            "port": int(parsed.port) if parsed.port else 8123,
+            "port": int(parsed.port) if parsed.port else default_port,
             "username": (
                 unquote_plus(parsed.username) if parsed.username else "default"
             ),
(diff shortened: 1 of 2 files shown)


Bartok9:
Thanks @coderabbitai — good catch on the `secure` override interaction. I looked at it and chose to keep this PR scoped to the scheme-derived default (the actual bug: `clickhouse+https` port-less URLs dialing 8123 with `secure=True`).

The `kwargs["secure"]=True` / `?secure=true` path you flag is a real but separate, pre-existing concern: it can flip TLS on after a plaintext default port was already chosen. Folding that in would mean re-deriving the port *after* `out.update(kwargs)` and reconciling it against any explicit port the user also passed — that's a behavior change with its own edge cases (explicit port + secure override) and deserves its own PR + tests rather than riding along here.

Keeping this change minimal and focused on the scheme case it advertises. Happy to open a follow-up for the kwargs-override path if maintainers want it tracked.

Bartok9:
@coderabbitai yes please — go ahead and open an issue to track the `kwargs["secure"]` / `?secure=true` port-reconciliation case so it doesn't get lost. Thanks!

goldmedal (approved):
@Bartok9 nice catch 👍

PR #2418, opened 2026-06-30
feat(memory): add 'wren memory watch' to auto-reindex on source changes

Bartok9:
## Problem

The semantic memory index (`wren memory index`) is a **derived artifact** built from a project's sources of truth — the compiled MDL manifest (`target/mdl.json`) and the NL→SQL pairs under `knowledge/sql/*.md`. During active modelling these sources change frequently (you edit YAML, run `wren context build`, `wren memory store` a new pair, etc.). Today the index only refreshes when you *remember* to re-run `wren memory index`. In between, semantic recall (`wren memory fetch` / `recall`) silently serves **stale schema context to the LLM** — the most expensive kind of bug, because it's invisible: the command succeeds, it just returns the wrong thing.

## Approach

Add `wren memory watch`: a polling loop that fingerprints the watched sources on an interval and reindexes whenever the content changes — closing the modelling→index loop automatically.

**Why this layer.** The change-detection logic lives in a new stdlib-only module `wren.memory.watch`, fully decoupled from both the CLI and the optional `memory` extra. That keeps the loop logic:
- **testable in the unit CI job** (no LanceDB/`memory` extra needed — tests are *guarded*, not `importorskip`-skipped), and
- **injectable** — `watch_loop` takes an injected `sleep` and a `reindex` callback, so the CLI wires in the real LanceDB reindex while tests drive the loop deterministically with zero real delays.

**Reliability detail (no silent drops).** `poll_once` only advances the fingerprint baseline *after* a clean reindex. If the reindex callback raises, the error is counted and surfaced, but the change stays **pending** and is retried on the next poll — a transient reindex failure can never silently swallow an update. This mirrors the repo's "escalate, don't fail silently" instinct.

Fingerprint = sha256 over each watched file's project-relative path + size + mtime_ns. That detects adds/edits/deletes without reading file bodies, so each poll stays cheap even on large projects.

The CLI guards the grep backend (no derived index → nothing to watch) and exits with a clear message pointing at `wrenai[memory]`.

## Files changed

| File | Change |
|------|--------|
| `core/wren/src/wren/memory/watch.py` | **New.** Stdlib-only loop: `compute_fingerprint`, `poll_once`, `watch_loop`, `WatchState`. |
| `core/wren/src/wren/memory/cli.py` | New `wren memory watch` command (`--interval`, `--reindex-on-start`, `--max-polls`, `--mdl`, `--path`). |
| `core/wren/tests/unit/test_memory_watch.py` | **New.** 11 unit tests — placed in `tests/unit/` so they run in the `unit tests` CI job. |
| `core/wren/README.md` | Document the command in the memory section. |

## Test evidence

Tests live in `tests/unit/` (the unit job runs the whole tree, no extra needed) — **not** behind `importorskip`, so the feature is guarded in CI.

```
$ pytest tests/unit/test_memory_watch.py -v
tests/unit/test_memory_watch.py::test_fingerprint_stable_when_unchanged PASSED
tests/unit/test_memory_watch.py::test_fingerprint_changes_on_new_pair PASSED
tests/unit/test_memory_watch.py::test_fingerprint_changes_on_mdl_edit PASSED
tests/unit/test_memory_watch.py::test_fingerprint_empty_project_is_deterministic PASSED
tests/unit/test_memory_watch.py::test_poll_once_no_change_does_not_reindex PASSED
tests/unit/test_memory_watch.py::test_poll_once_reindexes_on_change PASSED
tests/unit/test_memory_watch.py::test_failed_reindex_keeps_change_pending PASSED
tests/unit/test_memory_watch.py::test_watch_loop_demonstrates_autoreindex PASSED
tests/unit/test_memory_watch.py::test_watch_loop_reindex_on_start PASSED
tests/unit/test_memory_watch.py::test_watch_loop_clamps_interval PASSED
tests/unit/test_memory_watch.py::test_cli_watch_grep_backend_exits PASSED
============================== 11 passed in 0.29s ==============================
```

`test_watch_loop_demonstrates_autoreindex` is the **feature-working demonstration**: it drives the loop with an injected sleep that edits a watched source mid-loop, then asserts the watcher detects the change and reindexes exactly once.

Full unit suite stays green (`786 passed, 1 skipped`), and `ruff format --check` + `ruff check` pass on all touched files.

## Summary by CodeRabbit

* **New Features**
  * Added `wren memory watch` to monitor project semantic-search inputs (`target/mdl.json` and `knowledge/sql/*.md`) and automatically refresh the memory index when they change.
  * Supports `--reindex-on-start/--no-reindex-on-start` and `--max-polls`.
* **Documentation**
  * Updated CLI docs/reference and Quick Start to describe watch-mode behavior (including `grep` backend limitations).
* **Bug Fixes**
  * If reindexing fails, the change is kept pending and retried on the next poll (no silent drop).
* **Tests**
  * Added unit tests covering fingerprint stability, reindex triggering, failure retry, interval clamping, and the `grep` backend exit behavior.

### DIFF
--- a/core/wren/README.md
+++ b/core/wren/README.md
@@ -140,6 +140,7 @@ wren memory index                              # index MDL schema
 wren memory fetch -q "customer order price"    # fetch relevant schema context
 wren memory store --nl "top customers" --sql "SELECT ..."  # store NL→SQL pair
 wren memory recall -q "best customers"         # retrieve similar past queries
+wren memory watch                              # auto-reindex on schema/query changes
 ```
 
 **7. (Optional) Build a shareable GenBI app** — turn the context layer into a
--- a/core/wren/docs/cli.md
+++ b/core/wren/docs/cli.md
@@ -111,6 +111,31 @@ wren memory index                          # uses ~/.wren/mdl.json
 wren memory index --mdl /path/to/mdl.json  # explicit MDL file
 ```
 
+### `wren memory watch`
+
+Watch project sources and auto-reindex on change so semantic recall never serves a stale
+schema while you are actively modelling. Polls `target/mdl.json` and `knowledge/sql/*.md`
+on an interval; when their content fingerprint changes it runs the equivalent of
+`wren memory index`. A failed reindex leaves the change pending and is retried on the next
+poll — an update is never silently dropped. Runs until `Ctrl+C`.
+
+Requires the `memory` extra. With the grep backend there is no derived index to keep
+fresh, so this command exits with a message.
+
+| Flag | Description |
+|------|-------------|
+| `--interval`, `-i` | Seconds between polls (min 1). Default: `5`. |
+| `--reindex-on-start` / `--no-reindex-on-start` | Reindex once on startup before watching. Default: off. |
+| `--max-polls` | Stop after N polls (mainly for scripting/testing). Default: run until Ctrl+C. |
+| `--mdl` | Explicit MDL file (must live under the watched project root). |
+| `--path` | Project root to watch. Defaults to the discovered project. |
+
+```bash
+wren memory watch                    # poll every 5s, reindex on change
+wren memory watch -i 2                # poll every 2s
+wren memory watch --reindex-on-start  # ensure the index is fresh before the first interval
+```
+
 ### `wren memory describe`
 
 Print the full schema as structured plain text. No embedding or LanceDB required — this is a pure transformation of the MDL manifest into a human/LLM-readable format.
--- a/core/wren/src/wren/memory/cli.py
+++ b/core/wren/src/wren/memory/cli.py
@@ -560,6 +560,112 @@ def check(
 
 
 @memory_app.command()
+def watch(
+    mdl: MdlOpt = None,
+    path: PathOpt = None,
+    interval: Annotated[
+        float,
+        typer.Option(
+            "--interval",
+            "-i",
+            help="Seconds between polls (min 1).",
+        ),
+    ] = 5.0,
+    reindex_on_start: Annotated[
+        bool,
+        typer.Option(
+            "--reindex-on-start/--no-reindex-on-start",
+            help="Reindex once on startup before watching.",
+        ),
+    ] = False,
+    max_polls: Annotated[
+        Optional[int],
+        typer.Option(
+            "--max-polls",
+            help="Stop after N polls (mainly for scripting/testing). "
+            "Default: run until Ctrl+C.",
+        ),
+    ] = None,
+) -> None:
+    """Watch project sources and auto-reindex memory on change.
+
+    Polls ``target/mdl.json`` and ``knowledge/sql/*.md`` on an interval; when
+    their content fingerprint changes, runs the equivalent of
+    ``wren memory index`` so semantic recall never serves a stale schema while
+    you are actively modelling. A reindex that fails leaves the change pending
+    and is retried on the next poll — an update is never silently dropped.
+
+    Requires the ``memory`` extra (the index it maintains is LanceDB-backed).
+    With the grep backend there is no derived index to keep fresh.
+    """
+    from wren.context import discover_project_path  # noqa: PLC0415
+    from wren.memory.index_backend import resolve_backend  # noqa: PLC0415
+    from wren.memory.watch import watch_loop  # noqa: PLC0415
+
+    if resolve_backend() == "grep":
+        typer.echo(
+            "grep backend: knowledge/sql/ IS the index — nothing to watch. "
+            "Install `wrenai[memory]` for a derived index to keep fresh.",
+            err=True,
+        )
+        raise typer.Exit(1)
+
+    try:
+        project_path = discover_project_path()
+    except SystemExit as e:
+        typer.echo(str(e), err=True)
+        raise typer.Exit(1)
+
+    def _reindex() -> None:
+        manifest = _load_manifest(mdl)
+        mem_store = _get_store(path)
+        result = mem_store.index_schema(manifest, seed_queries=True)
+        from wren.memory.markdown import load_query_pairs  # noqa: PLC0415
+
+        md_pairs = load_query_pairs(project_path)
+        loaded = 0
+        if md_pairs:
+            res = mem_store.load_queries(md_pairs, upsert=True)
+            loaded = res["loaded"] + res["updated"]
+        typer.echo(
+            f"Reindexed {result['schema_items']} schema item(s)"
+            + (f", {loaded} pair(s)" if loaded else "")
+            + "."
+        )
+
+    def _on_event(event: str) -> None:
+        if event == "change-detected":
+            typer.echo("Change detected — reindexing...", err=True)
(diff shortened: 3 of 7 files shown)


goldmedal (changes requested):
Nice, well-scoped feature: the stdlib-only `watch.py` split, dependency injection of `sleep`/`reindex`, and guarded-in-CI tests are exactly the right shape. One **blocking bug** that defeats the PR's headline guarantee, plus a couple of minor notes.

### 🔴 Blocking: a reindex failure crashes the watcher — it does *not* "retry on the next poll"

The PR body, the `watch.py` module docstring, the `poll_once` docstring, and the CLI event message all promise the same reliability property:

> A reindex that fails leaves the change pending and is retried on the next poll — an update is never silently dropped.

But that property is not realized in the actual command. `poll_once` **re-raises** on a failed reindex:

```python
except Exception:  # noqa: BLE001 — surface count, keep change pending for retry
    state.errors += 1
    if on_event is not None:
        on_event("reindex-error")
    raise
```

…and `watch_loop` only catches `KeyboardInterrupt`:

```python
try:
    while max_polls is None or state.polls < max_polls:
        poll_once(project_path, state, reindex, on_event=on_event)  # <- raised exception escapes here
        ...
        sleep(interval)
except KeyboardInterrupt:
    ...
```

So the first time `_reindex` raises (e.g. LanceDB hiccup, a transient lock, a malformed `mdl.json` caught mid-write — precisely the "active modelling" window this feature targets), the exception propagates straight out of `watch_loop` and `wren memory watch` dies with a traceback. There is no next poll. The `"reindex-error"` CLI message ("change kept pending, will retry next poll") is effectively dead code in production — the process is already unwinding past it.

The test cited as covering this (`test_failed_reindex_keeps_change_pending`) only exercises `poll_once` in isolation and then *manually* calls it a second time. Nothing drives a reindex failure **through `watch_loop`**, so the suite gives false confidence on the one behavior the PR sells hardest.

**Fix** (keeps `poll_once`'s raise contract, which the tests assert): catch in the loop and continue.

```python
while max_polls is None or state.polls < max_polls:
    try:
        poll_once(project_path, state, reindex, on_event=on_event)
    except Exception:  # noqa: BLE001 — error counted+surfaced in poll_once; keep watching
        pass
    if max_polls is not None and state.polls >= max_polls:
        break
    sleep(interval)
```

Please also add a `watch_loop`-level test: reindex raises on poll 1, succeeds on poll 2; assert `state.reindexes == 1`, `state.errors == 1`, and that the loop ran to `max_polls` (i.e. it did *not* abort).

### 🟡 Minor

- **`reindex_on_start=True` hits the same crash on a fresh project.** With `--reindex-on-start` before `target/mdl.json` exists, the first `poll_once` fires `_reindex` → `_load_manifest(mdl)` raises → same unguarded propagation. Fixing the loop above resolves this too, but worth a test.
- **Fingerprint can miss a same-size edit within the same mtime tick.** `path + st_size + st_mtime_ns` is cheap and correct in the overwhelming majority of cases, but a content edit that preserves byte size *and* lands within the filesystem's mtime resolution is invisible. `mtime_ns` makes this near-nil in practice — just calling it out as a known limitation, not asking for a content hash.
- **`_reindex` ignores the instructions that `memory index` folds in.** The `index` command merges `manifest["_instructions"]` via `load_rules`, but `watch`'s `_reindex` doesn't, so an auto-reindex produces a slightly different index than an explicit `wren memory index`. Probably fine for a watch loop, but worth a deliberate call rather than silent divergence.

### ✅ Looks good
- Clean decoupling of change-detection from the CLI/`memory` extra; genuinely unit-testable.
- `poll_once` advancing the baseline only after a clean reindex is the right invariant.
- Grep-backend guard with an actionable message; interval clamping; injected `sleep`.

**Verdict:** request changes — the loop-level retry gap is the one real issue, and it's central because it invalidates the PR's stated reliability guarantee. The rest is polish.

Bartok9:
Thanks @goldmedal — this is exactly right, and it's the bug I'd most want caught. Your review landed against `a8f15b0c`; I'd just pushed `ebcfeac1` which fixes precisely this, so a quick re-pull should show all three points addressed:

**🔴 Blocking retry gap:** `watch_loop` now wraps `poll_once` in `try/except Exception`, emits an `"error"` event, and `continue`s to the next interval instead of unwinding. `poll_once` keeps its raise contract (so the isolated tests still hold) and leaves the fingerprint unchanged on failure, so the same change is retried next poll. Added `test_watch_loop_survives_transient_reindex_failure`: reindex raises on poll 1, succeeds on poll 2, and asserts `reindexes == 1`, `errors == 1`, and that the loop ran to `max_polls` rather than aborting — driving the failure *through* `watch_loop` as you asked.

**🟡 `reindex_on_start` on a fresh project:** same loop guard now covers the startup `_reindex` → `_load_manifest` raise; `test_watch_loop_reindex_on_start` exercises it.

**🟡 Cross-project `--mdl`:** added a guard so `--mdl` and the `knowledge/sql` root must resolve to the same project, failing fast otherwise.

On the two you flagged as call-outs rather than asks — the same-size/same-mtime-tick fingerprint blind spot and `_reindex` not folding in `_instructions` — happy to follow either up. The instructions divergence in particular I'd lean toward closing so `watch` and explicit `memory index` produce identical output; say the word and I'll fold `load_rules` into `_reindex` in this PR or a follow-up. CI is green/running on `ebcfeac1`.

goldmedal:
Confirmed the reindex-failure fix (`ebcfeac`) — `watch_loop` now catches, emits `error`, keeps the change pending, and retries on the next poll, and `test_watch_loop_survives_transient_reindex_failure` drives that path through the loop. 👍

One follow-up before merge: the new `wren memory watch` command is only documented in `core/wren/README.md`. The two per-subcommand CLI reference pages in this repo still list `index` / `fetch` / `store` / `recall` (and `export` / `check`) but not `watch`:

- `docs/core/reference/cli.md` — add a `### wren memory watch` section (it already documents `export` and `check`, so `watch` belongs here too)
- `core/wren/docs/cli.md` — same

Worth listing the flags in the new section (`--interval` / `-i`, `--reindex-on-start`, `--max-polls`, `--mdl`, `--path`).

Also `core/wren/src/wren/skills_content/usage/references/memory.md` — that skill doc (shipped to the agent) tells you to re-run `wren memory index` "when `wren memory fetch` returns stale results," which is exactly what `watch` automates. A one-line pointer to `wren memory watch` there would keep the agent guidance consistent.

Bartok9:
Done in `281b06b3` — added a `### wren memory watch` section (with the full flag table: `--interval`/`-i`, `--reindex-on-start`, `--max-polls`, `--mdl`, `--path`) to both `docs/core/reference/cli.md` and `core/wren/docs/cli.md`, placed right after `wren memory index`. Also added a pointer to `wren memory watch` in `core/wren/src/wren/skills_content/usage/references/memory.md` under the indexing section, so the agent guidance no longer stops at 'manually re-run index when fetch is stale.' Thanks for the thorough review.

goldmedal (approved):
Thanks @Bartok9, Look great 👍

PR #2419, opened 2026-06-30
feat(policy): govern data-reading TVFs in all positions + structure source-func allowlist

### LINKED ISSUE #2409: strict mode: govern data-reading TVFs in all positions + structure the source-func allowlist
## Background

`core/wren/src/wren/policy.py` `_check_tables` enforces strict mode: every table/source must correspond to an MDL model. PR #2405 widened the table-valued-function (TVF) source scan from `FROM` to `FROM`+`JOIN` (and unwrapped `LATERAL`). This issue tracks two follow-ups that are out of scope for that PR.

## C. Data-reading TVFs are only governed in the *top-level source* position

Strict mode blocks a data-reading TVF (e.g. duckdb `read_csv` / `read_parquet` / `read_json` / `glob`, `dblink`, external table functions) **only when it appears as a `FROM`/`JOIN` source**. It does **not** govern the same function elsewhere. Verified on `main` and on the #2405 branch (`strict_mode=True`, manifest = `{orders}`):

| SQL | Result |
|---|---|
| `SELECT * FROM read_csv('/etc/passwd')` | **BLOCKED** ✅ |
| `SELECT read_csv('/etc/passwd')` (projection) | **ALLOWED** ⚠️ |
| `SELECT * FROM orders WHERE id IN (SELECT read_csv('/etc/passwd'))` | **ALLOWED** ⚠️ |
| `SELECT * FROM orders CROSS JOIN UNNEST(read_csv('/etc/passwd')) t(i)` | nested arg, not scanned ⚠️ |

**Risk:** a data-reading TVF in a non-source position bypasses MDL governance — reads data outside the manifest (defeats RLAC/CLAC), enables path traversal (`read_csv('/etc/passwd')`, `~/.aws/credentials`) and SSRF / external reads (URLs, S3 paths).

**Direction:** govern data-reading TVFs in **all** AST positions (projection, WHERE/subqueries, nested function args), not just the source slot. Keep it fail-closed.

## #2. Structure the source-func allowlist by category (don't use one flat list)

TVFs aren't homogeneous — three categories, each governed differently:

| Category | Examples | Risk | Approach |
|---|---|---|---|
| **A. Row-expansion ops** | `UNNEST` / `FLATTEN` / `EXPLODE` | none (no external read) | **class allowlist** (`exp.Unnest`, `exp.Explode`) — closed, dialect-stable, no name matching. (Being added for the `UNNEST(model.column)` false-block, see #2405.) |
| **B. Synthetic generators** | `generate_series` / `sequence` / `range` | no exfiltration, but **DoS** via huge ranges | keep blocked by default; if allowed, make it a config opt-in with a **name allowlist** |
| **C. Data/file readers** | `read_csv` / `read_parquet` / `glob` / `dblink` | **real** (governance bypass, traversal, SSRF) | always block, in all positions (see section C) |

For **B**'s name allowlist, reuse the existing cross-dialect canonicalization (`_canonical_denied` / `_CANONICALISE_DIALECTS`, policy.py:155-200) — mirror it as `_canonical_allowed(frozenset[str])` so a name like `generate_series` matches whether sqlglot lands it on `exp.GenerateSeries` or `exp.Anonymous`. Principle: **allowlist, never denylist** for readers (a missed reader name = a hole).

## Acceptance criteria

- [ ] Data-reading TVFs are blocked in projection, WHERE/subquery, and nested-arg positions under strict mode (with tests covering each).
- [ ] Source-func handling is organised into the A/B/C categories above; row-expansion ops allowed by class, generators gated behind explicit config.
- [ ] Any reader/generator name matching goes through the shared canonicalization helper, not raw string compares.

## Refs
- PR #2405 — FROM+JOIN+LATERAL TVF source scan (the change this follows up on)

Bartok9:
Closes #2409.

## Threat addressed

`core/wren/src/wren/policy.py` strict mode enforces "every source must be an MDL model." PR #2405 widened the table-valued-function (TVF) source scan from `FROM` to `FROM`+`JOIN`+`LATERAL` — but only in the **top-level source position**. Data-reading TVFs in any *other* AST position bypassed MDL governance entirely:

| SQL (strict_mode=True, manifest={orders}) | Before | After |
|---|---|---|
| `SELECT * FROM read_csv('/etc/passwd')` | BLOCKED ✅ | BLOCKED ✅ |
| `SELECT read_csv('/etc/passwd')` (projection) | **ALLOWED ⚠️** | BLOCKED ✅ |
| `SELECT * FROM orders WHERE id IN (SELECT read_csv('/etc/passwd'))` | **ALLOWED ⚠️** | BLOCKED ✅ |
| `SELECT * FROM orders CROSS JOIN UNNEST(read_csv('s3://b/f.csv')) t(i)` | **ALLOWED ⚠️** | BLOCKED ✅ |
| `SELECT length(read_text('/etc/passwd')) FROM orders` | **ALLOWED ⚠️** | BLOCKED ✅ |

**Risk:** a data-reading TVF in a non-source position reads data outside the manifest (defeats RLAC/CLAC), enables **path traversal** (`read_csv('/etc/passwd')`, `~/.aws/credentials`), **SSRF / external reads** (`read_parquet('s3://...')`, `url(...)`), and **external-DB lateral movement** (`dblink`, `postgres_scan`).

## Approach (matches the issue's acceptance criteria)

**Section C — govern readers in ALL positions.** New `_check_data_readers` runs first under strict mode and walks the *entire* AST (`ast.find_all(exp.Func)`), failing closed on the first known data/file/remote reader regardless of position. Reader detection is matched both by raw function name (anonymous parses like `glob` / `dblink` / `pg_read_file`) and by canonical sqlglot class key (`read_csv` → `exp.ReadCSV` → `"readcsv"`).

**Section #2 — structure the source-func allowlist by category** (replaces the implicit flat handling):

| Cat | Examples | Risk | Handling |
|---|---|---|---|
| **A** row-expansion | `UNNEST`/`FLATTEN`/`EXPLODE` | none | class allowlist (`exp.Unnest`, `exp.Explode`) — unchanged |
| **B** generators | `generate_series`/`sequence`/`range` | DoS (unbounded range) | **blocked by default**; opt-in per name via new `allowed_source_functions` config |
| **C** readers | `read_csv`/`dblink`/`postgres_scan`/… | governance bypass, traversal, SSRF | **always blocked, all positions** |

Reader/generator name matching reuses the existing cross-dialect canonicalization — `_canonical_denied` was generalized to `_canonical_names` and now probes arity 0/1/2 so subclass-only-with-args readers (`read_csv('x')` → `exp.ReadCSV`, which fails to parse with zero args) are captured. **Readers can never be opted in** via `allowed_source_functions` — only generators.

### Why this layer
This is the single MDL-governance choke point for SQL sources. Fixing it here closes the bypass for *all* connectors at once without touching any connector, and keeps the policy fail-closed (allowlist for readers/generators, never denylist).

## New config (opt-in, backwards compatible)
```json
{ "strict_mode": true, "allowed_source_functions": ["generate_series"] }
```
Defaults to empty — existing configs are unaffected; generators stay blocked unless explicitly listed.

## Test evidence

New tests assert the enforcement (each fails without the fix) AND that legitimate queries still pass (no false positives). Reader-bypass tests verified failing on pre-fix code:

```
$ git stash push src/wren/policy.py src/wren/config.py   # revert the fix
$ pytest tests/unit/test_policy.py -k "projection or where_subquery or nested_in or unnest_over_reader or dblink or postgres_scan or anonymous_reader or can_never"
FAILED test_unnest_over_reader_tvf_now_blocked
FAILED test_reader_in_projection_blocked
FAILED test_reader_in_where_subquery_blocked
FAILED test_reader_nested_in_unnest_arg_blocked
FAILED test_reader_nested_in_function_arg_blocked
FAILED test_anonymous_reader_glob_in_projection_blocked
FAILED test_reader_can_never_be_opted_in_as_source_func
7 failed, 3 passed
```

With the fix (full policy + config unit suites):
```
$ pytest tests/unit/test_policy.py tests/unit/test_config.py -q
64 passed in 0.33s

$ pytest tests/unit -q
833 passed, 42 skipped in 8.92s

$ ruff check ...  -> All checks passed!
$ ruff format --check ... -> all formatted
```

Tests live in `tests/unit/test_policy.py` / `tests/unit/test_config.py` (the `pytest.mark.unit` suite that runs in the core CI job) — not behind any `importorskip`; they use sqlglot parsing only, no DB required.

## Files changed
- `core/wren/src/wren/policy.py` — all-position reader guard, A/B/C category handling, generalized canonicalizer
- `core/wren/src/wren/config.py` — new `allowed_source_functions` field + validation
- `core/wren/tests/unit/test_policy.py` — reader-in-all-positions, generator opt-in, no-false-positive tests
- `core/wren/tests/unit/test_config.py` — config parsing/validation tests

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in configuration for specific generator-style functions to be used as query sources in strict mode.
* **Bug Fixes**
  * Stricter enforcement now blocks data/file reader functions across more SQL locations (including nested expressions and joins), while continuing to honor the denylist when strict mode is off.
  * UNNEST over reader TVFs is now blocked in strict mode.
  * Improved function-name matching for consistent policy decisions.
* **Tests**
  * Expanded strict-mode and opt-in coverage, including configuration validation and deterministic error expectations.

### DIFF
--- a/core/wren/src/wren/config.py
+++ b/core/wren/src/wren/config.py
@@ -21,10 +21,17 @@ class WrenConfig:
     denied_functions:
         Set of function names (lowercase) that are forbidden in SQL queries.
         Matching is case-insensitive.
+    allowed_source_functions:
+        Set of synthetic-generator function names (lowercase) the operator
+        explicitly opts in as query sources under strict mode (e.g.
+        ``generate_series``). Empty by default — generators are blocked unless
+        listed here. Data/file readers (``read_csv``, ``dblink``, ...) can
+        NEVER be allowed via this list; they are always blocked in strict mode.
     """
 
     strict_mode: bool = False
     denied_functions: frozenset[str] = field(default_factory=frozenset)
+    allowed_source_functions: frozenset[str] = field(default_factory=frozenset)
 
 
 def load_config(wren_home: Path) -> WrenConfig:
@@ -71,4 +78,21 @@ def load_config(wren_home: Path) -> WrenConfig:
         )
     denied_functions = frozenset(f.lower() for f in denied_raw)
 
-    return WrenConfig(strict_mode=strict_mode_raw, denied_functions=denied_functions)
+    allowed_src_raw = raw.get("allowed_source_functions", [])
+    if not isinstance(allowed_src_raw, list):
+        raise WrenError(
+            ErrorCode.GENERIC_USER_ERROR,
+            f"{config_path}: 'allowed_source_functions' must be a JSON array.",
+        )
+    if any(not isinstance(f, str) for f in allowed_src_raw):
+        raise WrenError(
+            ErrorCode.GENERIC_USER_ERROR,
+            f"{config_path}: 'allowed_source_functions' must contain only strings.",
+        )
+    allowed_source_functions = frozenset(f.lower() for f in allowed_src_raw)
+
+    return WrenConfig(
+        strict_mode=strict_mode_raw,
+        denied_functions=denied_functions,
+        allowed_source_functions=allowed_source_functions,
+    )
--- a/core/wren/src/wren/policy.py
+++ b/core/wren/src/wren/policy.py
@@ -23,12 +23,76 @@ from wren.model.error import ErrorCode, ErrorPhase, WrenError
 # ``exp.Unnest`` (trino/postgres/duckdb), Snowflake ``FLATTEN`` -> ``exp.Explode``,
 # so no per-dialect name matching is needed.
 #
-# Note: ``UNNEST(read_csv(...))`` is therefore also allowed, but that is an
-# instance of the broader pre-existing gap (data-reading TVFs in non-source
-# positions — already reachable today via projection/WHERE subqueries, since
-# strict mode only governs the top-level source position) and is tracked
-# separately. It is not a regression introduced here.
+# Note: ``UNNEST(read_csv(...))`` used to be allowed because the data-reader
+# guard only governed the top-level source position. As of the all-positions
+# data-reader check below (issue #2409 section C), the inner ``read_csv`` is now
+# blocked regardless of being wrapped by UNNEST — the reader scan walks every
+# AST position, so it sees the nested reader before the row-expansion allow-list
+# is ever consulted.
 _ROW_EXPANSION_FUNCS: tuple[type[exp.Func], ...] = (exp.Unnest, exp.Explode)
+
+# ── Category C: data/file readers (issue #2409) ─────────────────────────────
+#
+# Data-reading table-valued functions read bytes from *outside* the MDL
+# manifest: local files (``read_csv('/etc/passwd')`` → path traversal), object
+# storage / URLs (``read_parquet('s3://...')`` → SSRF / exfiltration), or other
+# databases (``dblink`` / ``postgres_scan`` → lateral movement). Allowing any of
+# them defeats strict-mode governance (RLAC/CLAC), so they are ALWAYS blocked
+# under strict mode, in EVERY AST position — not just the FROM/JOIN source slot
+# (which is all PR #2405 covered). Reader detection is fail-closed: this list is
+# the source of truth for *known* reader names, matched by both the raw
+# function name (for ``exp.Anonymous`` parses like ``glob`` / ``dblink``) and by
+# the canonical sqlglot class key (e.g. ``read_csv`` → ``exp.ReadCSV`` →
+# ``"readcsv"``), via the same dialect-probe used for the denylist.
+_DATA_READER_NAMES: frozenset[str] = frozenset(
+    {
+        # duckdb file readers
+        "read_csv",
+        "read_csv_auto",
+        "read_parquet",
+        "read_json",
+        "read_json_auto",
+        "read_ndjson",
+        "read_ndjson_auto",
+        "read_json_objects",
+        "read_text",
+        "read_blob",
+        "read_xlsx",
+        "parquet_scan",
+        "glob",
+        # duckdb extension scanners (lakehouse / external db)
+        "iceberg_scan",
+        "delta_scan",
+        "postgres_scan",
+        "postgres_query",
+        "mysql_scan",
+        "mysql_query",
+        "sqlite_scan",
+        "sqlite_query",
+        # postgres file / remote readers
+        "pg_read_file",
+        "pg_read_binary_file",
+        "pg_ls_dir",
+        "lo_import",
+        "lo_get",
+        "dblink",
+        "dblink_exec",
+        # generic external fetch
+        "url",
+    }
+)
+
+# ── Category B: synthetic generators (issue #2409) ──────────────────────────
+#
+# Generators (``generate_series`` / ``sequence`` / ``range``) read nothing
+# outside the manifest, so they are not an exfiltration risk — but an unbounded
+# range is a denial-of-service vector (``generate_series(1, 1e12)`` materialises
+# a trillion rows). They are therefore blocked by default in source position and
+# only allowed when the operator explicitly opts a name in via
+# ``allowed_source_functions``. Matching goes through the shared canonicalizer
+# so an opt-in of ``generate_series`` matches whether sqlglot lands it on
+# ``exp.GenerateSeries`` / ``exp.ExplodingGenerateSeries`` or ``exp.Anonymous``.
+_GENERATOR_NAMES: frozenset[str] = frozenset({"generate_series", "sequence", "range"})
 
 # Dialects we probe when canonicalising the user's denylist. sqlglot can map
 # the same function name (e.g. ``version()``) onto different concrete AST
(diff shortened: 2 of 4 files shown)


goldmedal:
Strong PR — the threat model is articulated well, the source-position path is genuinely fail-closed, and the arity-0/1/2 canonicalization probe (so `read_csv('x')` → `exp.ReadCSV` is captured) is a nice catch. Tests are thorough and demonstrably fail on pre-fix code. One substantive gap worth addressing, plus a framing note.

## The all-position reader guard is a denylist, not the "fail-closed allowlist" the description claims — and it has real holes

The PR/issue frame the change as *"fail-closed (allowlist for readers/generators, never denylist)."* That holds for the **source position** (`_check_tables` blocks any unknown TVF), but **not** for the new all-positions guard: `_check_data_readers` matches against the hardcoded `_DATA_READER_NAMES` set. Anything not enumerated there passes in projection / WHERE-subquery / nested-arg positions — precisely the class of bypass this PR exists to close.

I ran the PR branch against readers that aren't in the list:

```
ALLOWED   mysql     SELECT load_file('/etc/passwd')
ALLOWED   mysql     SELECT load_file('/etc/passwd') FROM orders
ALLOWED   postgres  SELECT pg_stat_file('/etc/passwd')
ALLOWED   postgres  SELECT pg_ls_waldir()
ALLOWED   duckdb    SELECT sniff_csv('/etc/passwd') FROM orders
ALLOWED   duckdb    SELECT st_read('/x.shp') FROM orders
-- listed readers, correctly blocked:
BLOCKED   duckdb    SELECT read_csv('/etc/passwd')
BLOCKED   duckdb    SELECT length(read_text('/etc/passwd')) FROM orders
```

`SELECT load_file('/etc/passwd')` is the **same arbitrary-file-read primitive** the PR closes for DuckDB, just via MySQL — a first-class connector — and it's wide open in a projection. `pg_stat_file` / `pg_ls_dir` / `pg_ls_waldir` (Postgres filesystem enumeration) and duckdb `sniff_csv` / `st_read` (spatial reader over arbitrary paths) are missed too. (`parquet_metadata` shows blocked above only because it sits in a `FROM` slot where the fail-closed `_check_tables` catches it — move it to a projection and it leaks like the rest.)

Two suggestions:

1. **Correct the framing.** For non-source positions this is inherently a *blocklist* — you can't allowlist every scalar function in a projection. That's a legitimate design, but the description and the module docstring shouldn't call it a fail-closed allowlist, because operators will over-trust it. Better to state plainly: source position is fail-closed; non-source positions are a best-effort known-reader blocklist.
2. **Close the obvious cross-connector omissions.** At minimum add MySQL `load_file`, and the Postgres `pg_stat_file` / `pg_ls_dir` / `pg_ls_logdir` / `pg_ls_waldir` / `pg_ls_tmpdir` family; consider duckdb `sniff_csv`, `st_read`/`st_readosm`, and the `*_metadata` / `*_snapshots` scanners. A projection-position test for at least `load_file` would pin the MySQL vector.

Since the list is now the security boundary for non-source positions, a comment noting it must be maintained per-connector — and pointing operators at `denied_functions` as a defense-in-depth backstop for anything the built-in list misses — would help.

## Looks good
- Source-position handling is genuinely fail-closed (unknown TVF → blocked), so the denylist gap is bounded to non-source positions.
- Category B generator opt-in via `allowed_source_functions`, with readers explicitly non-opt-in-able (`test_reader_can_never_be_opted_in_as_source_func`) — correct and well-tested.
- Arity-probing canonicalizer + raw-name/class-key dual match handles the anonymous-vs-subclass dialect split cleanly.
- The `sequence`/`generate_series` class-collapse caveat is documented rather than hidden.
- Config validation is strict and backwards-compatible.

Overall the design is right; the main ask is to expand the reader list for the supported connectors and adjust the wording so the guarantee isn't oversold.

Bartok9:
Thanks — this is exactly right, and the concrete bypass table made it trivial to reproduce. Pushed `3127eeaa`:

**Framing corrected.** The module docstring now states plainly: source position (FROM/JOIN) is fail-closed via `_check_tables`; non-source positions are a *best-effort per-connector blocklist* keyed off `_DATA_READER_NAMES`, which is the security boundary there and must be maintained per-connector. Also points operators at `denied_functions` as a defense-in-depth backstop for anything the built-in list misses.

**Closed the omissions you found.** Added to the reader set:
- MySQL `load_file` (the same arbitrary-file-read primitive as `read_csv`, just via a first-class connector)
- Postgres `pg_stat_file`, `pg_ls_waldir`, `pg_ls_logdir`, `pg_ls_tmpdir`, `pg_ls_archive_statusdir`
- duckdb `sniff_csv`, `st_read`, `st_readosm`, and the `parquet_*`/`iceberg_*` metadata/snapshot scanners

All six ALLOWED rows from your run are now BLOCKED (including the projection-position `load_file` vector, which now has a dedicated test).

**Bonus (CodeRabbit):** the reader error echoed `func.sql()`, leaking the path/URL/DSN into messages and logs — now reports only the function name.

Added regression tests: projection-position blocks for `load_file` + the pg/duckdb families, an error-message no-leak assertion, and tightened the false-positive test to use a quoted `read_csv` identifier/alias so it actually exercises the identifier-vs-func distinction. 68 tests green, ruff clean.

goldmedal (approved):
Thanks @Bartok9 look great 👍

PR #2420, opened 2026-06-30
fix(redshift): strip trailing semicolon before subquery-wrapping

Bartok9:
## Summary

- The Redshift connector wraps user SQL as `SELECT * FROM (...) AS _q LIMIT N` (in both `query()` and `dry_run()`). When the user SQL ends in a `;`, Redshift rejects it inside the subquery with `syntax error at or near ";"`.
- User-facing impact: any saved query / preview SQL that ends in a semicolon fails on Redshift even though it runs fine standalone.

## Motivation

Mirrors the postgres fix in #2407 (merged). The same subquery-wrap idiom is used across the postgres/canner/trino/clickhouse connectors, all of which already strip a trailing semicolon before wrapping. Redshift was the remaining connector that wrapped without stripping.

The fix adds `_strip_trailing_semicolon()` (identical regex `[;\s]+\Z` to postgres) and applies it in both `query()` and `dry_run()`. Only the *terminating* run of semicolons/whitespace is stripped, so semicolons inside string literals (`SELECT 'a;b'`) are preserved.

## Verification

- `uv run --no-sync pytest tests/unit/test_redshift_semicolon.py -q` — 4 passed.

## Real behavior proof

**Regression test** `core/wren/tests/unit/test_redshift_semicolon.py` asserts the executed SQL and FAILS without the fix.

```
ImportError: cannot import name '_strip_trailing_semicolon' from 'wren.connector.redshift'
1 error in 0.41s
```

With the fix:
```
4 passed in 0.34s
```

The query test asserts the executed SQL is exactly `SELECT * FROM (SELECT 1) AS _q LIMIT 5` (no `;)`), proving the semicolon is stripped before wrapping.

## Scope / risk

- Touches only `core/wren/src/wren/connector/redshift.py` (Apache-2.0 path) + a new unit test.
- Does NOT touch type parsing, connection logic, or other connectors.
- String-literal semicolons preserved (covered by a helper test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Redshift query handling so SQL with a trailing semicolon now runs correctly when wrapped for limits or dry runs.
  * Preserved semicolons inside quoted text while removing only unnecessary trailing semicolons and whitespace.
* **Tests**
  * Added coverage for trailing-semicolon handling in query execution and dry runs.

### DIFF
--- a/core/wren/src/wren/connector/redshift.py
+++ b/core/wren/src/wren/connector/redshift.py
@@ -1,3 +1,4 @@
+import re
 from contextlib import closing
 
 import pandas as pd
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
Thanks @Bartok9

PR #2422, opened 2026-06-30
fix(databricks): strip trailing semicolon before subquery-wrapping in dry_run

Bartok9:
## Summary

- `DatabricksConnector.dry_run` wraps user SQL as `SELECT * FROM ({sql}) AS sub LIMIT 0`. When the user SQL ends in `;`, the subquery is invalid (`SELECT * FROM (SELECT 1;) AS sub LIMIT 0` → syntax error).
- User-facing impact: dry-running / validating a query that ends in a semicolon fails on Databricks even though the same query executes fine standalone.

## Motivation

The sibling connectors already strip a trailing semicolon before subquery-wrapping: postgres, canner, trino, clickhouse all have a `_strip_trailing_semicolon` helper. Databricks's `dry_run` was wrapping raw SQL. (`query()` uses `fetchmany_arrow(limit)` and never wraps, so it's unaffected.)

The fix adds `_strip_trailing_semicolon()` (same `[;\s]+\Z` regex as the postgres connector) and applies it in `dry_run`. Only the *terminating* run of semicolons/whitespace is stripped, so semicolons inside string literals (`SELECT 'a;b'`) are preserved.

## Verification

- `uv run --no-sync pytest tests/unit/test_databricks_semicolon.py -q` — 4 passed.

## Real behavior proof

**Regression test** `core/wren/tests/unit/test_databricks_semicolon.py` asserts the executed SQL and FAILS without the fix.

```
ImportError: cannot import name '_strip_trailing_semicolon' from 'wren.connector.databricks'
1 error in 0.24s
```

With the fix:
```
4 passed in 0.66s
```

The dry_run test asserts the executed SQL is exactly `SELECT * FROM (SELECT 1) AS sub LIMIT 0` (no `;)`).

## Scope / risk

- Touches only `core/wren/src/wren/connector/databricks.py` (Apache-2.0 path) + a new unit test.
- Does NOT touch `query()`, type handling, or connection logic.
- String-literal semicolons preserved (covered by a helper test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQL handling for Databricks dry runs so queries with trailing semicolons or extra whitespace now run correctly.
  * Preserved semicolons inside string values, avoiding unintended changes to valid SQL.
* **Tests**
  * Added coverage for dry-run SQL wrapping with trailing semicolons, whitespace, and unchanged queries without a trailing semicolon.

### DIFF
--- a/core/wren/src/wren/connector/databricks.py
+++ b/core/wren/src/wren/connector/databricks.py
@@ -1,3 +1,4 @@
+import re
 from contextlib import closing
 
 import pyarrow as pa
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
👍

PR #2423, opened 2026-06-30
fix(oracle): strip trailing semicolon before subquery-wrapping

Bartok9:
## Summary

- Both `OracleConnector.query` (when a limit is given) and `OracleConnector.dry_run` wrap user SQL as `SELECT * FROM ({sql}) t WHERE ROWNUM <= N`. When the user SQL ends in `;`, the subquery is invalid and Oracle raises `ORA-00911: invalid character`.
- User-facing impact: running (with a row limit) or validating a query that ends in a semicolon fails on Oracle even though the same query executes fine standalone.

## Motivation

The sibling connectors already strip a trailing semicolon before subquery-wrapping: postgres, canner, trino, clickhouse all have a `_strip_trailing_semicolon` helper. Oracle wrapped raw SQL in both code paths.

The fix adds `_strip_trailing_semicolon()` (same `[;\s]+\Z` regex as the postgres connector) and applies it in both `query()` and `dry_run()`. Only the *terminating* run of semicolons/whitespace is stripped, so semicolons inside string literals (`SELECT 'a;b'`) are preserved.

## Verification

- `uv run --no-sync pytest tests/unit/test_oracle_semicolon.py -q` — 4 passed (with the `oracle` extra installed).

## Real behavior proof

**Regression test** `core/wren/tests/unit/test_oracle_semicolon.py` asserts the executed SQL and FAILS without the fix. It `pytest.importorskip("oracledb")` so it skips cleanly when the oracle extra isn't installed.

```
ImportError: cannot import name '_strip_trailing_semicolon' from 'wren.connector.oracle'
1 error
```

With the fix:
```
4 passed in 0.21s
```

The query test asserts the executed SQL is exactly `SELECT * FROM (SELECT 1) t WHERE ROWNUM <= 5` (no `;)`); the dry_run test asserts `... WHERE ROWNUM <= 0`.

## Scope / risk

- Touches only `core/wren/src/wren/connector/oracle.py` (Apache-2.0 path) + a new unit test.
- Does NOT touch type parsing or connection logic.
- The no-limit `query()` path (which doesn't wrap) is unchanged; string-literal semicolons preserved (covered by a helper test).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Oracle query handling so statements ending with a trailing semicolon no longer fail when used in limited queries or dry runs.
  * Preserved semicolons inside string literals while only trimming unnecessary trailing semicolons and surrounding whitespace.
* **Tests**
  * Added coverage for Oracle SQL wrapping behavior, including trailing-semicolon handling and unchanged behavior when no trailing semicolon is present.

### DIFF
--- a/core/wren/src/wren/connector/oracle.py
+++ b/core/wren/src/wren/connector/oracle.py
@@ -1,5 +1,6 @@
 """Native oracledb connector — bypasses ibis oracle backend."""
 
+import re
 from decimal import Decimal as PyDecimal
 from urllib.parse import urlparse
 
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
👍

PR #2425, opened 2026-07-02
fix: finish wren-engine → wrenai rename cleanup

ttw225:
## Summary
- Quote `pip install 'wrenai[extra]'` hints so they don't break under zsh glob expansion
- Fix install hints/deps that still pointed at the wrong or pre-rename package name (`wren[extra]`, `wren-engine[trino]`, `wren-engine>=0.5.0` in wren-pydantic/wren-langchain)
- Bump `wrenai` version floor to `>=0.7.0` (0.5.0/0.6.0 were published as `wren-engine`, before the rename)
- Fix stale `wren-engine` metadata/links (Cargo.toml repo URL, CLAUDE.md, skill docs)

## Test plan
- [x] `pytest` in `core/wren` (871 passed, Docker-dependent tests excluded)
- [x] Both SDK `pyproject.toml` files validated as parseable TOML

## Summary by CodeRabbit

* **Documentation**
  * Updated installation and compatibility guidance to use the new `wrenai` naming and version requirements.
  * Clarified CLI, SDK, and skill instructions to match the current package and command names.

* **Bug Fixes**
  * Improved missing-dependency error messages to show the correct install commands with quoted extras.
  * Adjusted connector guidance so users see the right package names for optional components.

* **Tests**
  * Updated and added tests to verify the new install hints and error messaging.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -16,7 +16,7 @@ core/
 ├── wren-core-base/   Shared Rust crate — manifest types (Model, Column, Metric, Relationship, View) + ManifestBuilder
 ├── wren-core-py/     PyO3 bindings exposing wren-core to Python (PyPI: wren-core)
 ├── wren-core-wasm/   WebAssembly build of wren-core for in-browser semantic SQL (npm: wren-core-wasm)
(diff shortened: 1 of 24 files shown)


goldmedal (approved):
@ttw225 nice catch. Thanks 👍

PR #2426, opened 2026-07-02
fix(clickhouse): reconcile default port when `secure` is toggled via kwargs or query params

### LINKED ISSUE #2416: Follow up: reconcile default ClickHouse port when `secure` is enabled via kwargs or query params
## Summary
Track the separate ClickHouse URL/kwargs behavior where TLS can be enabled via `kwargs["secure"]=True` or `?secure=true` after the client kwargs have already picked the default plaintext port.

## Problem
The scoped fix in #2412 correctly derives the default port from the URL scheme when using `clickhouse+https://` without an explicit port.

However, there is a related pre-existing case where:
- the URL/scheme path initially selects the default plaintext port (`8123`), and
- a later override such as `kwargs["secure"]=True` or `?secure=true` enables TLS.

In that flow, the client can end up with `secure=True` while still targeting the plaintext default port, which can lead to connection failures or TLS handshake issues.

## Requested change
Update the ClickHouse client kwargs construction to reconcile the default port after secure-related overrides are applied, while preserving any explicit user-provided port.

## Affected area
- `core/wren/src/wren/connector/clickhouse.py`
- `core/wren/tests/connectors/test_clickhouse.py`

## Acceptance criteria
- If `secure` becomes true via kwargs or query parameters and no explicit port was provided, the default port resolves to the secure default (`8443`).
- If `secure` is false and no explicit port was provided, the default port resolves to the plaintext default (`8123`).
- If the user explicitly provides a port, that port is preserved regardless of `secure` overrides.
- Tests cover the kwargs override path and the query-parameter override path, including explicit-port edge cases.

## Backlinks
- PR: https://github.com/Canner/WrenAI/pull/2412
- Request comment: https://github.com/Canner/WrenAI/pull/2412#issuecomment-4838804079

Requested by @Bartok9.

RaidLZ:
## What

Follow-up to #2412, requested in #2416. The port-less default is derived from the URL **scheme alone**, but TLS can also be enabled *after* the scheme is inspected — via a `?secure=true` query parameter or a `kwargs["secure"]` override. In that flow the client kwargs come out as `secure=True` + `port=8123`: a TLS handshake against the plaintext listener.

This reconciles the default port with the **effective** secure value: the `secure` override is popped from the merged kwargs (kwargs win over query params, which win over the scheme — the merge order that already exists), normalised to a real boolean, and the port-less default follows it: `8443` when secure, `8123` when not. An explicit port is always preserved.

## Why it's a bug

- `clickhouse://host/db?secure=true` → `secure=True, port=8123` — TLS against the plaintext port, handshake fails.
- The inverse too: `clickhouse+https://host/db` + `kwargs={"secure": False}` → plaintext against `8443`.
- Adjacent string-truthiness bug: query parameters always arrive as strings, so `?secure=false` produced `secure="false"` — **truthy** — flipping the client into TLS mode nobody asked for. The override is now interpreted (`"false"/"0"/"no"/"off"/""` → off) and only a normalised boolean reaches the client.

## Evidence

**Before (unpatched)** — the new tests fail:
```
$ pytest tests/connectors/test_clickhouse.py -k "UrlKwargs"
...
FAILED ...::test_secure_query_param_uses_secure_default_port - assert 8123 == 8443
FAILED ...::test_secure_false_query_param_uses_plaintext_default_port
FAILED ...::test_falsy_secure_strings[false] ...
18 failed, 3 passed, 49 deselected
```

**After (patched)** — all pass:
```
$ pytest tests/connectors/test_clickhouse.py -k "UrlKwargs"
21 passed, 49 deselected

$ pytest tests/connectors/test_clickhouse.py -v -m clickhouse   # full suite, incl. containers
70 passed
```

`ruff format --check src/` and `ruff check src/` are clean.

## Scope

- `core/wren/src/wren/connector/clickhouse.py` — pop the `secure` override before picking the default port; `_parse_secure_flag` helper for string coercion
- `core/wren/tests/connectors/test_clickhouse.py` — `TestClickHouseUrlKwargs` extended per the acceptance criteria in #2416: kwargs path, query-param path, explicit-port edge cases, kwargs-over-query precedence, and truthy/falsy string forms

Closes #2416 (and its duplicate #2415).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed ClickHouse connection handling so TLS/`secure` correctly interprets common falsy string values (for example `false`, `0`, `no`) and treats blank `secure=` as disabled.
  * Corrected how TLS/`secure` and URL query settings interact to choose the right default port when the URL omits an explicit port.
  * Improved handling of blank query parameters: `statement_timeout=` is ignored, and `connect_timeout=` is not passed through as an empty value.
* **Tests**
  * Expanded ClickHouse URL/kwargs coverage for `secure` precedence, truthy/falsy parsing, explicit ports, and blank parameter behavior.


goldmedal:
## Review

Correct, well-targeted follow-up to #2412 — I traced every branch (precedence `kwargs > query > scheme`, string coercion, the `+https` + `?secure=false` contradiction case, and the blank-value routing for `?secure=`/`?statement_timeout=`) and it all behaves as the tests assert. No public-API/MDL/schema surface here, no cross-connector blast radius. Two things worth raising:

### 1. (Medium) The new tests never run in CI

`tests/connectors/test_clickhouse.py` has module-level `pytestmark = pytest.mark.clickhouse`. In `.github/workflows/wren-ci.yml`:

- the unit job runs only `tests/unit/` (excludes `tests/connectors/`);
- the `test-connector` matrix includes only `postgres` and `mysql` — there is no `clickhouse` row.

So all 21 new `TestClickHouseUrlKwargs` cases are green locally but not exercised by any CI job — exactly the silent-regression risk this fix guards against. These new cases are pure kwargs-assembly (`_FakeChUrl`, no container), so the cheapest fix is to move the URL/kwargs tests into `tests/unit/` (they'd run in the unit job with no Docker), or add a `clickhouse` row to the connector matrix.

### 2. (Minor) `keep_blank_values=True` widened the surface slightly

Blank params are now retained for **all** query keys, not just `secure`/`statement_timeout`, and unhandled ones flow to `clickhouse_connect.get_client(**kwargs)` via the trailing `out.update(kwargs)` (e.g. `?foo=` → `foo=""`). Previously blank params were dropped. Low real-world risk since the intended keys are popped first — just noting it goes a bit beyond the two params the comments call out.

### Positives

- Both commits use conventional `fix(clickhouse):` format.
- `_parse_secure_flag` handles bools, ints, and the truthy/falsy string forms with case/whitespace normalization — parametrized.
- Tests cover the full #2416 acceptance matrix: query-param path, kwargs path, kwargs-over-query precedence, explicit-port-wins, and blank-value handling.

Nice fix — main ask is closing the CI-execution gap so this can't silently regress.

goldmedal (approved):
Thanks @RaidLZ look great 👍

RaidLZ:
Thanks @goldmedal for the quick review and merge

PR #2430, opened 2026-07-03
fix(datafusion): strip trailing semicolon before subquery-wrapping in query

Bartok9:
## Summary

- The DataFusion connector's `query()` wraps user SQL as `SELECT * FROM ({sql}) AS _q LIMIT N` when a limit is supplied.
- A trailing semicolon in the user SQL produced `SELECT * FROM (SELECT 1;) AS _q LIMIT 5`, which DataFusion's parser rejects.
- Strip the terminating run of semicolons/whitespace before wrapping, matching the postgres/redshift/duckdb connectors.

## Motivation

This is the same subquery-wrapping semicolon bug already fixed for postgres (#2407) and hardened for duckdb (#2411), now present in the DataFusion connector. When a user's SQL ends in `;` and a `limit` is passed, wrapping it in a subquery yields `... (SELECT 1;) AS _q ...`; DataFusion raises `sql parser error: Expected: an expression, found: ;`.

The fix reuses the established pattern: a `_TRAILING_SEMICOLONS_RE = re.compile(r"[;\s]+\Z")` helper that strips only the *terminating* run of semicolons/whitespace, so semicolons inside string literals (`SELECT 'a;b'`) are preserved. Only `query()` is affected — `dry_run()` delegates to `ctx.dry_run(sql)` and is unchanged.

## Verification

- `python -m pytest tests/unit/test_datafusion_semicolon.py -v` — 4 passed.
- Real behavior proof of the wrapped SQL (before vs. after the strip):

```
BEFORE fix: SELECT * FROM (SELECT 1;) AS _q LIMIT 5   <- contains ";)": True
AFTER  fix: SELECT * FROM (SELECT 1) AS _q LIMIT 5    <- contains ";)": False
```

- Did NOT change: the no-limit path (SQL passed through verbatim) or `dry_run`.

Apache-2.0 path (`core/**`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQL handling for queries with trailing semicolons when a row limit is applied.
  * Prevented subquery parsing errors by removing only the final trailing semicolon/whitespace before wrapping the query.
  * Preserved semicolons inside string literals and kept unchanged queries without limits behaving as before.

* **Tests**
  * Added coverage for trailing-semicolon behavior with and without limits, including string-literal edge cases.

### DIFF
--- a/core/wren/src/wren/connector/datafusion.py
+++ b/core/wren/src/wren/connector/datafusion.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import io
+import re
 from pathlib import Path
 
 import pyarrow as pa
@@ -10,6 +11,22 @@ from loguru import logger
 from wren.connector.base import ConnectorABC
 from wren.model import DataFusionConnectionInfo
 from wren.model.error import ErrorCode, WrenError
+
+_TRAILING_SEMICOLONS_RE = re.compile(r"[;\s]+\Z")
+
+
+def _strip_trailing_semicolon(sql: str) -> str:
+    """Strip the terminating run of ``;`` characters and surrounding whitespace.
+
+    Wrapping user SQL as ``SELECT * FROM ({sql}) AS _q LIMIT N`` breaks when
+    ``sql`` ends in a semicolon — DataFusion rejects ``SELECT 1;`` inside a
+    subquery (``sql parser error: Expected: an expression, found: ;``). Only
+    the *terminating* run of semicolons/whitespace is stripped, so semicolons
+    inside string literals (e.g. ``SELECT 'a;b' FROM t``) are preserved.
+    Mirrors the postgres/redshift/duckdb connectors, which already strip
+    before subquery-wrapping.
+    """
+    return _TRAILING_SEMICOLONS_RE.sub("", sql)
 
 
 class DataFusionConnector(ConnectorABC):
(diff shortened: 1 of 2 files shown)


goldmedal:
The `_strip_trailing_semicolon` helper (regex `[;\s]+\Z`) is now duplicated **verbatim across 8 connectors** on `main` — `athena`, `canner`, `clickhouse`, `databricks`, `duckdb`, `oracle`, `postgres`, `redshift` — and this PR (`datafusion`) adds a 9th. On top of that, `trino` and `mysql` carry their own slightly-divergent variants (`rstrip().removesuffix(";").rstrip()` / `rstrip(';')`) that handle multiple trailing semicolons differently.

Could we extract this into a single shared helper — a module-level `strip_trailing_semicolon()` in `connector/base.py` (every connector already imports `ConnectorABC` from there, so there's no circular-import concern) — and have each connector call it instead of redefining the regex? The regex version is the most robust, so `trino`/`mysql` can migrate onto it too and the multiple-trailing-semicolon behavior gets unified.

Best done as a dedicated follow-up refactor PR rather than expanding this targeted bugfix — this one is fine to merge as-is.

Bartok9:
Thanks @goldmedal — agreed on all points, and thanks for the clear scope call. Happy to take the follow-up.

Plan for the dedicated refactor PR:
- Add a module-level `strip_trailing_semicolon()` in `connector/base.py` using the robust `_TRAILING_SEMICOLONS_RE = re.compile(r"[;\s]+\Z")` version.
- Migrate all 9 regex-based connectors (athena, canner, clickhouse, databricks, datafusion, duckdb, oracle, postgres, redshift) to call it, dropping their local copies.
- Migrate `trino` (`rstrip().removesuffix(";").rstrip()`) and `mysql` (`rstrip(';')`) onto the shared helper too, which unifies the multiple-trailing-semicolon behavior onto the regex semantics.
- Add a focused unit test for the shared helper covering: single/multiple trailing semicolons, trailing whitespace, and preservation of semicolons inside string literals.

I'll open it as a separate PR once this one lands so the bugfix stays clean and targeted.

goldmedal (approved):

goldmedal:
File https://github.com/Canner/WrenAI/issues/2441 for tracing

PR #2433, opened 2026-07-04
fix(type-mapping): fall back on sqlglot TokenError, not just ParseError

Bartok9:
## Summary

- `parse_type` / `translate_type` in `core/wren/src/wren/type_mapping.py` now fall back to the raw string on **any** sqlglot failure, not just `ParseError`.
- Fixes an uncaught `TokenError` crash on malformed type strings (unterminated quotes, stray control chars).

## Motivation

Both functions document: *"Falls back to original string if parsing fails."* They only caught `(sqlglot.errors.ParseError, ValueError)`. But sqlglot's tokenizer raises `sqlglot.errors.TokenError`, which is a `SqlglotError` and **not** a subclass of `ParseError`. So a type string like `"unterminated` (or one containing a stray `\x00`) escaped the fallback and propagated up, crashing callers (e.g. `parse_types` / `translate_types` over a batch of introspected columns).

The fix catches `sqlglot.errors.SqlglotError`, the common parent of both `ParseError` and `TokenError`, matching the documented fallback contract.

## Verification

```
>           raise TokenError(f"Error tokenizing '{context}'") from e
E           sqlglot.errors.TokenError: Error tokenizing '"unterminate'
...
FAILED tests/unit/test_type_mapping.py::test_parse_type_falls_back_on_tokenizer_error["unterminated]
FAILED tests/unit/test_type_mapping.py::test_translate_type_falls_back_on_tokenizer_error
2 failed, 1 passed
```

With the fix applied:

```
python3 -m pytest tests/unit/test_type_mapping.py -q
48 passed in 2.77s
```

- Added 3 regression cases (2 for `parse_type`, 1 for `translate_type`) that fail without the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of invalid SQL type inputs so type parsing and translation now safely fall back to the original value instead of failing.
  * Fixed edge cases involving malformed text such as unterminated quotes or stray control characters.
  * Added tests to cover the new fallback behavior and ensure these inputs are returned unchanged.

### DIFF
--- a/core/wren/src/wren/type_mapping.py
+++ b/core/wren/src/wren/type_mapping.py
@@ -40,7 +40,10 @@ def parse_type(type_str: str, dialect: str) -> str:
         return type_str
     try:
         return sqlglot.parse_one(type_str, into=DataType, dialect=dialect).sql()
-    except (sqlglot.errors.ParseError, ValueError):
+    except (sqlglot.errors.SqlglotError, ValueError):
+        # SqlglotError covers both ParseError and TokenError (the tokenizer
+        # raises TokenError on unterminated quotes / stray control chars, and
+        # it is NOT a subclass of ParseError). Fall back to the raw string.
         return type_str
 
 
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
Thanks @Bartok9, make sense to me 👍

PR #2434, opened 2026-07-04
fix(oracle): URL-decode credentials from connection_url

Bartok9:
## Summary

- The Oracle connector's `connection_url` path now `unquote()`s the username, password, and service name parsed out of the URL.
- Fixes auth failures when a credential contains a reserved character that must be percent-encoded in the URL.

## Motivation

`_make_oracle_connection` (in `core/wren/src/wren/connector/oracle.py`) parsed a `oracle://user:pass@host:port/svc` URL and passed `parsed.username` / `parsed.password` / `parsed.path` straight to `oracledb.connect()`.

`urllib.parse.urlparse` leaves percent-encoded characters in the userinfo and path. Credentials routinely contain reserved characters (`@`, `/`, `:`, `?`) that **must** be percent-encoded for the URL to parse at all — so a password like `p@ss/word` arrives as `oracle://user:p%40ss%2Fword@...`, and `oracledb` receives the literal `p%40ss%2Fword`, causing an authentication failure.

The mssql and mysql connectors already `unquote()` these components; this brings Oracle in line with them. (clickhouse uses `unquote_plus`, which differs on a literal `+`.) `unquote` (not `unquote_plus`) is used here so a literal `+` in a credential is preserved — `+` only means space in query strings, not userinfo.

## Verification

```
E         - us@er
E         + us%40er
FAILED tests/unit/test_oracle_url_decode.py::test_connection_url_decodes_percent_encoded_credentials
1 failed, 1 passed
```

With the fix applied:

```
python3 -m pytest tests/unit/test_oracle_url_decode.py -q
2 passed in 0.19s
```

- Two regression tests (mocked `oracledb.connect`, no live DB): one asserts percent-decoding, one asserts a literal `+` is preserved.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Oracle connection URL handling by percent-decoding usernames, passwords, and service names before connecting.
  * Preserved literal `+` characters in Oracle credentials to avoid unintended character conversion.
* **Tests**
  * Added unit tests covering percent-decoded credential parsing and ensuring `+` remains literal.
  * Added coverage that the Oracle connection logic can be exercised without requiring a live Oracle installation.
* **Chores**
  * Updated CI dependency installation to use `uv sync --locked --no-install-project` for the unit test job.

### DIFF
--- a/.github/workflows/wren-ci.yml
+++ b/.github/workflows/wren-ci.yml
@@ -62,11 +62,8 @@ jobs:
           uv sync --locked --no-install-project
           uv run --no-sync maturin build
       - name: Install dependencies
-        # Install the `oracle` extra so tests/unit/test_oracle_url_decode.py
-        # (guarded by pytest.importorskip("oracledb")) actually runs here
-        # instead of silently skipping.
         run: |
-          uv sync --locked --extra oracle
+          uv sync --locked --no-install-project
           uv pip install --reinstall --no-index --find-links ../wren-core-py/target/wheels/ wren-core-py
       - name: Run unit tests
         # Run the whole tests/unit/ tree rather than filtering on `-m unit`:
--- a/.github/workflows/wren-ci.yml
+++ b/.github/workflows/wren-ci.yml
@@ -63,7 +63,7 @@ jobs:
           uv run --no-sync maturin build
       - name: Install dependencies
         run: |
-          uv sync --locked --no-install-project
+          uv sync --locked
           uv pip install --reinstall --no-index --find-links ../wren-core-py/target/wheels/ wren-core-py
       - name: Run unit tests
         # Run the whole tests/unit/ tree rather than filtering on `-m unit`:
--- a/.github/workflows/wren-ci.yml
+++ b/.github/workflows/wren-ci.yml
@@ -62,8 +62,11 @@ jobs:
           uv sync --locked --no-install-project
           uv run --no-sync maturin build
       - name: Install dependencies
+        # Install the `oracle` extra so tests/unit/test_oracle_url_decode.py
+        # (guarded by pytest.importorskip("oracledb")) actually runs here
+        # instead of silently skipping.
         run: |
-          uv sync --locked
+          uv sync --locked --extra oracle
           uv pip install --reinstall --no-index --find-links ../wren-core-py/target/wheels/ wren-core-py
       - name: Run unit tests
         # Run the whole tests/unit/ tree rather than filtering on `-m unit`:
--- a/core/wren/src/wren/connector/oracle.py
+++ b/core/wren/src/wren/connector/oracle.py
@@ -3,8 +3,16 @@
 from decimal import Decimal as PyDecimal
 from urllib.parse import unquote, urlparse
 
-import oracledb
 import pyarrow as pa
+
+# Lazy driver import: keep the module importable without the ``oracle`` extra so
+# URL-decode/helper tests run under plain unit deps (mirrors the mysql/mssql/
+# clickhouse connectors). ``oracledb`` is only touched when a connection is
+# actually made.
+try:
+    import oracledb
+except ImportError:  # pragma: no cover
+    oracledb = None
 
 from wren.connector.base import ConnectorABC
 from wren.model.error import DIALECT_SQL, ErrorCode, ErrorPhase, WrenError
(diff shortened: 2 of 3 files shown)


goldmedal:
The fix itself looks correct, but the two new tests won't actually run in CI, so the regression coverage this PR adds never executes:

- `test_oracle_url_decode.py` starts with `pytest.importorskip("oracledb")`, and the module under test does a top-level `import oracledb` — so the tests only run when the **`oracle`** extra is installed.
- `oracle = ["oracledb>=2"]` is an optional extra in `core/wren/pyproject.toml`, not a base dependency.
- The `test-unit` job in `wren-ci.yml` runs `uv sync --locked` with **no extras**, so `oracledb` is absent and both tests silently **skip**.
- The `test-connector` matrix only covers `postgres` and `mysql` — there's no job that installs the `oracle` extra.

Net effect: the tests pass locally (where the extra happens to be installed) but are skipped on CI. Suggest one of:
- install the `oracle` extra in the `test-unit` job (e.g. `uv sync --locked --extra oracle`), or
- add an `oracle` row to the `test-connector` matrix.

Otherwise the guard resolves to "skip" everywhere in CI and a future regression here would go unnoticed.

Bartok9:
Good catch, thanks @goldmedal — you're right that the guard was resolving to skip everywhere in CI.

Pushed 7467724 which installs the `oracle` extra in the `test-unit` job (`uv sync --locked --extra oracle`), your first suggested option. That keeps the new regression tests in the existing unit lane rather than spinning up a full connector matrix row (no live Oracle DB needed since `oracledb.connect` is mocked). Both tests now execute instead of skipping.

I went with the `test-unit` approach over a `test-connector` matrix row because these are pure URL-parsing unit tests with a mocked connect — a connector-matrix row would imply a live datasource that isn't actually exercised here. Happy to switch to a matrix row instead if you'd prefer to keep all Oracle coverage under `test-connector`.

goldmedal:
Nice fix — matches the mssql/mysql `unquote` handling and the tests cover both the percent-decode and the literal-`+` case.

One minor doc nit (non-blocking): the code comment and PR body say this "mirrors the mssql/mysql/clickhouse connectors," but `clickhouse.py` actually uses `unquote_plus`, not `unquote` — so it differs in `+` handling (clickhouse would turn a literal `+` into a space in userinfo). The choice of `unquote` here correctly follows mssql/mysql, so I'd just drop "clickhouse" from the "mirrors" list (or note that it differs) to avoid implying uniformity that isn't there. Whether clickhouse's `unquote_plus` is itself a latent bug is out of scope for this PR.

Bartok9:
Good catch @goldmedal — you're right, clickhouse uses `unquote_plus` so it differs on the literal-`+` case. Pushed a doc-only commit dropping clickhouse from the "mirrors" list in both the code comment and the PR body (now reads mssql/mysql, with a note that clickhouse differs). No behavior change.

goldmedal:
One optional follow-up thought while you're in here — totally your call whether to fold it into this PR or leave it out of scope.

`oracle.py` does a top-level `import oracledb`, which is why these tests need `pytest.importorskip("oracledb")` and why the unit job now has to install the `oracle` extra to run them. The `mysql`/`mssql`/`clickhouse` connectors instead lazy-import their drivers inside the connection functions, so their connector modules import fine without the driver and their URL-decode/helper tests run under plain unit deps.

If you moved `import oracledb` into `_make_oracle_connection` (and the other functions that use it) to match that pattern:
- importing `wren.connector.oracle` would no longer require the driver,
- the URL-decode tests could run via mocking without the `oracle` extra (the CI `--extra oracle` change could then be dropped),
- runtime behavior is unchanged — the driver is only touched when a connection is actually made.

Would you be up for including that in this PR, or would you rather keep this one focused on the decode fix and do the lazy-import as a separate change? Either is fine by me.

Bartok9:
Done — folded it into this PR (d544123).

- Moved `import oracledb` to a `try/except ImportError` lazy import, matching the mysql/mssql/clickhouse pattern, so `wren.connector.oracle` imports fine without the driver.
- The URL-decode tests now inject a stand-in module (no `pytest.importorskip`) and run under plain unit deps.
- Reverted the `uv sync --locked --extra oracle` change to `test-unit` — no longer needed.

Verified locally: both tests execute (not skip) with `oracledb` blocked from import, and `ruff check` is clean. Runtime is unchanged — the driver is only touched when a connection is actually made.

goldmedal (approved):
Thanks @Bartok9, look great 👍

PR #2435, opened 2026-07-04
fix(trino): URL-decode credentials and identifiers from connection URL

Bartok9:
## Summary

- `_parse_trino_url` now `unquote()`s the username, password, catalog, and schema parsed out of the connection URL.
- Fixes auth/connection failures when a credential or identifier contains a reserved character that must be percent-encoded in the URL.

## Motivation

`_parse_trino_url` (in `core/wren/src/wren/connector/trino.py`) parsed a `trino://user:pass@host:port/catalog/schema` URL and put `parsed.username` / `parsed.password` / the catalog+schema path parts straight into the connect kwargs.

`urllib.parse.urlparse` leaves percent-encoded characters in the userinfo and path. Credentials and identifiers can contain reserved characters (`@`, `/`, `:`, `?`) that **must** be percent-encoded for the URL to parse — so a password like `p@ss/word` (`...:p%40ss%2Fword@...`) reached Trino as the literal `p%40ss%2Fword`, and a catalog named `cat/alog` (`cat%2Falog`) was mis-parsed.

The mssql, mysql, clickhouse, and oracle connectors already `unquote()` these components. This brings Trino in line. `unquote` (not `unquote_plus`) is used so a literal `+` in a credential is preserved.

## Verification

```
E         - us@er
E         + us%40er
FAILED tests/connectors/test_trino.py::test_parse_trino_url_decodes_percent_encoded_credentials
1 failed, 4 passed
```

With the fix applied:

```
python3 -m pytest tests/connectors/test_trino.py -k parse_trino_url -q
5 passed, 66 deselected in 0.46s
```

- Two new regression tests (pure parser, no Trino container): one asserts percent-decoding of user/password/catalog/schema, one asserts a literal `+` is preserved.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Trino connection URL handling by percent-decoding credential and identifier components (including user, password, catalog/schema, and path).
  * Preserved literal `+` characters in Trino userinfo.
  * Kept existing URL validation and connection defaults unchanged.
* **Tests**
  * Added new unit-test coverage for Trino parsing helpers (data types, column building, trailing-semicolon stripping, and URL parsing/validation).
  * Added checks for install-hint errors when the Trino dependency is missing and ensured native connector import doesn’t load `ibis`.

### DIFF
--- a/core/wren/src/wren/connector/trino.py
+++ b/core/wren/src/wren/connector/trino.py
@@ -11,7 +11,7 @@ import contextlib
 import datetime as dtlib
 import json
 from decimal import Decimal as PyDecimal
-from urllib.parse import parse_qsl, urlparse
+from urllib.parse import parse_qsl, unquote, urlparse
 
 import pyarrow as pa
 import sqlglot
@@ -322,14 +322,17 @@ def _parse_trino_url(url: str, extra_kwargs: dict | None) -> dict:
     if extra_kwargs:
         query_kwargs.update(extra_kwargs)
 
+    # urlparse leaves percent-encoded characters in userinfo/path, so decode
+    # them before they reach Trino. Use unquote (not unquote_plus) so a
+    # literal ``+`` in a credential is preserved.
     out: dict = {
         "host": parsed.hostname,
         "port": int(parsed.port or 8080),
-        "user": parsed.username,
-        "catalog": catalog,
-        "schema": schema,
+        "user": unquote(parsed.username),
+        "catalog": unquote(catalog) if catalog else None,
+        "schema": unquote(schema) if schema else None,
         "timezone": "UTC",
-        "_password": parsed.password,
+        "_password": unquote(parsed.password) if parsed.password else None,
     }
     if parsed.scheme == "trino+https":
         out["http_scheme"] = "https"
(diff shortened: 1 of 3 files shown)


goldmedal:
The fix looks correct, but heads-up: **the new regression tests don't run in CI.**

`wren-ci.yml` only executes `tests/unit/` (the `test-unit` job) and a connector matrix limited to postgres + mysql (`test-connector`). `tests/connectors/test_trino.py` is in neither, so no CI job runs these new tests — meaning a future re-break of the URL-decoding won't be caught automatically.

This is a pre-existing gap (the whole `test_trino.py` file, parser + integration, is un-run in CI), not something this PR introduced. But it does undercut the value of adding the regression tests. The parser tests are pure (no container), yet they can't simply move to `tests/unit/` because the file's top-level `from testcontainers.trino import TrinoContainer` forces the extra even to collect them.

Cheapest durable fix: split the pure `_parse_trino_url` tests into a `tests/unit/` module (no testcontainer import) so the `test-unit` job runs them. Not a blocker for this PR.

Bartok9:
Good catch, thanks — done. I split the pure `_parse_trino_url` / `_parse_trino_data_type` / `_build_trino_column` / `_strip_trailing_semicolon` / lazy-import tests into a new `tests/unit/test_trino_parser.py` (no `testcontainers.trino` import), so the `test-unit` job now collects and runs the URL-decode regression coverage. The container-backed integration suite stays in `tests/connectors/test_trino.py`. Verified locally: 56 passed. Pushed in 93e86615.

goldmedal (approved):
👍

PR #2436, opened 2026-07-04
fix(clickhouse): use unquote (not unquote_plus) for URL credentials; decode database

Bartok9:
## Summary

- The ClickHouse `connection_url` path now decodes userinfo with `unquote` instead of `unquote_plus`, and also `unquote`s the database name parsed from the path.
- Fixes silent credential corruption when a password contains a literal `+`, and a missing-decode bug for percent-encoded database names.

## Motivation

In `_build_clickhouse_client_kwargs` (`core/wren/src/wren/connector/clickhouse.py`), the URL username/password were decoded with `unquote_plus`. But `+` only means "space" inside a URL **query string** — never in the userinfo or path (RFC 3986). So a password like `pw+1` (written literally in the URL) was silently turned into `pw 1`, and authentication failed with no clear cause.

Every other connector in this package (mssql, mysql, oracle, trino) uses plain `unquote` for exactly this reason. This brings ClickHouse in line.

Separately, the `database` name parsed from the URL path was passed through **raw** (never decoded), so a percent-encoded database name (e.g. `my%20analytics`) reached ClickHouse verbatim.

`unquote` still decodes `%20`→space, `%40`→`@`, etc., so the existing decode test continues to pass unchanged.

## Verification

```
E         - my analytics
E         + my%20analytics
FAILED tests/unit/test_clickhouse_helpers.py::test_clickhouse_url_preserves_literal_plus_in_credentials
FAILED tests/unit/test_clickhouse_helpers.py::test_clickhouse_url_decodes_percent_encoded_database
2 failed, 2 passed
```

With the fix applied:

```
python3 -m pytest tests/unit/test_clickhouse_helpers.py -q
25 passed in 0.77s
```

- Two new regression tests (literal `+` preserved; database path decoded); the pre-existing `%40`/`%20` decode test still passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ClickHouse connection URL parsing so username, password, and database values are decoded correctly.
  * Preserved literal `+` characters in credential userinfo instead of converting them to spaces.
  * Properly decoded percent-encoded database names from the URL path.
* **Tests**
  * Added regression coverage to verify correct `+` handling in credentials and percent-decoding of database names.

### DIFF
--- a/core/wren/src/wren/connector/clickhouse.py
+++ b/core/wren/src/wren/connector/clickhouse.py
@@ -273,9 +273,7 @@ def _build_clickhouse_client_kwargs(connection_info: Any) -> dict:
         out: dict = {
             "host": parsed.hostname,
             "port": int(parsed.port) if parsed.port else default_port,
-            "username": (
-                unquote(parsed.username) if parsed.username else "default"
-            ),
+            "username": (unquote(parsed.username) if parsed.username else "default"),
             "password": unquote(parsed.password) if parsed.password else "",
             "settings": settings,
         }
--- a/core/wren/src/wren/connector/clickhouse.py
+++ b/core/wren/src/wren/connector/clickhouse.py
@@ -13,7 +13,7 @@ import json
 import re
 from decimal import Decimal as PyDecimal
 from typing import Any
-from urllib.parse import parse_qsl, unquote_plus, urlparse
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
Thanks @Bartok9, nice catch 👍

PR #2444, opened 2026-07-06
fix(sdk): URL-decode ClickHouse connection-URL credentials and database

Bartok9:
## Summary

- `DataSource.clickhouse` now URL-decodes the **username**, **password**, and **database** parsed from a `connectionUrl`.
- The password is decoded with `unquote` instead of `unquote_plus`, so a literal `+` in a password is preserved instead of being turned into a space.

## Motivation

`_handle_clickhouse_url` in `core/wren/src/wren/model/data_source.py`:

- decoded the password with `unquote_plus`, which maps `+` → space. A ClickHouse password containing a `+` therefore silently connected with the wrong credential.
- did **not** decode the username or database at all, so percent-escapes (e.g. `%40` for `@`, `%20` for a space) survived as literal text and produced auth/lookup failures.

This brings the SDK path in line with the connector-level handling already used for `mssql` and `mysql`, which use `unquote`.

## Verification

```
$ python -m pytest tests/unit/test_clickhouse_url.py -q
3 passed

# on current main (git stash of the fix):
FAILED test_plus_in_password_is_preserved     (pa ss != pa+ss)
FAILED test_username_and_database_are_decoded  (us%40er != us@er)
2 failed, 1 passed
```

The new regression tests fail on `main` and pass with the fix. `ruff format` and `ruff check` clean. Change is confined to Apache-2.0 `core/wren/**`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ClickHouse connection URL handling so percent-encoded usernames, passwords, and database names are decoded correctly.
  * Preserved literal `+` characters in credentials instead of converting them to spaces.

* **Tests**
  * Added coverage for ClickHouse connection parsing to verify decoding behavior and prevent regressions.

### DIFF
--- a/core/wren/src/wren/model/data_source.py
+++ b/core/wren/src/wren/model/data_source.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import urllib
 from enum import StrEnum, auto
 from typing import Any
-from urllib.parse import unquote_plus
+from urllib.parse import unquote
 
 from wren.model import (
     AthenaConnectionInfo,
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
Thanks @Bartok9 👍

PR #2448, opened 2026-07-08
fix(oracle): sanitize brackets in connection-URL userinfo before parse

Bartok9:
## Summary

- `oracle://` connection URLs containing `[` or `]` in the userinfo (username/password) crashed with `ValueError` from `urlparse`'s IPv6 netloc check.
- Bracketed credentials now parse, while legitimate bracketed IPv6 hosts still work.

## Motivation

`_make_oracle_connection` called `urlparse(url)` directly. Python's `urlparse` treats `[`/`]` anywhere in the netloc as IPv6 host delimiters, so a URL like `oracle://user:p[a]ss@host:1521/svc` raises `ValueError: 'host' does not appear to be an IPv4 or IPv6 address` before any connection is attempted — the credential can never be used.

The MySQL connector already solved this exact problem with `_parse_mysql_connection_url`, which escapes brackets in the userinfo segment only. This change mirrors that helper for Oracle (`_parse_oracle_connection_url`), keeping valid IPv6 hosts (`oracle://user:pass@[::1]:1521/svc`) intact.

## Verification

```
$ uv run pytest tests/unit/test_oracle_url_decode.py -q
4 passed in 0.20s
```

Regression test proof (fails without the fix):
```
$ git stash push -- src/wren/connector/oracle.py
$ uv run pytest tests/unit/test_oracle_url_decode.py -k "brackets or ipv6" -q
FAILED ...test_connection_url_with_brackets_in_credentials - ValueError: 'host' does not appear to be an IPv4 or IPv6 address
1 failed, 1 passed
```
The `ipv6` test passes both with and without the fix, confirming no regression for legitimate bracketed IPv6 hosts.

Apache-2.0 path (`core/**`). One commit, two files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Oracle connection URL handling so credentials containing `[` or `]` no longer break parsing.
  * Preserved correct handling of IPv6 host addresses while decoding connection details.
* **Tests**
  * Added coverage for bracketed credentials in the connection URL.
  * Added coverage to ensure IPv6 hosts remain unchanged during parsing.

### DIFF
--- a/core/wren/src/wren/connector/oracle.py
+++ b/core/wren/src/wren/connector/oracle.py
@@ -21,6 +21,38 @@ from wren.model.error import DIALECT_SQL, ErrorCode, ErrorPhase, WrenError
 _TRAILING_SEMICOLONS_RE = re.compile(r"[;\s]+\Z")
 
 
+def _parse_oracle_connection_url(url: str):
+    """Parse an Oracle URL after escaping raw brackets in userinfo only.
+
+    ``urllib.parse.urlparse`` treats ``[`` / ``]`` anywhere in the netloc as
+    IPv6 host delimiters. That is correct for hosts like ``oracle://[::1]/db``
+    but it raises ``ValueError`` on raw credentials such as
+    ``oracle://user:p[a]ss@host/svc``. Sanitise only the userinfo segment so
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
Thanks @Bartok9

PR #2449, opened 2026-07-08
fix(core): prune inaccessible CLS columns in wildcard expansion instead of denying

fmcmac:
## Problem

Column-level access control (CLAC/CLS) denies **any** query that references a model with a **table alias**, even when the protected column is never selected. `SELECT * FROM model` correctly prunes an inaccessible CLS column, but `SELECT alias.other_col FROM model alias` (or a join using an alias) fails with `Access denied to column … violates access control rule …`. `SELECT count(*) FROM model` on a model with a CLS column is denied for the same reason.

## Root cause

For `FROM customer c`, the plan is `Projection → SubqueryAlias(c) → TableScan(customer)`. `ModelAnalyzeRule` collects required columns keyed by the alias `c`, but the bottom-up transform processes the inner bare `TableScan(customer)` first with no alias, looks required columns up under the model name, and finds none. That builds a throwaway inner `ModelPlanNode` with **empty** required fields, which triggers the wildcard expansion in `RelationChain::source` (`ModelSourceNode::new`). That loop expanded every physical column and called `get_remote_column_exp` per column, which hard-denies via `validate_clac_rule`. `analyze_subquery_alias_model` later rebuilds the node correctly with the alias's required columns, but the throwaway build errors first.

`SELECT *` avoids this because it doesn't route through the empty-required wildcard path.

## Fix

In the wildcard-expansion loop, prune columns the caller cannot access (`validate_clac_rule` + `continue`) instead of denying — matching `SELECT *` semantics. This path is an *implicit* all-columns expansion (`count(*)`, or the aliased throwaway plan); a protected column the query never explicitly selected should be dropped, not turned into a hard error. Explicit references still deny, because they arrive as named required fields through `ModelPlanNodeBuilder::build` (the non-wildcard branch), not this path. Pruning a column can only ever remove it, so it cannot widen access.

## Tests

Adds `test_clac_unreferenced_column_pruned_not_denied`:
- **Pruned (must succeed):** aliased/unaliased join, aliased/unaliased standalone, aliased/unaliased `count(*)`, model-qualified `alias.*`, and bare `*` — none reference the protected columns.
- **Still denied (must error with "violates access control rule"):** explicit reference to a protected column in SELECT (aliased and unaliased), a second protected column, and references in WHERE, GROUP BY, and JOIN ON.

Verified locally against `main`: the new test passes and the fix is a minimal, localized change in `ModelSourceNode::new`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Wildcard queries now omit columns the caller is not allowed to access, instead of failing with a permission error.
  * Protected columns that are not explicitly referenced are now safely pruned, including in aliased queries and `count(*)`/`*` scenarios.
  * Explicitly referenced protected columns continue to return access-denied errors as expected.

### DIFF
--- a/core/wren-core/core/src/logical_plan/analyze/plan.rs
+++ b/core/wren-core/core/src/logical_plan/analyze/plan.rs
@@ -1046,6 +1046,25 @@ impl ModelSourceNode {
                     if column.is_calculated {
                         continue;
                     }
+                    // Prune columns the caller cannot access under column-level
+                    // security instead of denying. This is an *implicit* all-columns
+                    // expansion (a `count(*)`, or the throwaway inner plan built for
+                    // a model scan referenced with a table alias, whose required
+                    // columns are keyed by the alias so the model is wildcard-
+                    // expanded). A protected column the query never explicitly
+                    // selected must be dropped here — matching `SELECT *` semantics —
+                    // not turned into a hard permission error. Explicit references
+                    // still deny, because they arrive as named required fields (the
+                    // non-wildcard branch / ModelPlanNodeBuilder::build), not here.
+                    let (is_valid, _) = validate_clac_rule(
+                        model.name(),
+                        &column,
+                        &session_properties,
+                        Some(Arc::clone(&analyzed_wren_mdl)),
+                    )?;
+                    if !is_valid {
+                        continue;
+                    }
                     fields_buffer.insert((
                         Some(TableReference::bare(quoted(model.name()))),
                         Arc::new(Field::new(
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
Thanks @fmcmac, makes sense to me.

One note for future readers (no change requested): the aliased-scan case here routes through a `build → discard → rebuild` — under the bottom-up transform the bare `TableScan` is visited before its `SubqueryAlias`, looked up by model name instead of the alias, so required columns miss → empty → wildcard expansion → a `ModelPlanNode` that's then discarded and rebuilt by `analyze_subquery_alias_model`.

That's a separate, behavior-neutral inefficiency (and fixing it wouldn't remove the need for this PR — genuine `count(*)` still needs the prune here). Opened #2451 to track it. Not a blocker.

PR #2450, opened 2026-07-08
fix(core): emit all relationship JOINs for multi-relationship calculated columns

fmcmac:
## Summary

A wren-core model with calculated columns spanning **two or more distinct relationships** fails to plan with `No field named __relation__1.<col>` — only the first relationship's JOIN is emitted. This surfaces whenever a narrow fact model joins to multiple dimensions and projects an attribute from each (e.g. a DC-inventory fact with `market`/`retailer` via a market-retailer relationship and `item_num` via a product relationship).

## Root cause

Two interacting bugs:

1. `ModelPlanNodeBuilder::merge_graph` (`plan.rs`) blindly added every node from each per-calc-col sub-graph into the directed graph without deduping. With calc cols spanning two relationships, the shared base model ended up as multiple disconnected nodes in the merged graph.
2. `RelationChain::with_chain` (`relation_chain.rs`) walked the merged graph as a linear chain via `find_edge(start, next)`, updating `start = next` each iteration. Combined with the duplicate base-model nodes, the walk hit a duplicate node as `next`, `find_edge` returned `None`, and the join chain was silently truncated after the first relationship.

## Fix

- `merge_graph` dedupes nodes by `Dataset` value (`Dataset` already derives `Eq + Hash`) and skips parallel edges between the same pair.
- The fallback `add_node` for the base model in `build()` is now conditional on the model not already being present.
- `with_chain` tracks `prev` separately from the original `start` and prefers `find_edge(prev, next)`, falling back to `find_edge(start, next)` for star fan-out from the base model. Linear-chain behaviour is preserved (`prev == start` on the first step); star fan-out (A → B and A → C in one model) now works.

## Test

Adds `test_multiple_relationship_calculated_columns`: a base model with two relationships and three calculated columns (two through one relationship, one through the other), asserting `SELECT *` transforms successfully and projects all three. Without the fix the test fails with `FieldNotFound { __relation__1.item_num }`.

The test runs on a dedicated 8 MiB-stack thread — the analyzer's recursion can exceed the 2 MiB default test-thread stack at several chained joins; production tokio worker threads have a larger stack and are unaffected (this is not infinite recursion).

## Context

This is one of a small set of patches ZURU has carried on a wren-core fork; we are upstreaming them as we migrate our pin onto the current `Canner/WrenAI` monorepo. Companion to #2449 (CLS wildcard-prune).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed planning for calculated columns when multiple relationships overlap, preventing missing fields and incorrect query truncation.
  * Improved graph merging so existing model nodes and relationships are reused instead of duplicated, reducing redundant joins.
  * Updated relation-chain handling to support both linear and fan-out relationship traversal.

* **Tests**
  * Added regression coverage for models with multiple relationship-based calculated columns.

### DIFF
--- a/core/wren-core/core/src/mdl/mod.rs
+++ b/core/wren-core/core/src/mdl/mod.rs
@@ -2610,6 +2610,171 @@ mod test {
         Ok(())
     }
 
+    /// Regression test for the multi-relationship calculated-column bug.
+    ///
+    /// A model with multiple calculated columns that traverse *different*
+    /// relationships used to fail at planning time because the merged dataset
+    /// graph contained duplicate copies of the base model (one per merged
+    /// sub-graph). `RelationChain::with_chain` located the next link via
+    /// `directed_graph.find_edge(start, next)`; when the second sub-graph
+    /// started with a duplicate base-model node, no edge was found from the
+    /// first relationship's terminal node and the chain was silently
+    /// truncated. The result was an error like
+    ///
+    ///     No field named __relation__1.<col_from_second_relationship>
+    ///
+    /// because only the first relationship's columns ended up in the
+    /// projection. `merge_graph` now dedupes nodes/edges by `Dataset`, the base
+    /// model is added conditionally, and `with_chain` prefers an edge from the
+    /// previously visited node (falling back to the start node) so star
+    /// fan-outs across multiple relationships plan correctly.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn test_multiple_relationship_calculated_columns() -> Result<()> {
+        // Run the transform on a dedicated thread with an enlarged stack:
+        // wren-core's plan analyzer uses deep recursion and can exceed the
+        // default 2 MiB test-thread stack when joining several relationships.
+        // Production tokio worker threads have a larger stack and are
+        // unaffected; this is not infinite recursion.
+        std::thread::Builder::new()
+            .stack_size(8 * 1024 * 1024)
+            .spawn(|| {
+                let rt = tokio::runtime::Builder::new_current_thread()
+                    .enable_all()
+                    .build()
+                    .unwrap();
+                rt.block_on(test_multiple_relationship_calculated_columns_inner())
+            })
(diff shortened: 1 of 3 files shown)


core/wren-core/core/src/mdl/mod.rs
```diff
@@ -2610,6 +2610,171 @@ mod test {
...
+    /// fan-outs across multiple relationships plan correctly.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn test_multiple_relationship_calculated_columns() -> Result<()> {
+        // Run the transform on a dedicated thread with an enlarged stack:
+        // wren-core's plan analyzer uses deep recursion and can exceed the
+        // default 2 MiB test-thread stack when joining several relationships.
+        // Production tokio worker threads have a larger stack and are
+        // unaffected; this is not infinite recursion.
+        std::thread::Builder::new()
+            .stack_size(8 * 1024 * 1024)
```
goldmedal, line 2639: Minor / non-blocking: consider following the repo test convention here instead of the bespoke thread.

Every other test in `mdl/mod.rs` is a plain `#[tokio::test]`, and the suite is already run under `RUST_MIN_STACK=8388608` (see `wren-core/CLAUDE.md` and the CI test step). So the hand-rolled `std::thread::Builder::stack_size(8 MiB)` + current-thread runtime shouldn't be necessary — a plain `#[tokio::test]` should get the enlarged stack from the env var like the rest of the tests, and drops ~20 lines of boilerplate.

Could you switch `test_multiple_relationship_calculated_columns` to a plain `#[tokio::test]` (inlining `_inner`) to stay consistent? If it turns out the enlarged stack genuinely isn't picked up in that context, keeping the explicit thread is fine — just leave the justification comment as-is.

goldmedal (approved):
Thanks @fmcmac, look great 👍.  Because I'm planning to work on the crate publishing work, I'll merge this PR first. I filed the follow-up issue #2435  for the refactoring.
Nice work.

PR #2454, opened 2026-07-09
fix(clickhouse): sanitize brackets in connection-URL userinfo before parse

Bartok9:
## Summary

- Sanitize raw `[` / `]` in ClickHouse connection-URL userinfo before `urlparse`.
- Preserve legitimate IPv6 hosts that still need brackets around the host.

## Motivation

`urlparse` treats `[` / `]` anywhere in the netloc as IPv6 delimiters. A password like `p[a]ss` therefore raises `ValueError: Invalid IPv6 URL` and the connector never opens. Same root cause already fixed for MySQL / MSSQL / Oracle on this tree — ClickHouse was still on bare `urlparse`.

## Verification

- `PYTHONPATH=src pytest tests/unit/test_clickhouse_helpers.py` → **50 passed**
- Regression tests:
  - `test_clickhouse_url_allows_raw_brackets_in_password`
  - `test_clickhouse_url_preserves_ipv6_host_with_bracketed_password`
- Real behavior without fix: `clickhouse://user:p[a]ss@host:9000/db` → `ValueError: Invalid IPv6 URL`
- With fix: username `user`, password `p[a]ss`, host intact; `clickhouse://user:p[a]ss@[::1]:9000/db` keeps host `::1`.
- Did NOT change decoding of `%`-encoded characters, `unquote_plus` credential handling, secure-port defaults, or query wrapping.

## License

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of ClickHouse connection URLs when credentials include special characters like `[` or `]`, preventing URL parsing failures.
  * Ensured IPv6 host addresses are parsed correctly while continuing to extract the right username and password from the URL.
* **Tests**
  * Added unit test coverage for bracketed credentials and IPv6 host edge cases in ClickHouse URL parsing.

### DIFF
--- a/core/wren/src/wren/connector/clickhouse.py
+++ b/core/wren/src/wren/connector/clickhouse.py
@@ -245,7 +245,6 @@ def _parse_secure_flag(value: Any) -> bool:
     return bool(value)
 
 
-
 def _parse_clickhouse_connection_url(url: str):
     """Parse a ClickHouse URL after escaping raw brackets in userinfo only.
 
--- a/core/wren/src/wren/connector/clickhouse.py
+++ b/core/wren/src/wren/connector/clickhouse.py
@@ -245,6 +245,39 @@ def _parse_secure_flag(value: Any) -> bool:
     return bool(value)
 
 
+
+def _parse_clickhouse_connection_url(url: str):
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
Thanks @Bartok9  👍

PR #2456, opened 2026-07-09
fix(snowflake): push LIMIT into SQL and strip trailing semicolon on wrap

Bartok9:
## Summary

- Push `query(..., limit=N)` into Snowflake SQL via `SELECT * FROM (<sql>) AS _wren_sub LIMIT N`.
- Strip a trailing semicolon before wrapping so `;`-terminated user SQL cannot break composition.
- Lazy-import `snowflake.connector` so unit tests do not require the optional extra.

## Motivation

The previous path ran the full statement and used `arrow_table.slice(0, limit)`. Large queries paid full scan + transfer cost on the client even when Aster only needed a sample/page. Other connectors (postgres/redshift/clickhouse/trino) already apply LIMIT server-side; Snowflake was the outlier. Trailing-semicolon strip matches the established wrap helpers.

## Verification

- `PYTHONPATH=src pytest tests/unit/test_snowflake_limit_pushdown.py` → **3 passed**
- Asserted executed SQL for `SELECT 1 AS x;` + `limit=5` is  
  `SELECT * FROM (SELECT 1 AS x) AS _wren_sub LIMIT 5`
- Without `limit`, original SQL is executed unchanged.

## License

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Snowflake query execution with `limit` by pushing the LIMIT into the SQL run on Snowflake rather than slicing results in Python.
  * More reliable SQL wrapping by stripping only a trailing semicolon/whitespace (while keeping semicolons inside quoted strings and respecting line-comment boundaries).
  * Updated Snowflake diagnostics so reported SQL reflects the actual composed query that was executed (including for `dry_run`).
* **Tests**
  * Added unit coverage for trailing-semicolon cleanup, LIMIT pushdown SQL rewriting, and the no-limit execution path.

### DIFF
--- a/core/wren/src/wren/connector/snowflake.py
+++ b/core/wren/src/wren/connector/snowflake.py
@@ -2,13 +2,12 @@
 
 from __future__ import annotations
 
+import re
+
 import pyarrow as pa
 
 from wren.connector.base import ConnectorABC
 from wren.model.error import DIALECT_SQL, ErrorCode, ErrorPhase, WrenError
-
-
-import re
 
 _TRAILING_SEMICOLONS_RE = re.compile(r"[;\s]+\Z")
 
@@ -21,7 +20,6 @@ def _strip_trailing_semicolon(sql: str) -> str:
     postgres/redshift connectors.
     """
     return _TRAILING_SEMICOLONS_RE.sub("", sql)
-
 
 
 def _build_connection_params(connection_info) -> dict:
--- a/core/wren/src/wren/connector/snowflake.py
+++ b/core/wren/src/wren/connector/snowflake.py
@@ -3,10 +3,25 @@
 from __future__ import annotations
 
 import pyarrow as pa
-import snowflake.connector
 
 from wren.connector.base import ConnectorABC
 from wren.model.error import DIALECT_SQL, ErrorCode, ErrorPhase, WrenError
+
+
+import re
+
+_TRAILING_SEMICOLONS_RE = re.compile(r"[;\s]+\Z")
+
+
+def _strip_trailing_semicolon(sql: str) -> str:
+    """Strip terminating ``;`` / whitespace so we can wrap SQL as a subquery.
+
+    Snowflake rejects ``SELECT 1;`` inside a subquery. Only the trailing run is
+    removed so semicolons inside string literals stay intact. Mirrors
+    postgres/redshift connectors.
+    """
+    return _TRAILING_SEMICOLONS_RE.sub("", sql)
+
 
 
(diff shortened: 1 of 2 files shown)


Bartok9:
Fixed both CodeRabbit findings in 5501e94: the user SQL now sits on its own line inside the wrapper — `SELECT * FROM (\n<sql>\n) AS _wren_sub LIMIT N` — so a trailing `-- comment` in the inner SQL is terminated by the newline and can no longer swallow the closing paren, alias, or LIMIT clause. ORDER BY semantics are preserved (the outer LIMIT wraps the fully-ordered subquery, matching the prior Python-slice behavior). Added `test_query_limit_survives_trailing_line_comment` covering the comment case.

goldmedal:
LGTM overall — Snowflake was the last outlier fetching a full result set to slice in Python, so pushing LIMIT into the SQL is a good fix, and the newline-delimited wrap + trailing-line-comment test are a nice touch. The lazy `snowflake.connector` import (and `_programming_error()`) to keep the unit tests free of the optional extra is reasonable.

One nit on the `ORDER BY` comment:

> statements that already contain an ORDER BY keep their ordering under the outer LIMIT

Per Snowflake's docs, an `ORDER BY` inside a subquery isn't guaranteed to survive to the outer query, so `ORDER BY … LIMIT N` previews can now return a different set of rows than the old Python-slice path (which ran the full ordered query, then took the first N). The behavior change itself is fine — it matches all the other pushdown connectors — but the comment asserts an ordering guarantee Snowflake doesn't make. Suggest softening it to note the outer LIMIT yields an arbitrary N rows unless ORDER BY and LIMIT sit at the same level.

Non-blocking follow-up thought: `_strip_trailing_semicolon` + the wrap expression are now duplicated across ~8 connectors with small differences (alias name, single-line vs newline, `{limit}` vs `{int(limit)}`). A shared `wrap_with_limit(sql, limit)` in `connector/base.py` would unify all of that — including the comment-safety this PR adds — in one place.

goldmedal (approved):

Bartok9:
Good catch — you're right, that comment overstates the guarantee. Snowflake doesn't guarantee a subquery's `ORDER BY` survives to the outer query, so `ORDER BY … LIMIT N` previews can now return a different N rows than the old full-query-then-Python-slice path. The behavior is consistent with how the other connectors already push LIMIT down, so the change is fine, but the code comment is misleading. I'll follow up with a one-line PR to reword it to something like "the outer LIMIT bounds the row count; inner ORDER BY is best-effort and not guaranteed to be preserved per Snowflake semantics." Thanks for the precise reference.

PR #2457, opened 2026-07-09
fix(athena): push LIMIT into SQL and strip trailing semicolon on wrap

Bartok9:
## Summary

- Push `query(..., limit=N)` into Athena SQL via `SELECT * FROM (<sql>) AS _wren_sub LIMIT N`.
- Strip a trailing semicolon before wrapping so `;`-terminated user SQL cannot break composition.

## Motivation

Previously the connector ran the full statement and used `table.slice(0, limit)`. Athena (Presto/Trino-flavoured) can stop early when LIMIT is in the plan; client-side slicing paid full scan + transfer cost. Same pattern as redshift/postgres/clickhouse/snowflake pushdown helpers.

## Verification

- `PYTHONPATH=src pytest tests/unit/test_athena_limit_pushdown.py` → **3 passed**
- Asserted executed SQL for `SELECT 1;` + `limit=3` is  
  `SELECT * FROM (SELECT 1) AS _wren_sub LIMIT 3`
- Without `limit`, original SQL is executed unchanged.

## License

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Athena query execution by pushing `limit` into the generated SQL via subquery wrapping and appending `LIMIT <n>` (instead of slicing results in Python).
  * Clean input SQL by removing only trailing semicolons/extra whitespace; trailing `--` comments are handled safely.
  * Enhanced error reporting by recording the exact composed SQL that was executed.

* **Tests**
  * Added coverage for trailing semicolon stripping (preserving semicolons inside quoted strings) and correct handling of trailing line comments.
  * Expanded Athena connector tests to verify SQL wrapping/`LIMIT` pushdown and that returned Arrow tables respect the requested row count.

### DIFF
--- a/core/wren/src/wren/connector/athena.py
+++ b/core/wren/src/wren/connector/athena.py
@@ -315,13 +315,20 @@ class AthenaConnector(ConnectorABC):
         self.connection = connect(**_build_connect_kwargs(connection_info))
 
     def query(self, sql: str, limit: int | None = None) -> pa.Table:
+        # Push LIMIT into Athena when requested so Presto/Trino-flavoured
+        # engines can stop early instead of us downloading a full result and
+        # slicing in Python. Subquery-wrap + trailing-semicolon strip keeps
+        # composition valid for client SQL terminated with ``;``.
+        executed = sql
+        if limit is not None:
+            executed = (
+                f"SELECT * FROM ({_strip_trailing_semicolon(sql)}) "
+                f"AS _wren_sub LIMIT {int(limit)}"
+            )
         try:
             with contextlib.closing(self.connection.cursor()) as cursor:
-                cursor.execute(sql)
-                table = _build_athena_arrow_table(cursor)
-            if limit is not None:
-                table = table.slice(0, limit)
-            return table
+                cursor.execute(executed)
+                return _build_athena_arrow_table(cursor)
         except (WrenError, TimeoutError):
             raise
         except Exception as e:
@@ -329,7 +336,7 @@ class AthenaConnector(ConnectorABC):
                 ErrorCode.INVALID_SQL,
                 str(e),
                 phase=ErrorPhase.SQL_EXECUTION,
-                metadata={DIALECT_SQL: sql},
+                metadata={DIALECT_SQL: executed},
             ) from e
 
     def dry_run(self, sql: str) -> None:
--- a/core/wren/src/wren/connector/athena.py
+++ b/core/wren/src/wren/connector/athena.py
@@ -321,9 +321,13 @@ class AthenaConnector(ConnectorABC):
         # composition valid for client SQL terminated with ``;``.
         executed = sql
         if limit is not None:
+            # Multiline wrap: trailing `--` line comments must not eat the
+            # closing `) AS _wren_sub LIMIT n` (goldmedal review on #2457;
+            # same guard as Snowflake #2456).
             executed = (
-                f"SELECT * FROM ({_strip_trailing_semicolon(sql)}) "
-                f"AS _wren_sub LIMIT {int(limit)}"
+                "SELECT * FROM (\n"
+                f"{_strip_trailing_semicolon(sql)}\n"
+                f") AS _wren_sub LIMIT {int(limit)}"
             )
         try:
             with contextlib.closing(self.connection.cursor()) as cursor:
--- a/core/wren/src/wren/connector/athena.py
+++ b/core/wren/src/wren/connector/athena.py
@@ -321,9 +321,10 @@ class AthenaConnector(ConnectorABC):
         # composition valid for client SQL terminated with ``;``.
         executed = sql
         if limit is not None:
-            # Multiline wrap: trailing `--` line comments must not eat the
-            # closing `) AS _wren_sub LIMIT n` (goldmedal review on #2457;
-            # same guard as Snowflake #2456).
+            # Multiline wrap so a trailing `-- line comment` in the inner SQL
+            # is terminated by the newline instead of swallowing the closing
+            # `) AS _wren_sub LIMIT n`. (Single-line sibling connectors don't
+            # guard this.)
             executed = (
                 "SELECT * FROM (\n"
                 f"{_strip_trailing_semicolon(sql)}\n"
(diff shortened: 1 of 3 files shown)


goldmedal:
Nice cleanup — pushing LIMIT server-side brings Athena in line with the other pushdown connectors (postgres/trino/redshift/clickhouse), and dropping the `table.slice()` is the right call. Core change looks correct and the tests are solid.

One thing worth fixing before merge:

**Single-line wrap is broken by a trailing `--` line comment.** Athena is Presto/Trino-flavored, so `--` line comments run to end-of-line. With the current wrap:

```
query("SELECT 1 -- pick", limit=3)
→ "SELECT * FROM (SELECT 1 -- pick) AS _wren_sub LIMIT 3"
```

everything after `--` is commented out, leaving `SELECT * FROM (SELECT 1` — unbalanced parens / syntax error. `_strip_trailing_semicolon` doesn't help here (there's no trailing `;`).

This is exactly the case flagged on the sibling Snowflake PR (#2456), which fixed it by putting the inner SQL on its own line and adding a regression test. Applying the same guard here keeps the two PRs consistent:

```python
executed = (
    "SELECT * FROM (\n"
    f"{_strip_trailing_semicolon(sql)}\n"
    f") AS _wren_sub LIMIT {int(limit)}"
)
```

plus a small test mirroring `test_query_limit_survives_trailing_line_comment` from the Snowflake PR.

For fairness: the single-line form is shared by several existing connectors, so this fragility is pre-existing across the family — but within this Snowflake+Athena series it's an avoidable inconsistency, and a trailing comment in generated SQL isn't far-fetched.

Non-blocking, but a thought for a follow-up: `_strip_trailing_semicolon` + the wrap expression are now duplicated across ~8 connectors with cosmetic drift (alias name, newline-or-not, `{limit}` vs `{int(limit)}`). A single `wrap_with_limit(sql, limit)` helper in `connector/base.py` would unify the comment-safety and injection-safety behavior in one place.

goldmedal (commented):
## Review

**Verdict: sound change, safe to merge after fixing one misleading comment.** The core swap — full result + `table.slice(0, limit)` → wrapping in `SELECT * FROM (<sql>) AS _wren_sub LIMIT N` — is correct, well-tested, and consistent with the existing pushdown helpers in `postgres`/`trino`/`clickhouse`/`duckdb`/`redshift`/`canner`/`datafusion`/`oracle`/`databricks`. No public API / MDL / schema surface is touched, and the `fix:` prefix is valid for release-please. Athena is AWS-only and can't be testcontainerized, so unit-only coverage is the right call.

### Findings (severity order)

**1. [medium — accuracy] The new inline comment references a Snowflake guard that doesn't exist.**
The wrap block comments `... same guard as Snowflake #2456`, but `snowflake.py` still does client-side `arrow_table.slice(0, limit)` and has no `_strip_trailing_semicolon` helper and no subquery-`LIMIT` wrapping — there is no analogous guard there. Additionally, Athena is the *only* connector using a multiline wrap (`(\n{sql}\n)`); every sibling is single-line, so the shapes wouldn't match even if Snowflake did push down. The cross-reference is incorrect and will send the next maintainer on a dead-end. Please drop it and keep only the real justification (see #2).

**2. [low — style/consistency] Reviewer handle + PR numbers baked into permanent source.**
No sibling connector embeds PR numbers or reviewer names in its wrap comment. These references go stale immediately and read as review-thread residue. The substantive reason, though, is genuinely good and worth keeping — suggest replacing the whole parenthetical with something like:

> Multiline wrap so a trailing `-- line comment` in the inner SQL is terminated by the newline instead of swallowing the closing `) AS _wren_sub LIMIT n`. (Single-line sibling connectors don't guard this.)

**3. [low — tests] `test_athena_limit_pushdown.py` over-mocks and duplicates existing coverage.**
It bypasses `__init__` via `__new__`, hand-rolls a `_CM` context manager, and patches `contextlib.closing` — brittle against refactors. And `test_query_pushes_limit_into_sql` asserts the same wrapped SQL that `test_connector_query_returns_arrow_table_and_respects_limit` already asserts in `test_athena_connector.py` using that file's existing `_make_cursor` harness. The one net-new case worth keeping — `test_query_limit_survives_trailing_line_comment` — could live in the existing file and reuse that harness instead of standing up a parallel mocking style.

**4. [nit — pre-existing, cross-connector] `;` followed by a trailing comment isn't stripped.**
`_TRAILING_SEMICOLONS_RE = r"[;\s]+\Z"` only anchors at the end, so `SELECT 1; -- foo` keeps the interior `;` and the wrap becomes invalid Trino SQL. Shared by every sibling helper and not realistic planner output — noting for completeness, not asking for a change here.

### Things done right (no action)
- `LIMIT {int(limit)}` casts to int, closing the injection vector — safer than siblings that interpolate `{limit}` raw.
- `if limit is not None` correctly handles `limit=0` (empty table) rather than the buggy `if limit:`.
- Error metadata switched to the executed SQL, so failures show what actually ran.
- No ORDER-BY determinism regression: the old `slice(0, N)` on an unordered result was equally non-deterministic; this is a preview-row cap, not a semantic limit.

Bartok9:
Thanks @goldmedal — addressed the review.

**Fixed the misleading comment (findings #1 + #2).** Dropped the incorrect `Snowflake #2456` cross-reference — you're right, `snowflake.py` still does client-side `arrow_table.slice(0, limit)` with no `_strip_trailing_semicolon` and no subquery wrap, so there's no analogous guard there. Also removed the PR-number/reviewer-handle residue. Replaced with just the substantive reason, using your suggested wording:

```python
# Multiline wrap so a trailing `-- line comment` in the inner SQL
# is terminated by the newline instead of swallowing the closing
# `) AS _wren_sub LIMIT n`. (Single-line sibling connectors don't guard this.)
```

On the non-blocking items:
- **#3 (test over-mocking / duplication):** happy to fold the trailing-comment case into `test_athena_connector.py` using the existing `_make_cursor` harness and drop the parallel mocking file in a follow-up if you'd prefer that here rather than a separate cleanup.
- **#4 (`;` + trailing comment) and the `wrap_with_limit()` base helper:** agree both are cross-connector and best as a dedicated refactor unifying all ~8 sibling helpers, rather than growing this Athena-scoped PR.

Pushed as 895460bf.

goldmedal (approved):
👍

PR #2465, opened 2026-07-10
fix(bigquery): push LIMIT into SQL and strip trailing semicolon

Bartok9:
## Summary

- BigQuery `query(..., limit=N)` now appends `LIMIT N` after stripping a trailing semicolon instead of relying only on `result(max_results=...)`.
- `dry_run` also strips trailing `;` for consistent single-statement validation.

## Motivation

`google.cloud.bigquery` `max_results` controls pagination of the results API; it does not push a limit into the job SQL, so expensive full-table scans still run when GenBI only needs a preview page. Other connectors already push limits into SQL (Athena append-LIMIT path, postgres subquery wrap). Match that pushdown and reuse the terminating-run semicolon strip so `SELECT 1;` cannot become multi-statement noise.

## Verification

- `core/wren/.venv/bin/python -m pytest tests/unit/test_bigquery_semicolon.py -q` — 4 passed
- Mocked client: `query("SELECT 1;", limit=5)` submits `SELECT 1\nLIMIT 5`; dry_run sends `SELECT 1`
- Did NOT change: credentials/scopes path

## License

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved BigQuery query handling by preprocessing SQL to strip trailing semicolons and apply the requested row limit via an in-SQL `LIMIT` clause when needed.
  * Updated dry-run behavior to ignore trailing semicolons while preserving the existing dry-run configuration.
  * Ensured semicolons within SQL string literals remain unchanged.
* **Tests**
  * Added unit tests covering semicolon stripping, `LIMIT` injection logic, and dry-run SQL/config behavior.

### DIFF
--- a/core/wren/src/wren/connector/bigquery.py
+++ b/core/wren/src/wren/connector/bigquery.py
@@ -22,24 +22,18 @@ def _strip_trailing_semicolon(sql: str) -> str:
     return _TRAILING_SEMICOLONS_RE.sub("", sql)
 
 
-def _has_outer_limit(sql: str) -> bool:
-    """True when *sql* already ends with a top-level LIMIT clause."""
-    cleaned = _strip_trailing_semicolon(sql).rstrip()
-    return bool(re.search(r"(?i)\bLIMIT\s+\d+(\s+OFFSET\s+\d+)?\s*\Z", cleaned))
-
-
 def _apply_limit(sql: str, limit: int) -> str:
-    """Push LIMIT into SQL rather than relying solely on API max_results.
+    """Push LIMIT into SQL via outer subquery wrap.
 
     ``max_results`` only caps the page size of job results-reading; it does
-    not rewrite the query plan and still fetches a full result set stage that
-    can be expensive. Append ``LIMIT n`` after stripping a trailing semicolon
-    (mirrors Athena/MySQL pushdown helpers) so the engine short-circuits.
+    not rewrite the query plan. Wrapping as ``SELECT * FROM (sql) LIMIT n``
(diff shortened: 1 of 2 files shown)


Bartok9:
Thanks @coderabbitai — applied the cleanup in 82555c64: dropped the dead `fake_bq`/outer `patch.dict` block and added the missing `assert kwargs["job_config"].use_query_cache is False` assertion so the dry-run job config contract is fully verified. All 4 tests still pass.

Bartok9:
Addressed CodeRabbit: outer LIMIT already present → no second LIMIT appended.

goldmedal:
https://github.com/Canner/WrenAI/pull/2478 fixed the building failure of `wren-core-py`. Could you rebase on main to fix it?

Bartok9:
@goldmedal Done — rebased onto current `main` (which now includes #2478's `wren-core-py` build fix). Force-pushed; CI should be green now. Thanks!

goldmedal:
This PR should also change to the shared `strip_trailing_semicolon` function.

goldmedal:
@Bartok9, could you modify this PR? I think it should use the shared `strip_trailing_semicolon` function

Bartok9:
@goldmedal Done — switched to the shared `strip_trailing_semicolon` from `wren.connector.base` (dropped the local duplicate), matching redshift/trino/mssql/clickhouse. Also rebased onto current main and cleaned up the test imports. All unit tests pass and ruff is green. Thanks!

goldmedal (approved):
Thanks @Bartok9

PR #2468, opened 2026-07-10
feat(memory): replace LanceDB with Qdrant + Volcengine Ark embeddings

ahhblss:
- vector store: LanceDB -> Qdrant (remote server, QDRANT_URL)
- embeddings: sentence-transformers -> Volcengine Ark (openai SDK, VOLC_ARK_API_KEY)
- drop lancedb/sentence-transformers/torch; add qdrant-client/openai
- CLI: --path -> --url; forget --id takes Qdrant point id (str); .env auto-load
- SDK: LocalLanceDBMemoryProvider -> QdrantMemoryProvider; enable via QDRANT_URL
- WREN_EMBEDDING_PROVIDER=fake for local dev without API key
- CI: Qdrant service container; drop huggingface cache
- tests + docs updated

BREAKING CHANGE: `wren memory forget --id` takes a Qdrant point id (str), not a positional index (int); CLI `--path` renamed to `--url`; SDK memory auto-enable is QDRANT_URL-based, not `.wren/memory/` directory-based.

## Summary by CodeRabbit

* **New Features**
  * Memory storage and semantic recall now use Qdrant instead of local storage.
  * Added support for Volcengine Ark-compatible embeddings, with a deterministic option for offline testing.
  * Memory tools now activate through Qdrant configuration and environment variables.
  * Updated memory CLI commands to connect using Qdrant URLs and string-based record IDs.

* **Documentation**
  * Updated installation guides, CLI references, migration instructions, architecture notes, and SDK documentation for the new memory backend.

* **Tests**
  * Updated integration and unit tests to validate Qdrant-backed memory workflows.


goldmedal:
Thanks for this — dropping the torch/sentence-transformers weight is a real win, and the new `EmbeddingProvider` ABC and `FakeEmbedding` test seam are nice cleanups. My main concern is that the change is a **forced, breaking switch** rather than an added option, and I'd like to see it land as **selectable providers** instead.

**Two orthogonal changes are bundled into one mandatory path.** The PR couples (a) the vector store (LanceDB → Qdrant) and (b) the embeddings (sentence-transformers → Volcengine Ark). These are independent axes, and each removes a capability rather than adding one.

**The bigger issue is the out-of-box experience for an OSS project.** After this change, using semantic memory *requires* both a running Qdrant server (`QDRANT_URL`) **and** a Volcengine Ark API key (`ark.cn-beijing.volces.com`, a region-specific commercial API). That turns a zero-config, offline, local-file feature into "stand up a server + obtain a regional API key," which many users won't be able to do. For the default path I think we should keep something that works offline with no credentials.

**It's also breaking with no migration.** Existing `~/.wren/memory/` LanceDB data is orphaned, `--path` → `--url`, and `forget --id` changes from a positional int to a Qdrant point-id string. A selectable design keeps existing users working.

**Good news: the seams for making this selectable already exist**, so I don't think this is a big lift:
- `index_backend.py` already has the `MemoryIndex` ABC + a `WREN_MEMORY_BACKEND=grep|lancedb` selector — extending it to `grep|lancedb|qdrant` is exactly what that pattern is for.
- Both the old and new `MemoryStore` expose an *identical* public method surface, so it's effectively already the vector-store interface — we can keep both implementations side by side.
- The old LanceDB store already computed vectors explicitly and derived its dimension dynamically, so it can consume the new `EmbeddingProvider` as-is — no reason to delete it.

**Suggested shape:**
- **Store**: `WREN_MEMORY_BACKEND=grep|lancedb|qdrant`, both store impls retained; default to the local/offline store when its extra is present.
- **Embeddings**: `WREN_EMBEDDING_PROVIDER=sentence-transformers|openai|volc-ark|fake`; size the index dim from `embedding.dim`. Make the OpenAI-compatible client generic with a configurable `base_url` so Volcengine Ark is just one preset rather than the hardcoded default.
- **Extras**: split so torch is opt-in (e.g. `memory` = local default, `memory-qdrant` / `memory-openai` for the API path) — this still delivers your "no mandatory torch" goal without forcing a remote provider on everyone.
- Keep Qdrant + Ark fully supported — just as *a* choice, not *the* choice.

Happy to help with the selectable wiring if useful. What was the main driver here — the torch/install weight, or a need for a shared/remote index in deployment? That'd help decide which piece to prioritize.

PR #2472, opened 2026-07-10
fix(trino): sanitize bracketed userinfo and URL-decode credentials

Bartok9:
## Summary

- Escape raw `[` / `]` in Trino URL userinfo before `urlparse` so bracketed passwords no longer raise ValueError on IPv6 netloc checks.
- Keep real IPv6 host parsing (`[::1]`) intact and continue URL-decoding credentials/catalog/schema with `unquote` (not `unquote_plus`) so literal `+` is preserved.
- Move the pure Docker-free URL-parser regression tests into `tests/unit/test_trino_parser.py` so the `test-unit` CI job covers them (CodeRabbit nitpick).

## Motivation

Follows / completes #2435.

Passwords containing raw brackets in `trino://` connection URLs crash `urllib.parse.urlparse` because it treats any bare `[` / `]` in netloc as IPv6 markers. This is a small, MySQL-style sanitizer limited to the userinfo segment.

## Verification

- Local pure-helper checks on `_parse_trino_url`:
  - `trino://user:p[a]ss@host:8080/catalog/schema` → password `p[a]ss`
  - `trino://user:p[a]ss@[::1]:8080/catalog/schema` → host `::1`, password `p[a]ss`
  - percent-encoded creds still decode
  - literal `+` preserved
- Diff is 2 files only vs `main`.
- Earlier CI reds on #2455 were Cargo.lock `--locked` maturin failures across unit/postgres/mysql/ui/memory (infra/stale, shared with other open PRs), not trino-logic failures.

## Notes

#2455 closed after a force-push mishap; this is a clean single-commit recreate on current `main` with the CodeRabbit nitpick applied.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Trino connection URL parsing when credentials include literal square brackets (`[`/`]`), ensuring they no longer break parsing and are retained correctly.
  * Correctly preserves literal `+` characters in credentials and decodes `%`-escapes as expected.
  * Maintains proper parsing of IPv6 hosts in bracket notation (`[::1]`) even when the password contains bracket characters.

### DIFF
--- a/core/wren/src/wren/connector/trino.py
+++ b/core/wren/src/wren/connector/trino.py
@@ -295,13 +295,45 @@ def _build_trino_connect_kwargs(connection_info) -> dict:
     return out
 
 
+def _parse_trino_connection_url(url: str):
+    """Parse a Trino URL after escaping raw brackets in userinfo only.
+
+    ``urllib.parse.urlparse`` treats ``[`` / ``]`` anywhere in the netloc as
+    IPv6 host delimiters. That is correct for hosts like ``trino://alice@[::1]``
+    but it breaks on raw credentials such as ``trino://user:p[a]ss@host/c/s``.
+    Sanitise only the userinfo segment so valid IPv6 hosts still parse.
+    Mirrors the MySQL connector helper.
+    """
+    scheme_idx = url.find("://")
+    if scheme_idx == -1:
+        return urlparse(url)
+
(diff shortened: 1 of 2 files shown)


goldmedal:

Bartok9:

goldmedal (approved):
Thanks @Bartok9 👍

Issue #2441, opened 2026-07-06
refactor(wren): consolidate per-connector strip_trailing_semicolon into a shared helper in connector/base.py


goldmedal:
## Problem

The trailing-semicolon stripping helper used before subquery-wrapping user SQL is duplicated across the `core/wren` connectors, with three slightly different implementations:

- **9 verbatim copies** of the regex helper `_TRAILING_SEMICOLONS_RE = re.compile(r"[;\s]+\Z")` + `_strip_trailing_semicolon()`: `athena`, `canner`, `clickhouse`, `databricks`, `datafusion`, `duckdb`, `oracle`, `postgres`, `redshift`.
- `trino` uses `sql.rstrip().removesuffix(";").rstrip()`.
- `mysql` uses inline `rstrip(';')` (in two places).

The variants diverge on **multiple trailing semicolons** (e.g. `SELECT 1 ; ;`): the regex strips the whole terminating run, `removesuffix(";")` strips only one, and `rstrip(';')` differs again.

## Proposed refactor

Extract a single shared helper and migrate every connector onto it:

- Add a module-level `strip_trailing_semicolon()` in `core/wren/src/wren/connector/base.py` using the robust regex (`[;\s]+\Z`). Every connector already imports `ConnectorABC` from `base`, so there is no circular-import concern.
- Migrate all 9 regex-based connectors to call it, dropping their local copies.
- Migrate `trino` and `mysql` onto the shared helper too, unifying the multiple-trailing-semicolon behavior onto the regex semantics.
- Add a focused unit test for the shared helper: single/multiple trailing semicolons, trailing whitespace, and preservation of semicolons inside string literals.

## Context

- Follow-up to the per-connector semicolon bugfixes (#2407 postgres, #2411 duckdb, #2420 redshift, #2423 oracle, #2430 datafusion, plus athena/databricks/canner/clickhouse).
- Agreed in the review thread on #2430: https://github.com/Canner/WrenAI/pull/2430#issuecomment-4888479641 (author plan: https://github.com/Canner/WrenAI/pull/2430#issuecomment-4888484716)
- Scope: `core/wren` (Apache-2.0 path). Pure refactor — no behavior change except unifying the multiple-trailing-semicolon edge case onto the regex semantics.

Bartok9:
[...]

PR #2474, opened 2026-07-11
fix(trino): strip multi trailing semicolons before subquery wrap

Bartok9:
## Summary

- Strengthen Trino `_strip_trailing_semicolon` to drop the full terminating run of `;` / whitespace (not just one `removesuffix(";")`).
- `SELECT 1;;` / `; ;\n` no longer break limit/dry_run subquery wraps.

## Motivation

Limit/dry_run wrap user SQL as `SELECT * FROM ({sql}) AS _sub LIMIT N`. A single trailing semicolon was handled, but double terminators (common from copy-paste / client dsn tooling) still left `;)` in the wrapped SQL. Aligns with postgres/canner/clickhouse.

Apache-2.0 path: `core/wren/**` only.

## Verification

```
........                                                                 [100%]
8 passed in 0.23s
```

Python check on main (behavior proof):

- before: `"SELECT 1;;".rstrip().removesuffix(";").rstrip()` → `"SELECT 1;"` (still broken)
- after: helper → `"SELECT 1"`

## Duplicates

No open issue/PR solely for multi-semicolon Trino strip (semicolon work historically landed one-terminator helpers). Focused non-shim fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Trino SQL handling by stripping only a terminating run of trailing semicolons (including surrounding whitespace), while preserving semicolons elsewhere (including within string literals).
* **Tests**
  * Expanded unit test coverage with additional edge cases for multiple trailing semicolons and semicolons followed by spaces/newlines.

### DIFF
--- a/core/wren/src/wren/connector/trino.py
+++ b/core/wren/src/wren/connector/trino.py
@@ -8,9 +8,9 @@ lex with sqlglot to build an equivalent PyArrow schema.
 from __future__ import annotations
 
 import contextlib
-import re
 import datetime as dtlib
 import json
+import re
 from decimal import Decimal as PyDecimal
 from urllib.parse import parse_qsl, unquote, urlparse
 
--- a/core/wren/src/wren/connector/trino.py
+++ b/core/wren/src/wren/connector/trino.py
@@ -8,6 +8,7 @@ lex with sqlglot to build an equivalent PyArrow schema.
 from __future__ import annotations
 
 import contextlib
+import re
 import datetime as dtlib
 import json
 from decimal import Decimal as PyDecimal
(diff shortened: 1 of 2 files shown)


goldmedal:

Bartok9:

goldmedal:
This overlaps with #2480 — I think we can consolidate on that one.

Both make the same functional change at the same call sites:

- Here: the local `_strip_trailing_semicolon` body goes from `sql.rstrip().removesuffix(";").rstrip()` (strips a single trailing `;`) to `_TRAILING_SEMICOLONS_RE.sub("", sql)` (strips the whole terminating run).
- #2480: deletes trino's local helper and imports the shared `strip_trailing_semicolon` from `connector/base.py`, which is the same `[;\s]+\Z` regex, at the same call sites.

So the outcome is identical (`SELECT 1;;` / `SELECT 1 ; ;` now strip fully), but #2480 also removes the per-connector copy instead of keeping one — which is the whole point of the consolidation (#2441). The multi-semicolon fix from this PR is fully preserved by #2480, so nothing is lost.

Plan is to keep #2480 and close this one once it lands. Thanks for the fix — it's carried over intact.

Bartok9:
Agreed — #2480 folding both call sites into the shared `strip_trailing_semicolon` is the right move over keeping trino's local copy (#2441). Behavior's identical and the multi-semicolon fix is preserved, so nothing lost. Feel free to close this in favor of #2480. Thanks @goldmedal! 🙏

goldmedal:
@Bartok9, there are some conflicts. I think the PR should also use the shared trailing function.

Bartok9:
Thanks @goldmedal! Since #2480 has now merged and carries this multi-semicolon fix intact via the shared `strip_trailing_semicolon`, this PR is fully superseded — closing in favor of #2480 as we agreed. Nothing lost. 🙏

PR #2475, opened 2026-07-11
fix(mysql): strip multi trailing semicolons before LIMIT/EXPLAIN

Bartok9:
## Summary

- Replace MySQL `sql.rstrip().rstrip(";").rstrip()` with a terminating-run regex strip before `LIMIT` injection and `EXPLAIN`.
- Fixes `SELECT 1; ;` style inputs that still ended with `;` and broke EXPLAIN/LIMIT rewrite.

## Motivation

Chained `rstrip` only removes a contiguous run of `;` after whitespace strip. When whitespace appears **between** trailing semicolons, a `;` remains. Apache-2.0 path: `core/wren/**`.

## Verification

```
4 passed in 0.31s
```

Proof without the fix: `"SELECT 1; ;".rstrip().rstrip(";").rstrip()` → `'SELECT 1;'`.

## Duplicates

No open PR on MySQL multi-semicolon strip.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MySQL/Doris SQL generation by more reliably trimming trailing semicolons, including multiple semicolons separated by whitespace.
  * `LIMIT` queries and `EXPLAIN` dry-runs now construct SQL consistently after semicolon trimming.
  * Preserves semicolons that appear inside quoted string literals.

* **Tests**
  * Added unit coverage for single/double trailing semicolons, multiple trailing semicolons with interior whitespace, and semicolons inside string literals.

### DIFF
--- a/core/wren/src/wren/connector/mysql.py
+++ b/core/wren/src/wren/connector/mysql.py
@@ -10,9 +10,8 @@ only differs in how the connection is opened.
 
 from __future__ import annotations
 
-import re
-
 import json
+import re
 from contextlib import closing
 from decimal import Decimal as PyDecimal
 from functools import cache
@@ -35,7 +34,6 @@ def _strip_trailing_semicolon(sql: str) -> str:
     *between* trailing semicolons (``SELECT 1; ;`` → still ends with ``;``).
     """
     return _TRAILING_SEMICOLONS_RE.sub("", sql)
-
 
 
 def _apply_limit(sql: str, limit: int) -> str:
--- a/core/wren/src/wren/connector/mysql.py
+++ b/core/wren/src/wren/connector/mysql.py
@@ -10,6 +10,8 @@ only differs in how the connection is opened.
 
 from __future__ import annotations
 
+import re
+
 import json
 from contextlib import closing
 from decimal import Decimal as PyDecimal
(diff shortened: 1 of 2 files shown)


goldmedal:

Bartok9:

goldmedal:
This overlaps with #2480 — I think we can consolidate on that one.

Both make the same functional change at the same call sites:

- Here: adds a local regex `_strip_trailing_semicolon` and routes `_apply_limit` and the `dry_run` `EXPLAIN` through it, replacing the inline `sql.rstrip().rstrip(';').rstrip()` (which misses whitespace *between* trailing semicolons, e.g. `SELECT 1; ;`).
- #2480: routes those same two call sites (`_apply_limit`, `dry_run` EXPLAIN) through the shared `strip_trailing_semicolon` in `connector/base.py` — same `[;\s]+\Z` regex.

So the outcome is identical, but #2480 uses the shared helper instead of adding another per-connector copy — which is the point of the consolidation (#2441). The multi-semicolon fix from this PR is fully preserved by #2480, so nothing is lost.

Plan is to keep #2480 and close this one once it lands. Thanks for the fix — it's carried over intact.

Bartok9:
Sounds good — makes total sense to consolidate on #2480 via the shared `strip_trailing_semicolon` helper rather than keep a per-connector copy; that's exactly the intent of #2441. Glad the multi-semicolon behavior carries over intact. Happy to close this once #2480 lands, or you can close it whenever's convenient. Thanks @goldmedal! 🙏

goldmedal:
@Bartok9, there are some conflicts.

Bartok9:
Closing in favor of #2480, which landed the shared `strip_trailing_semicolon` helper and preserves the multi-semicolon behavior at the same MySQL call sites. Thanks @goldmedal! 🙏

PR #2478, opened 2026-07-11
fix(core-py): sync Rust crate versions in Cargo.lock

ttw225:
## Summary

The Rust core release in #2467 bumped `wren-core-base`, `wren-manifest-macro`, and `wren-semantic-core` to `0.2.0`, but `core/wren-core-py/Cargo.lock` still referenced `0.1.0`.

As a result, subsequent Wren SDK CI runs failed during `maturin build` because Cargo could not update the lockfile with `--locked`.

This PR synchronizes the three versions in the lockfile.

## Verification

- `cargo metadata --locked` with Rust 1.96 and 1.97
- `maturin build`
- Rust and Python tests

### DIFF
--- a/core/wren-core-py/Cargo.lock
+++ b/core/wren-core-py/Cargo.lock
@@ -3595,7 +3595,7 @@ dependencies = [
 
 [[package]]
 name = "wren-core-base"
-version = "0.1.0"
+version = "0.2.0"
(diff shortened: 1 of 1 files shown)


goldmedal (approved):
Thanks @ttw225 👍

PR #2479, opened 2026-07-12
fix(core): address Rust 1.97 Clippy warnings

ttw225:
## Summary

Rust 1.97 introduced the `useless_borrows_in_formatting` Clippy lint. Because CI treats warnings as errors, redundant references in formatting macros caused the Rust workflow to fail.

This PR removes those unnecessary borrows without changing runtime behavior.

## Verification

- `cargo +1.97.0 clippy --all-targets --all-features -- -D warnings`
- Cargo check
- Rust test suite

> Some Wren SDK CI jobs currently fail because the required Cargo lockfile update has not reached `main` yet. The fix is available in #2478; these jobs should pass after it is merged and this branch is rebased onto the updated `main`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved formatting for schema mismatch and access-control related error messages to make diagnostics easier to read.
  * Corrected expression formatting during measure dependency resolution, improving the accuracy of generated diagnostic output.
* **Documentation / Examples**
  * Updated the row-level access control example’s session-property printing for cleaner, more straightforward output.

### DIFF
--- a/core/wren-core/core/src/logical_plan/analyze/access_control.rs
+++ b/core/wren-core/core/src/logical_plan/analyze/access_control.rs
@@ -1054,7 +1054,7 @@ mod test {
                 Err(_) => {}
                 _ => panic!(
                     "should be error: {}",
-                    &headers.get("session_id").unwrap().as_ref().unwrap()
+                    headers.get("session_id").unwrap().as_ref().unwrap()
                 ),
             }
         }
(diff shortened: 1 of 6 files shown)


goldmedal:

goldmedal (approved):
Thanks @ttw225 👍

PR #2463, opened 2026-07-10
fix(datafusion): strip trailing semicolon before dry_run

Bartok9:
## Summary

- Strip trailing semicolons in DataFusion `dry_run` via the existing helper already used by `query`.
- User SQL ending in `;` no longer fails validation while limit wraps already worked.

## Motivation

Other connectors (postgres, redshift, duckdb, databricks, athena) neutralize trailing semicolons before dry-run / subquery wrap. DataFusion already stripped before LIMIT wraps in `query`, but `dry_run` passed SQL through verbatim, so `SELECT 1;` still hit the parser as multi-statement / invalid at plan time. Reuse `_strip_trailing_semicolon` so dry-run matches query behavior without inventing a second path.

## Verification

- `cd core/wren && .venv/bin/python -m pytest tests/unit/test_datafusion_semicolon.py -q` — 7 passed
- Real behavior proof (mocked ctx): `dry_run("SELECT 1;")` now invokes `ctx.dry_run("SELECT 1")`; `SELECT ';' AS x` preserved
- Did NOT change: query LIMIT-wrap path (already correct); Apache-2.0 path `core/**`

## License

Touches `core/wren/**` only (Apache-2.0).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `dry_run` SQL preprocessing by stripping only trailing semicolons and surrounding trailing whitespace/newlines.
  * Preserved semicolons inside quoted string literals.

* **Tests**
  * Added unit tests covering `dry_run("SELECT 1;")`, `dry_run("SELECT 1;  \n")`, and `dry_run("SELECT ';' AS x")`.

PR #2480, opened 2026-07-12
refactor(connector): share strip_trailing_semicolon in base

### LINKED ISSUE #2441: refactor(wren): consolidate per-connector strip_trailing_semicolon into a shared helper in connector/base.py

Bartok9:
## Summary

- Move `strip_trailing_semicolon` (regex terminating-run) into `wren.connector.base` and migrate connectors onto it.
- Unifies multi-semicolon behavior (including trino/mysql) and removes 10 local helper copies.

## Motivation

Closes #2441.

Per-connector copies of `_strip_trailing_semicolon` drifted (regex run vs single `removesuffix` / `rstrip`). Maintainers agreed in the #2430 review thread to consolidate on one helper in `connector/base.py`. Scope is `core/wren` only (Apache-2.0).

## Verification

- `core/wren/.venv/bin/python -m pytest tests/unit/test_strip_trailing_semicolon.py tests/unit/test_athena_semicolon.py tests/unit/test_databricks_semicolon.py tests/unit/test_redshift_semicolon.py tests/unit/test_oracle_semicolon.py tests/unit/test_datafusion_semicolon.py tests/unit/test_trino_parser.py::test_strip_trailing_semicolon tests/connectors/test_postgres.py -k "strip or semicolon" -q` — 38 passed
- `tests/connectors/test_canner.py -k strip` — 7 passed
- Did NOT change: connector query/dry_run control flow beyond the shared helper

Real behavior proof (shared helper):
`strip_trailing_semicolon("SELECT 1 ; ;") == "SELECT 1"`
`strip_trailing_semicolon("SELECT 'a;b' AS x;") == "SELECT 'a;b' AS x"`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved SQL handling across supported connectors by consistently removing only trailing semicolons (and surrounding whitespace) before wrapping queries and generating dry-run SQL.
  - Preserves semicolons inside quoted string literals.
  - Reduces failures in limited-result queries, including `EXPLAIN`-style dry runs.

- **Refactor**
  - Consolidated trailing-semicolon stripping into a shared helper used by multiple connectors.

- **Tests**
  - Added unit test coverage for the shared trailing-semicolon behavior.
  - Updated connector tests to use the centralized semicolon-handling logic.

### DIFF
--- a/core/wren/src/wren/connector/athena.py
+++ b/core/wren/src/wren/connector/athena.py
@@ -11,29 +11,14 @@ from __future__ import annotations
 import contextlib
 import datetime as dtlib
 import json
-import re
 from decimal import Decimal as PyDecimal
 from typing import Any
 
 import pyarrow as pa
 
-from wren.connector.base import ConnectorABC
+from wren.connector.base import ConnectorABC, strip_trailing_semicolon
 from wren.model.error import DIALECT_SQL, ErrorCode, ErrorPhase, WrenError
 
-_TRAILING_SEMICOLONS_RE = re.compile(r"[;\s]+\Z")
-
-
-def _strip_trailing_semicolon(sql: str) -> str:
-    """Strip any trailing ``;`` characters and surrounding whitespace.
-
-    ``dry_run`` runs ``EXPLAIN {sql}``; Athena (Trino-flavoured) rejects a
-    trailing semicolon there (``EXPLAIN SELECT 1;`` → syntax error). Only the
-    terminating run of semicolons/whitespace is stripped, so semicolons inside
-    string literals (e.g. ``SELECT 'a;b'``) are preserved. Mirrors the
-    trino/postgres/mysql connectors, which already strip before EXPLAIN or
-    subquery-wrapping.
-    """
-    return _TRAILING_SEMICOLONS_RE.sub("", sql)
 
 
 # Athena's DB-API cursor returns Trino-style type names. We delegate the
@@ -347,7 +332,7 @@ class AthenaConnector(ConnectorABC):
     def dry_run(self, sql: str) -> None:
         try:
             with contextlib.closing(self.connection.cursor()) as cursor:
-                cursor.execute(f"EXPLAIN {_strip_trailing_semicolon(sql)}")
+                cursor.execute(f"EXPLAIN {strip_trailing_semicolon(sql)}")
         except (WrenError, TimeoutError):
             raise
         except Exception as e:
--- a/core/wren/src/wren/connector/athena.py
+++ b/core/wren/src/wren/connector/athena.py
@@ -310,7 +310,7 @@ class AthenaConnector(ConnectorABC):
             # guard this.)
             executed = (
                 "SELECT * FROM (\n"
-                f"{_strip_trailing_semicolon(sql)}\n"
+                f"{strip_trailing_semicolon(sql)}\n"
                 f") AS _wren_sub LIMIT {int(limit)}"
             )
         try:
--- a/core/wren/src/wren/connector/athena.py
+++ b/core/wren/src/wren/connector/athena.py
@@ -20,7 +20,6 @@ from wren.connector.base import ConnectorABC, strip_trailing_semicolon
 from wren.model.error import DIALECT_SQL, ErrorCode, ErrorPhase, WrenError
 
 
-
 # Athena's DB-API cursor returns Trino-style type names. We delegate the
 # lexing to sqlglot so we get nested type support (array<row<a int, b varchar>>,
 # decimal(p, s), map<K, V>, etc.) for free.
--- a/core/wren/src/wren/connector/athena.py
+++ b/core/wren/src/wren/connector/athena.py
@@ -19,7 +19,6 @@ import pyarrow as pa
 from wren.connector.base import ConnectorABC, strip_trailing_semicolon
 from wren.model.error import DIALECT_SQL, ErrorCode, ErrorPhase, WrenError
 
-
 # Athena's DB-API cursor returns Trino-style type names. We delegate the
 # lexing to sqlglot so we get nested type support (array<row<a int, b varchar>>,
 # decimal(p, s), map<K, V>, etc.) for free.
--- a/core/wren/src/wren/connector/base.py
+++ b/core/wren/src/wren/connector/base.py
@@ -1,8 +1,23 @@
 from __future__ import annotations
 
+import re
 from abc import ABC, abstractmethod
 
 import pyarrow as pa
+
+_TRAILING_SEMICOLONS_RE = re.compile(r"[;\s]+\Z")
+
+
+def strip_trailing_semicolon(sql: str) -> str:
+    """Strip any trailing ``;`` characters and surrounding whitespace.
+
+    Connectors often subquery-wrap or EXPLAIN user SQL. Engines reject a
+    trailing semicolon inside those forms (e.g. ``SELECT * FROM (SELECT 1;)``
+    or ``EXPLAIN SELECT 1;``). Only the *terminating* run of semicolons and
+    whitespace is removed, so semicolons inside string literals
+    (``SELECT 'a;b'``) are preserved.
+    """
+    return _TRAILING_SEMICOLONS_RE.sub("", sql)
 
 
 class ConnectorABC(ABC):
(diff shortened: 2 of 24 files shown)


goldmedal:

Bartok9:
Thanks @goldmedal 🙏 This branch already includes the `core/wren-core-py/Cargo.lock` sync from #2478 (and the follow-up #2484 release CI fix), so the `maturin --locked` failure should be resolved. Also addressed CodeRabbit's note: dropped the now-unused `_strip_trailing_semicolon` alias in `oracle.py` and kept the `wren.model.error` import contiguous. `mysql.py` had no duplicate import. CI should be green now.

goldmedal:
One gap before this lands: **`snowflake` isn't migrated.** `core/wren/src/wren/connector/snowflake.py` still carries its own identical copy on `main`:

```python
_TRAILING_SEMICOLONS_RE = re.compile(r"[;\s]+\Z")

def _strip_trailing_semicolon(sql: str) -> str:
    ...
    return _TRAILING_SEMICOLONS_RE.sub("", sql)
```

This PR migrates 11 connectors (athena, canner, clickhouse, databricks, datafusion, duckdb, oracle, postgres, redshift, trino, mysql) but leaves snowflake with a 12th, byte-identical copy — so the "single shared helper" goal isn't fully met, and the description ("migrate every connector onto it" / "removes 10 local helper copies") slightly overstates it. It's almost certainly just because snowflake's copy landed after #2441 was scoped (it's not in that issue's enumeration).

Since it uses the exact same regex, folding it in is a one-file mechanical change:

```python
- _TRAILING_SEMICOLONS_RE = re.compile(r"[;\s]+\Z")
- def _strip_trailing_semicolon(sql: str) -> str:
-     ...
-     return _TRAILING_SEMICOLONS_RE.sub("", sql)
+ from wren.connector.base import strip_trailing_semicolon
```
(plus the two call sites, and drop the now-unused `import re` if nothing else uses it).

Suggest including snowflake here so the shared helper is genuinely the single source of truth before other PRs rebase onto it — the open snowflake fix adds another strip call site, and it'd be better for that to land on the shared helper than to grow the local copy.

Otherwise the migration itself looks clean: no leftover local helpers in the 11, all wired to `base`, and the new `test_strip_trailing_semicolon.py` nicely locks in the intended unification (`"SELECT 1 ; ;" → "SELECT 1"`, string-literal preservation) that trino/mysql previously diverged on.

Bartok9:
Good catch @goldmedal 🙏 — snowflake's copy landed after #2441 was scoped, so it wasn't in the enumeration. Just folded it in (b3ae60b): dropped snowflake's local `_strip_trailing_semicolon` + `_TRAILING_SEMICOLONS_RE` and the now-unused `import re`, wired the query() call site to `base.strip_trailing_semicolon`. All 12 connectors now share the single helper — genuine single source of truth before other PRs rebase onto it. Same regex/behavior, so no functional change.

Bartok9:
Rebased onto latest `main` and pushed (9d9cd4e). Heads up: `main` had reintroduced a fresh local `_strip_trailing_semicolon` in `athena.py` after this branch was cut, so the CI merge left a dangling call site — I've now folded athena onto `base.strip_trailing_semicolon` too and confirmed no other connector carries a local copy. `ruff check` + `ruff format --check` pass clean locally. All connectors now genuinely share the single helper. 🙏

goldmedal:
Filed the follow-up plan so the remaining trailing-semicolon work stays coherent after this consolidation lands, instead of a stream of per-connector PRs:

- #2497 — strip on the **unlimited `query()` path** across connectors (also covers trino/mysql, which have no PR for this path yet)
- #2498 — strip on **`dry_run` / `EXPLAIN` / `describe`** across connectors
- #2499 — (follow-up) **hoist the strip to the `engine.query`/`dry_run` boundary** and drop the per-connector calls

Both #2497 and #2498 are blocked on this PR (they should import the shared helper rather than re-add local copies). Redundant PRs will be closed in favor of these: #2463 (⊂ #2483), and #2474 / #2475 (their trino/mysql multi-semicolon fix is already delivered by this PR's migration).

Bartok9:
Perfect — thanks for sequencing this @goldmedal 🙏. Consolidating to a single `base.strip_trailing_semicolon` here first is exactly the right base for #2497 (unlimited `query()` path) and #2498 (`dry_run`/EXPLAIN/describe) to import rather than re-add local copies. Helper is stable, all 12 connectors share it, ruff clean. Happy to rebase promptly if #2483 or anything else lands first so those two stay unblocked.

Bartok9:
Pushed aa09bfd — fixed the failing `unit tests` job. The athena and snowflake **limit-pushdown** test modules (`test_athena_limit_pushdown.py`, `test_snowflake_limit_pushdown.py`) were still importing the now-removed private `_strip_trailing_semicolon` from their connector modules, causing an ImportError at collection time. Repointed both at the shared `wren.connector.base.strip_trailing_semicolon` (aliased), matching the pattern already used by the other migrated tests. `ruff check` + `ruff format` clean. CI should go green now. 🙏

goldmedal:
Thanks — and nice catch folding athena + snowflake back in; good to have all 12 connectors on the single helper.

Let's keep **#2480 first**, so no need to pre-rebase for a #2483-first ordering. Once this merges, #2497 and #2498 build straight on top of `base.strip_trailing_semicolon`.

#2483 will be **absorbed, not merged independently**: its unlimited-`query()` half folds into #2497 and its `dry_run` half into #2498 (both importing the shared helper rather than the datafusion-local copy), then we'll close it along with #2463. That keeps datafusion consistent with every other connector instead of fixing it via its own PR.

So the path is: merge #2480 → #2497 / #2498 → close #2483 / #2463 / #2474 / #2475. Appreciate the quick turnaround on the rebase 🙏

goldmedal (approved):
Thanks @Bartok9, nice refactoring 👍

Bartok9:
🎉 Merged — thanks @goldmedal. Confirmed path: #2480 (done) → #2497 (unlimited `query()`) / #2498 (`dry_run`/EXPLAIN/describe), both importing `base.strip_trailing_semicolon` rather than any datafusion-local copy → then close #2483 / #2463 / #2474 / #2475. I'll get #2497 and #2498 rebased on the new base tip and green so datafusion lands consistent with the other 11 connectors. 🙏

PR #2481, opened 2026-07-12
fix(trino): coerce verify=false URL/kwargs for SSL connections

### LINKED ISSUE #1502: SSL connection to TRINO
Hello, Could you please help me ! 
I tried to connect to  Trino which is developed on Openshift platform. Whatever i do i get error like -failed to execute: HTTPSConnectionPool(host='dafa-test-trino-dafa-test.apps.ocptest.lbank.lt', port=443): Max retries exceeded with url: /v1/statement (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1002)'))),
I tried to attach python certificate directly to container, i tried disbale ssl validation , nothing helped, the same message. Could you please help us, advise us how we can avoid such kind of error. Disbale SSL verification for me good too, if no other way to solve it.

Bartok9:
## Summary

- Coerce Trino ``verify`` / ``ssl_verify`` / ``insecure`` connection overrides from string URL/query kwargs into real bools or CA paths before ``trino.dbapi.connect``.
- ``?verify=false`` on ``trino+https://`` no longer leaves a truthy string that still enforces system CA checks.

## Motivation

Closes #1502.

Users connecting to private/OpenShift Trino with self-signed certs set ``verify=false`` (or want a CA path). Because kwargs/query params are strings, ``\"false\"`` is truthy and verification still fails with ``CERTIFICATE_VERIFY_FAILED``.

## Verification

- `core/wren/.venv/bin/python -m pytest tests/unit/test_trino_ssl_verify.py -q` — 4 passed
- Real behavior: `_parse_trino_url(..., ?verify=false)` yields `verify is False`.

Apache path: `core/wren/**`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Trino SSL connection handling so `verify` values passed via connection options or URLs are interpreted correctly.
  * Properly handles common inputs like `"true"`, `"false"`, and `"0"` as booleans, not certificate paths.
  * Added support for SSL aliases such as `ssl_verify` and `insecure`, including correct mapping/removal of those fields.
  * Preserves CA certificate file paths for secure connections and emits a warning when SSL verification is disabled.

### DIFF
--- a/core/wren/src/wren/connector/trino.py
+++ b/core/wren/src/wren/connector/trino.py
@@ -265,6 +265,62 @@ def _build_trino_arrow_table(cursor) -> pa.Table:
     )
 
 
+
+def _coerce_trino_verify(value):
+    """Normalize a ``verify`` SSL override for ``trino.dbapi.connect``.
+
+    Trino's client accepts ``verify`` as a bool or a CA path. Connection URLs
+    and form kwargs always deliver strings, so ``?verify=false`` must not be
+    treated as a truthy path. Returns the coerced value, or ``None`` when the
+    key should be omitted (leave package defaults).
+    """
+    if value is None:
+        return None
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, (int, float)) and value in (0, 1):
+        return bool(value)
+    text = str(value).strip()
+    if text == "":
+        return None
+    low = text.lower()
+    if low in {"0", "false", "no", "off"}:
+        return False
+    if low in {"1", "true", "yes", "on"}:
+        return True
+    # Treat as CA bundle / cert path.
+    return text
+
+
+def _apply_trino_ssl_overrides(connect_kwargs: dict) -> dict:
+    """Mutate connect kwargs so SSL verify overrides are real booleans/paths.
+
+    Users connecting to private Trino (OpenShift, corporate CA) often need
+    ``verify=false`` or a custom CA. Without coercion, the string ``"false"``
+    is truthy and still attempts system CA verification (#1502).
+    """
+    if "verify" in connect_kwargs:
+        coerced = _coerce_trino_verify(connect_kwargs.pop("verify"))
+        if coerced is not None:
+            connect_kwargs["verify"] = coerced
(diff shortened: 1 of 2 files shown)


goldmedal:

goldmedal:
Nice, focused fix — the coercion correctly resolves the `verify="false"` truthy-string problem, and popping `ssl_verify`/`insecure` keeps them from reaching `trino.dbapi.connect`. LGTM overall. Two small actionable items:

**1. Add a test for the `ssl_verify` alias.** `_apply_trino_ssl_overrides` handles `ssl_verify` (trino.py:306–311), but only `insecure` has coverage (`test_apply_insecure_alias`). Cheap coverage of a real path:

```python
def test_apply_ssl_verify_alias():
    out = _apply_trino_ssl_overrides({"host": "h", "ssl_verify": "false"})
    assert out["verify"] is False
    assert "ssl_verify" not in out
```

**2. (Optional) Log when verification ends up disabled.** When `verify` resolves to `False`, the connection silently skips certificate validation. A `logger.warning(...)` on that branch would help operators notice an unexpected insecure Trino connection in production.

goldmedal:
Hi @Bartok9, what's the status of this PR? I left the comment above. Could you check it? Thanks.

Bartok9:
Thanks for the nudge and the review, @goldmedal! 🙏 Sorry for the lag. Both items are now addressed in the latest push:

1. **`ssl_verify` alias test** — added `test_apply_ssl_verify_alias` exactly as you sketched (verifies `verify=False` and the key is popped).
2. **Warning on disabled verification** — added a `logger.warning(...)` on the branch where `verify` resolves to `False`, so an unexpected insecure Trino connection is visible in the logs.

Branch is already rebased onto latest `main` (picked up the #2478/#2484 lockfile fix), all 5 unit tests pass, and `ruff format`/`check` are clean. Ready for another look whenever you have a moment.

goldmedal (approved):
Thanks @Bartok9 👍

PR #2482, opened 2026-07-12
fix(redshift): strip trailing semicolon on unlimited query path

Bartok9:
## Summary

- Redshift `query()` now strips a terminating semicolon/whitespace run even when `limit is None`.
- Keeps behavior aligned with the limited-query wrap path and other connectors that sanitize client SQL before execute.

## Motivation

Limited Redshift queries already subquery-wrap after stripping trailing ``;``. The unlimited path executed raw SQL, so client SQL ending with ``;`` could still fail depending on session/driver batching. Strip for both paths (string-literal-safe terminating-run regex already used by this connector).

## Verification

- `core/wren/.venv/bin/python -m pytest tests/unit/test_redshift_semicolon.py -q` — 5 passed
- Real behavior: `query("SELECT 1;")` sends `SELECT 1` on the unlimited path.

Apache path: `core/wren/**`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Redshift query execution when no row limit is specified by stripping trailing semicolons and surrounding whitespace.
  * Ensures submitted SQL is normalized consistently whether a limit is provided or not.
* **Tests**
  * Added a unit test to verify unlimited queries don’t send SQL that ends with `;`.

### DIFF
--- a/core/wren/src/wren/connector/redshift.py
+++ b/core/wren/src/wren/connector/redshift.py
@@ -65,6 +65,10 @@ class RedshiftConnector(ConnectorABC):
                 f"SELECT * FROM ({_strip_trailing_semicolon(sql)}) "
                 f"AS _q LIMIT {int(limit)}"
             )
+        else:
+            # Unlimited path also rejects trailing ``;`` for single statements
+            # depending on driver/session settings — strip for consistency.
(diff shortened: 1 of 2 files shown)


goldmedal:

goldmedal (approved):

PR #2484, opened 2026-07-12
ci(core): sync downstream lockfile during Rust releases

ttw225:
## Summary

Keep `core/wren-core-py/Cargo.lock` synchronized when the linked Rust core crates are released.

Release Please now updates the three corresponding lockfile entries in the release PR, preventing locked downstream builds from failing after a version bump.

## Verification

- Validated against the Release Please version bundled by `release-please-action@v4`
- Verified the generated lockfile with `cargo metadata --locked`

## Summary by CodeRabbit

* **Chores**
  * Improved automated release version tracking for additional Rust component packages.
  * Release metadata now derives component versions from Cargo lockfiles instead of manifest-based selectors, improving consistency and accuracy across builds.
  * Updates include sourcing the relevant core and semantic-core component versions from the appropriate lockfiles.

### DIFF
--- a/core/wren-core-wasm/Cargo.lock
+++ b/core/wren-core-wasm/Cargo.lock
@@ -4004,7 +4004,7 @@ dependencies = [
 
 [[package]]
 name = "wren-core-base"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "serde",
  "serde_json",
@@ -4043,7 +4043,7 @@ dependencies = [
 
 [[package]]
 name = "wren-manifest-macro"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "quote",
  "syn",
@@ -4051,7 +4051,7 @@ dependencies = [
 
 [[package]]
 name = "wren-semantic-core"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "async-trait",
  "csv",
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -56,7 +56,19 @@
     "core/wren-core-base/manifest-macro": {
       "component": "wren-manifest-macro",
       "release-type": "rust",
-      "bump-minor-pre-major": true
+      "bump-minor-pre-major": true,
+      "extra-files": [
+        {
+          "path": "/core/wren-core-py/Cargo.lock",
+          "type": "toml",
+          "jsonpath": "$.package[?(@.name.value=='wren-manifest-macro')].version"
+        },
+        {
+          "path": "/core/wren-core-wasm/Cargo.lock",
+          "type": "toml",
+          "jsonpath": "$.package[?(@.name.value=='wren-manifest-macro')].version"
+        }
+      ]
     },
     "core/wren-core-base": {
       "component": "wren-core-base",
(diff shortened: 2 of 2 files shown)


goldmedal:
Mechanics look correct — I checked the details and they hold up:

- The `$.package[?(@.name.value=='…')].version` filter form matches the existing `uv.lock` entries in this same config (that's the proven pattern for lockfile package arrays), and all three crates are present in `core/wren-core-py/Cargo.lock`.
- The three crates are single-version workspace **path deps**, so there's no `checksum` line and their cross-references in other packages' `dependencies` arrays are versionless — bumping only `version` leaves nothing stale and keeps `cargo metadata --locked` valid.
- The leading-slash `/core/wren-core-py/Cargo.lock` correctly resolves repo-root-relative, which is needed since the lockfile lives outside each triggering crate's directory.

**One gap before merge: a second committed lockfile is left stale.** There are two lockfiles that pin all three crates:

| Lockfile | Covered here? |
| --- | --- |
| `core/wren-core-py/Cargo.lock` | ✅ |
| `core/wren-core-wasm/Cargo.lock` | ❌ |

`wren-core-wasm` transitively depends on all three (→ `wren-semantic-core` → `wren-core-base` → `wren-manifest-macro`), and its committed `Cargo.lock` pins them in the identical versionless path-dep shape. After a linked bump it stays at the old version — the same drift this PR prevents, just for the wasm build. Suggest mirroring the same three entries with `"path": "/core/wren-core-wasm/Cargo.lock"`:

```json
"core/wren-core-base/manifest-macro": {
  "extra-files": [
    { "path": "/core/wren-core-py/Cargo.lock",   "type": "toml", "jsonpath": "$.package[?(@.name.value=='wren-manifest-macro')].version" },
    { "path": "/core/wren-core-wasm/Cargo.lock", "type": "toml", "jsonpath": "$.package[?(@.name.value=='wren-manifest-macro')].version" }
  ]
},
"core/wren-core-base": {
  "extra-files": [
    { "path": "Cargo.toml", "type": "toml", "jsonpath": "$.dependencies.wren-manifest-macro.version" },
    { "path": "/core/wren-core-py/Cargo.lock",   "type": "toml", "jsonpath": "$.package[?(@.name.value=='wren-core-base')].version" },
    { "path": "/core/wren-core-wasm/Cargo.lock", "type": "toml", "jsonpath": "$.package[?(@.name.value=='wren-core-base')].version" }
  ]
},
"core/wren-core": {
  "extra-files": [
    { "path": "Cargo.toml", "type": "toml", "jsonpath": "$.workspace.package.version" },
    { "path": "Cargo.toml", "type": "toml", "jsonpath": "$.workspace.dependencies.wren-core-base.version" },
    { "path": "/core/wren-core-py/Cargo.lock",   "type": "toml", "jsonpath": "$.package[?(@.name.value=='wren-semantic-core')].version" },
    { "path": "/core/wren-core-wasm/Cargo.lock", "type": "toml", "jsonpath": "$.package[?(@.name.value=='wren-semantic-core')].version" }
  ]
}
```

(One thing I couldn't confirm from the workflows: whether the wasm build runs with `--locked`/`--frozen` — a plain `cargo build` would just rewrite the lockfile, so this only becomes a hard failure under `--locked`. Either way, syncing one lockfile and leaving its sibling stale is inconsistent and only half-solves the goal, so mirroring it seems worth doing.)

ttw225:
Thanks, you're right. I confirmed that the WASM workflows currently use plain `wasm-pack build`, so Cargo refreshes a stale lockfile instead of failing.

I've mirrored the three Release Please entries to `core/wren-core-wasm/Cargo.lock` and synchronized its current entries to 0.2.0. Both downstream lockfiles now pass `cargo metadata --locked`.

I kept enforcing `--locked` in CI and publishing as a separate hardening change, since that changes the build policy rather than the release-file synchronization covered by this PR.

Sorry for the commit history churn—the follow-up commit was accidentally squashed into the original commit during rebase, so these changes now appear together in a single rewritten commit.

goldmedal (approved):
Thanks @ttw225 👍

PR #2485, opened 2026-07-12
perf(core-py): release the GIL while blocking on the tokio runtime

ttw225:
## Summary

Release the Python GIL while `PySessionContext` waits on Tokio
`Runtime::block_on`, allowing calls on different `SessionContext`
instances to run concurrently.

This covers all 13 Python-reachable `block_on` sites, including
construction, SQL transformation, query execution, table registration,
dry runs, MDL loading, and function discovery.

## Problem

Every blocking `SessionContext` call previously held the GIL for the
entire planning or execution operation. Python threads therefore
serialized even when they used independent contexts and independent
Tokio runtimes.

This also prevented an asyncio event loop from progressing while a
synchronous Wren call was offloaded to a worker thread: the worker still
held the GIL until the call completed.

## Implementation

Each blocking operation now runs inside `Python::detach`.

Python-borrowed arguments such as SQL, function names, and paths are
converted to Rust-owned values before entering the detached closure.
Session properties are also converted eagerly during construction, and
the futures do not call back into Python.

Releasing the GIL exposed a pre-existing same-context catalog race.
`transform_sql_with_ctx` re-applies the MDL catalog to shared context
state, so concurrent calls could observe a partially registered catalog.

A per-context `call_lock` preserves the previous ordering for calls on
the same context. The lock is acquired only after releasing the GIL to
avoid a lock/GIL deadlock. Calls using different contexts can run in
parallel.

`list_tables` does not use `block_on`, but it reads the same catalog
state and therefore participates in the same locking protocol.

The unused `Clone` implementation is removed because cloned wrappers
could share catalog and lock state while later diverging in their
`ctx`, `exec_ctx`, and `mdl` fields.

## Behavioral note

Methods taking `&mut self`, currently `load_mdl`, may now return a PyO3
borrow error when called concurrently with another in-flight operation
on the same object. Previously such calls implicitly waited for the GIL.

Same-context engine calls otherwise retain their previous serialized
ordering.

## Benchmark

Release build on an Apple M5 Pro with 18 cores. Each thread used its own
`SessionContext` and performed 200 `transform_sql` calls. Measurements
were collected with Python 3.14 using the abi3-py311 wheel.

| Threads | Before calls/s | Before scaling | After calls/s | After scaling |
|---:|---:|---:|---:|---:|
| 1 | 956 | 1.00x | 953 | 1.00x |
| 2 | 987 | 1.03x | 1,070 | 1.12x |
| 4 | 980 | 1.03x | 1,721 | 1.80x |
| 8 | 984 | 1.03x | 3,142 | 3.30x |
| 16 | 991 | 1.04x | 5,142 | 5.39x |

Single-thread throughput remained effectively unchanged (`956` versus
`953` calls/s). At 16 threads, throughput improved from `991` to `5,142`
calls/s, a 5.19x improvement at the same concurrency level.

Scaling remains sublinear because every context currently owns a
multi-thread Tokio runtime, which oversubscribes the machine at higher
thread counts. Sharing one process-wide runtime is planned as a
follow-up. I plan to follow up with a PR for this.

## Tests

- Added a concurrent-calls smoke test using separate contexts.
- Added a shared-context mode covering mixed methods and catalog access.
- The test checks correctness, crashes, and deadlocks without timing
  assertions.

Validation:

- `cargo test`: 37 passed
- `pytest`: 34 passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved thread-safety for Python session operations by serializing blocking runtime calls, reducing race conditions during SQL transformation, querying, table registration, and semantic-layer loading.
  * Improved Python responsiveness by releasing the interpreter lock while engine work runs.

* **Tests**
  * Added concurrent coverage for `transform_sql` across threads (both separate and shared session contexts).
  * Updated an expected SQL rewrite output to match the latest escaping/quoting behavior.

### DIFF
--- a/core/wren-core-py/src/context.rs
+++ b/core/wren-core-py/src/context.rs
@@ -26,11 +26,13 @@ use std::collections::HashMap;
 use std::hash::Hash;
 use std::ops::ControlFlow;
 use std::str::FromStr;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex, PoisonError};
 use std::vec;
 use tokio::runtime::Runtime;
 use wren_core::array::{AsArray, GenericByteArray};
-use wren_core::ast::{Expr, LimitClause, Statement, Value, ValueWithSpan, visit_statements_mut};
+use wren_core::ast::{
+    visit_statements_mut, Expr, LimitClause, Statement, Value, ValueWithSpan,
+};
 use wren_core::datatypes::GenericStringType;
 use wren_core::dialect::GenericDialect;
 use wren_core::ipc::writer::StreamWriter;
@@ -47,7 +49,6 @@ use wren_core_base::mdl::DataSource;
 
 /// The Python wrapper for the Wren Core session context.
 #[pyclass(name = "SessionContext")]
-#[derive(Clone)]
 pub struct PySessionContext {
     /// Base context — physical tables are registered here.
     /// Used as the source for `load_mdl()` (two-phase init).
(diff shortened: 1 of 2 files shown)


goldmedal:

ttw225:
Related core fix: #2495 

While validating this PR's same-context concurrency path, I traced the
underlying catalog race to `apply_wren_on_ctx` sharing and mutating the
base context's top-level catalog list. #2495 gives each
apply call a private catalog-list snapshot.

This PR keeps the per-context `call_lock` for now. Removing it and adding
same-context throughput measurements will be a separate follow-up after
both PRs land.

goldmedal (approved):
Thanks @ttw225, nice improvement 👍. It makes Wren efficient in concurrent cases.

ttw225:
Thanks for taking the time to review and merge this. Really appreciate it!

PR #2495, opened 2026-07-13
fix(core): prevent catalog races in concurrent same-context transforms

ttw225:
## Problem

`apply_wren_on_ctx` builds derived DataFusion session states from a base `SessionContext`. `SessionStateBuilder::new_from_existing` shares the base context's top-level `catalog_list`.

`register_table_with_mdl` then registers a new catalog before adding its models and views. Concurrent calls using the same base context can therefore replace or observe each other's partially constructed catalog.

This can cause:

- `table not found`
- `table already exists`
- cross-call catalog/schema contamination
- mutation of the base context's catalog list

## Fix

Create a private top-level catalog-list snapshot for each `apply_wren_on_ctx` call and attach it to the derived session states.

The isolation contract is:

- Top-level catalog membership and replacement are private to one apply call.
- The two session states created within one apply call share the same private list.
- Catalog providers that already exist at snapshot time remain `Arc`-shared, preserving access to live physical catalogs.
- Callers must complete top-level physical catalog registration before starting a transform.

An internal invariant check returns a DataFusion internal error if the final derived state does not retain the per-call private list.

## Tests

Added coverage for:

- Preserving a sentinel catalog on the base context while the derived context receives the current MDL model and view.
- Concurrent calls using one base context and catalog key, with unique marker columns detecting cross-call contamination.
- Excluding top-level catalogs registered on the base context after the snapshot.
- Retaining live visibility into providers that existed before the snapshot.

The concurrent stress test uses eight OS threads with current-thread Tokio runtimes and repeatedly contends on the same catalog key.

## Verification

- `cargo fmt --all -- --check`
- `cargo check --all-targets`
- `RUST_MIN_STACK=8388608 cargo test --lib --tests --bins` — 147 core tests and sqllogictests passed
- Concurrent isolation stress test repeated successfully

## Related

#2485 exposed this latent race by allowing concurrent Python calls to enter the Rust execution path. This PR fixes the underlying catalog-list isolation in `wren-core` and can merge independently.

The per-context `call_lock` in #2485 remains in place. Removing it and measuring same-context throughput will be a separate follow-up after both PRs land.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved catalog isolation during semantic model application by snapshotting the top-level catalog list for derived contexts.
  - Prevented concurrent applications on the same base context from interfering with each other.
  - Ensured derived contexts keep a consistent view of newly added models/views while still reflecting live updates inside catalogs that existed at apply time.
- **Tests**
  - Added contract-style coverage for catalog isolation and visibility, including concurrent stress scenarios and snapshot immutability.

### DIFF
--- a/core/wren-core/core/src/mdl/context.rs
+++ b/core/wren-core/core/src/mdl/context.rs
@@ -73,6 +73,11 @@ fn clone_catalog_list(
 }
 
 /// Apply Wren Rules to the context for sql generation.
+///
+/// Callers must complete top-level physical-catalog registration before
+/// starting a transform. Catalog enumeration produces a best-effort,
+/// non-atomic snapshot, so concurrent top-level registrations may or may not
+/// be observed by the derived context.
 pub async fn apply_wren_on_ctx(
     ctx: &SessionContext,
     analyzed_mdl: Arc<AnalyzedWrenMDL>,
--- a/core/wren-core/core/src/mdl/context.rs
+++ b/core/wren-core/core/src/mdl/context.rs
@@ -15,10 +15,11 @@ use crate::mdl::type_planner::WrenTypePlanner;
 use crate::mdl::{AnalyzedWrenMDL, SessionStateRef};
 use async_trait::async_trait;
 use datafusion::arrow::datatypes::SchemaRef;
-use datafusion::catalog::memory::MemoryCatalogProvider;
+use datafusion::catalog::memory::{MemoryCatalogProvider, MemoryCatalogProviderList};
 use datafusion::catalog::CatalogProvider;
+use datafusion::catalog::CatalogProviderList;
 use datafusion::catalog::{MemorySchemaProvider, Session};
-use datafusion::common::Result;
+use datafusion::common::{internal_err, Result};
 use datafusion::datasource::{TableProvider, TableType, ViewTable};
 use datafusion::execution::session_state::SessionStateBuilder;
 use datafusion::logical_expr::Expr;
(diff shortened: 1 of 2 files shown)


goldmedal:
Nice fix — the root cause is real: `new_from_existing` moves `catalog_list: Some(existing.catalog_list)`, so the derived state shares the base context's list Arc and `register_table_with_mdl`'s `register_catalog` mutates it. The private per-call snapshot is the right level to isolate it, and having both derived states share the *same* private Arc (enforced by the `Arc::ptr_eq` invariant) is exactly what keeps the analyzer-rule state and the registration ctx consistent.

One thing worth calling out for the consumer side (`wren-core-py`), which is the path that actually depends on this behavior:

`register_parquet` / `register_csv` document *"Tables registered on base_ctx are visible to exec_ctx via shared catalog."* Before this change that was trivially true — `exec_ctx` literally shared `base_ctx`'s list. After this change `exec_ctx` holds a private snapshot, so that contract now survives **only** via the Layer-2 semantics (internals of catalogs that existed at snapshot time stay Arc-shared): `register_parquet/csv` target `base_ctx`'s default catalog, which exists at apply time, so mutations remain visible.

Two implications:

1. The invariant is now load-bearing but implicit on the Python side: *physical tables must land in a catalog that already exists at apply time.* If a caller ever registered a **new top-level catalog** on `base_ctx` after `exec_ctx` is built, it would silently not be visible (it postdates the snapshot). Might be worth mirroring the "complete top-level physical-catalog registration before starting a transform" note onto the Python methods too.

2. There's currently no `wren-core-py` test exercising `register_parquet/csv` → query via `exec_ctx`. The new Layer-2 test here is a good proxy, but a small Python integration test (register a parquet, `load_mdl`, assert the table resolves through `exec_ctx`) would lock the contract at the boundary that actually relies on it.

Neither blocks this PR (it's wren-core-only and correct as-is) — just flagging the follow-up so the guarantee doesn't quietly regress later.

ttw225:
Thanks for the careful review — this is very helpful. Your reading is correct: the current Python methods register physical tables into `base_ctx`’s existing default catalog, so visibility through `exec_ctx` is preserved by the shared internals of that catalog provider even though the top-level catalog list is snapshotted.

I’d like to address this in a separate `wren-core-py` follow-up. The original goal was to remove the per-context `call_lock` and enable true same-context concurrency. Your suggestions add an important consumer-side safeguard to that work: I’ll clarify the `register_parquet` / `register_csv` documentation and add integration coverage for physical-table visibility through `exec_ctx`, including registration after `exec_ctx` has been created. The lock removal would also retain the same-context concurrency regression coverage and include throughput validation.

I’ll keep this PR scoped to `wren-core`. Thanks again for identifying the missing boundary coverage.

goldmedal:
Filed https://github.com/Canner/WrenAI/issues/2504 to track the follow-up task.
Thanks @ttw225, nice work 👍

goldmedal (approved):

ttw225:
Thanks for filing #2504 and for the thoughtful review!

I also opened #2510 to share one Tokio runtime across Python session
contexts and keep native thread growth bounded. It is related to the
same concurrency work, but it does not address #2504: the `call_lock`
remains in place, and the catalog documentation and integration
coverage are unchanged.

Addressing #2504 after #2510 should be cleaner, since removing the
`call_lock` will then build on a single shared worker pool instead of
per-context runtimes. I’ll handle those items separately.

PR #2502, opened 2026-07-14
docs(wren): fix cube hierarchy YAML examples

ttw225:
## What

- Use the schema-compatible map format for cube `hierarchies`.
- Document that hierarchy levels must reference declared dimensions or time dimensions.
- Update the bundled `enrich-context` skill template.

## Why

The previous list-of-objects examples passed YAML validation but produced an invalid hierarchy shape, causing hierarchy metadata to be ignored.

## Verification

- Validated the examples against `mdl.schema.json`.
- Ran context and skill tests: 146 passed.
- Confirmed `wren context build` emits a hierarchy object map.

## Summary by CodeRabbit

* **Documentation**
  * Updated cube metadata examples to use map-style hierarchy definitions.
  * Clarified that hierarchy entries map names to ordered lists of dimensions or time dimensions.
  * Documented that hierarchy levels must match declared dimension or time-dimension names.


goldmedal (approved):
Thanks @ttw225, nice catch 👍

PR #2505, opened 2026-07-15
fix(trino): fall back on Sqlglot TokenError for type strings

Bartok9:
## Summary

- `_parse_trino_data_type` now catches `sqlglot.errors.SqlglotError` (and `ValueError`), not only `ParseError`.
- Tokenizer failures (TokenError on bad `cursor.description` type strings) degrade to `pa.string()` instead of aborting result materialization.

## Motivation

`type_mapping.parse_type` already documents and tests that TokenError is a SqlglotError but **not** a ParseError, so catching only ParseError lets tokenizer crashes escape. Trino's type parser had the same bug: a single column with a corrupt/unterminated type string could fail the whole Arrow table build for an otherwise valid query.

Apache-2.0 path: `core/wren/**`.

## Verification

- `cd core/wren && PYTHONPATH=src python -m pytest tests/unit/test_trino_parser.py -k 'tokenizer_error or unparseable or handles_none' -v` — **3 passed**
- New regression: token-error inputs (`"unterminated`, control character) return `pa.string()`
- Did NOT change: Arrow mappings for well-formed Trino types; strip_trailing_semicolon / connect paths

## Real behavior proof

```
tests/unit/test_trino_parser.py::test_parse_trino_data_type_handles_none PASSED
tests/unit/test_trino_parser.py::test_parse_trino_data_type_unparseable_falls_back PASSED
tests/unit/test_trino_parser.py::test_parse_trino_data_type_tokenizer_error_falls_back PASSED
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed or unsupported Trino data types during query processing.
  * Prevented tokenizer and parsing errors from interrupting processing by safely falling back to a string type.

### DIFF
--- a/core/wren/src/wren/connector/trino.py
+++ b/core/wren/src/wren/connector/trino.py
@@ -40,8 +40,12 @@ def _parse_trino_data_type(type_str: str | None) -> pa.DataType:
         return pa.string()
     try:
         parsed = sqlglot.parse_one(type_str, into=DataType, dialect="trino")
-    except sqlglot.errors.ParseError:
-        logger.warning(f"Failed to parse trino type string: {type_str}")
+    except (sqlglot.errors.SqlglotError, ValueError):
+        # SqlglotError covers ParseError *and* TokenError (tokenizer failures
+        # on stray control chars / unterminated quotes are NOT subclasses of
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
Thanks @Bartok9  👍

PR #2508, opened 2026-07-15
fix(type_mapping): skip non-dict rows in parse/translate_types

Bartok9:
## Summary

- `parse_types` / `translate_types` skip non-dict column rows instead of raising TypeError on `dict(col)`.

## Motivation

Batch helpers assume list-of-dicts. Dynamic schema exporters sometimes interleave null/non-mapping entries; converting those with `dict(col)` crashed the whole batch and dropped every valid column.

Apache-2.0: `core/wren/**`.

## Verification

- `cd core/wren && PYTHONPATH=src python -m pytest tests/unit/test_type_mapping.py -k non_dict -v` — **2 passed**
- Did NOT change: normalization for well-formed column dicts; CLI contracts

## Real behavior proof

```
tests/unit/test_type_mapping.py::test_parse_types_skips_non_dict_rows PASSED
tests/unit/test_type_mapping.py::test_translate_types_skips_non_dict_rows PASSED
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type parsing and translation to safely skip non-mapping entries in batch inputs instead of erroring.
  * Enhanced acceptance of mapping-like inputs (e.g., read-only mappings) while keeping output rows consistent.
  * Batch CLI commands now show a stderr note when some input rows can’t be processed.

* **Tests**
  * Added unit test coverage for batch parsing/translation behavior, including skipping invalid rows and handling mapping-like inputs.

### DIFF
--- a/core/wren/src/wren/type_mapping.py
+++ b/core/wren/src/wren/type_mapping.py
@@ -20,6 +20,8 @@ Use as a library:
 
 from __future__ import annotations
 
+from collections.abc import Mapping
+
 import sqlglot
 import sqlglot.errors
 from sqlglot.expressions import DataType
@@ -55,15 +57,21 @@ def parse_types(
 ) -> list[dict]:
     """Batch-normalize types for a list of column dicts.
 
-    Each dict must have a key matching *type_field* (default "raw_type").
+    Each mapping should have a key matching *type_field* (default "raw_type").
     Returns a new list with an added "type" key containing the normalized type.
-    Original dicts are not mutated.
+    Original mappings are not mutated.
+
+    Non-mapping rows (e.g. ``None`` padding or malformed entries from dynamic
+    schema exporters) are skipped, so ``len(out) <= len(columns)``. Any
+    ``collections.abc.Mapping`` is accepted, not just ``dict`` (e.g. SQLAlchemy
+    ``RowMapping`` or ``types.MappingProxyType``).
     """
     results = []
     for col in columns:
         # Callers occasionally pass sparse / non-mapping rows from dynamic
-        # schemas; skip rather than TypeError on dict(col).
-        if not isinstance(col, dict):
+        # schemas; skip rather than TypeError on dict(col). Accept any Mapping
+        # (not just dict) so mapping-likes from schema exporters survive.
+        if not isinstance(col, Mapping):
             continue
         row = dict(col)
         row["type"] = parse_type(row.get(type_field, ""), dialect)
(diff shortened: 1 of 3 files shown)


goldmedal (commented):
**Approve** — the minimal fix is a net improvement. A few suggestions below; none block merge, and the larger one is best as a follow-up rather than piled onto this PR.

The fix is correct and the tests are solid: they cover a mixed list and assert both surviving order (`["id", "name"]`) and specific translated types (`BIGINT`, `INT64`/`STRING`), so a regression dropping the wrong rows would fail. `dict(None)`/`dict(42)` raise `TypeError` and `dict("...")` raises `ValueError`, so the "one bad row kills the whole batch" motivation is real.

**1. Guard on `collections.abc.Mapping`, not `dict` — this touches the stated motivation.**
`dict(col)` accepts *any* mapping, but `isinstance(col, dict)` only accepts `dict` and its subclasses. The motivation cites "dynamic schema exporters," which commonly yield mapping-likes that are **not** `dict` subclasses (e.g. SQLAlchemy `RowMapping`, `types.MappingProxyType`). Those worked before and would now be silently skipped — a potential regression for exactly the described case. Prefer:

```python
from collections.abc import Mapping
...
if not isinstance(col, Mapping):
    continue
```

**2. Don't drop completely silently.** In the CLI path (`parse-types`/`translate-types` read a JSON array and write one back), a skipped row makes the output count silently smaller than the input with no signal. For a schema tool, a subtly-incomplete result is arguably worse than a loud failure. For this PR, a one-line `stderr` note from the CLI wrapper is enough, e.g. `skipped N non-mapping row(s)`.

Worth distinguishing intent: `None` is plausibly benign padding (skip quietly), whereas a `str`/`int` is more likely real corruption worth surfacing.

*Follow-up (not this PR):* the fuller version — accumulate the bad entries and report them once, with a `--strict` flag to turn the warning into an error — already has a precedent in this package (`context_cli.py` collects warnings, applies a summary threshold, and honors `--strict`). Keeping this PR minimal and doing the collector as a separate, `context_cli`-aligned change avoids loading a policy decision ("what counts as an error vs. benign") onto a small bugfix.

**Minor:** both docstrings still say "Each dict must have a key matching *type_field*" and imply 1:1 output. Add a line noting non-mapping rows are skipped, so `len(out) <= len(in)` (also nudges the docstring-coverage check CodeRabbit flagged).

Bartok9:
Thanks so much for the thorough review @goldmedal — really appreciated. Pushed 355cd2ad addressing all three:

1. **Mapping guard** — switched `isinstance(col, dict)` → `isinstance(col, collections.abc.Mapping)` in both `parse_types` and `translate_types`. You're right that `dict(col)` accepted any mapping before, so `RowMapping`/`MappingProxyType` from dynamic exporters would've been silently dropped — that's exactly the regression the motivation warned against. Added tests covering `MappingProxyType` acceptance.
2. **Don't drop silently** — the `parse-types`/`translate-types` CLI wrappers now emit a one-line `Note: skipped N non-mapping row(s)` to stderr when the output count shrinks.
3. **Docstrings** — both now note non-mapping rows are skipped so `len(out) <= len(in)`, and that any `Mapping` (not just `dict`) is accepted. Also nudges the docstring-coverage check.

Kept the fuller `--strict` + accumulate-and-summarize collector (aligned with `context_cli.py`) as a follow-up rather than piling the policy decision onto this bugfix, per your suggestion. All 52 unit tests green.

goldmedal (approved):
Thanks @Bartok9 LGTM. I filed #2528 for the follow-up enhancement.

PR #2509, opened 2026-07-15
fix(wren): treat non-string SQL as not exploratory

Bartok9:
## Summary

- `is_exploratory` returns `False` for non-string or blank SQL instead of TypeErroring through `sqlglot.parse`.

## Motivation

Store-tip / UI callers may pass `None` or other placeholders. `sqlglot.parse(None)` raises `TypeError: object of type 'NoneType' has no len()`, which crashes the exploratory tip path instead of just suppressing the tip (same outcome as empty/unparseable SQL).

Apache-2.0: `core/wren/**`.

## Verification

- `cd core/wren && PYTHONPATH=src python -m pytest tests/unit/test_sql_classify.py -v` — **15 passed**
- Did NOT change: classification for real SQL strings

## Real behavior proof

```
tests/unit/test_sql_classify.py::test_none_and_non_string_sql_return_false PASSED
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQL classification reliability by safely handling empty, whitespace-only, and non-text inputs.
  * Invalid inputs now return a consistent negative result instead of causing a parsing error.
* **Tests**
  * Added test coverage to confirm `is_exploratory` returns `False` for `None`, numeric values, and byte-string inputs.

### DIFF
--- a/core/wren/src/wren/sql_classify.py
+++ b/core/wren/src/wren/sql_classify.py
@@ -4,7 +4,7 @@ import sqlglot
 from sqlglot import exp
 
 
-def is_exploratory(sql: str) -> bool:
+def is_exploratory(sql: object) -> bool:
     """Return True if *sql* looks like an exploratory peek query.
 
     Exploratory = bare SELECT with no WHERE/GROUP BY/HAVING/aggregate.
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
Thanks @Bartok9, 👍

Issue #2504, opened 2026-07-15
Follow-up (wren-core-py): document + test physical-table visibility through exec_ctx after per-call catalog snapshot


goldmedal:
### Context

#2495 makes each `apply_wren_on_ctx` call use a **private top-level catalog-list snapshot** so concurrent same-context transforms can't race on catalog registration. That fix is scoped to `wren-core`.

As noted in the [review discussion](https://github.com/Canner/WrenAI/pull/2495#issuecomment-4976223172), this shifts a `wren-core-py` consumer guarantee onto an implicit invariant. `PySessionContext::register_parquet` / `register_csv` document *"Tables registered on base_ctx are visible to exec_ctx via shared catalog."* After #2495, `exec_ctx` holds a private snapshot of the top-level catalog list, so that guarantee survives **only** because those methods register into `base_ctx`'s **default** catalog — which exists at apply time and stays `Arc`-shared (Layer-2 shared internals). A physical table registered into a *new top-level catalog* after `exec_ctx` is built would silently not be visible.

This issue tracks the `wren-core-py`-side follow-up so that guarantee is documented and test-covered, and to carry the original same-context concurrency goal.

### Follow-up items

- [ ] **Docs** — clarify `register_parquet` / `register_csv`: physical tables register into `base_ctx`'s default catalog, and visibility through `exec_ctx` relies on that catalog already existing at apply time. State the precondition explicitly: complete top-level physical-catalog registration before starting a transform (mirroring the note now on `apply_wren_on_ctx`).
- [ ] **Tests** — add `wren-core-py` integration coverage for physical-table visibility through `exec_ctx`, including registration performed **after** `exec_ctx` has been created (register a parquet/csv, then resolve/query it via `exec_ctx`).
- [ ] **`call_lock` removal** — remove the per-context `call_lock` to enable true same-context concurrency; retain the same-context concurrency regression coverage and add throughput validation.

### Related

- Builds on #2495 (`wren-core` catalog-list isolation).

ttw225:
Thanks for putting together this follow-up and clearly outlining the remaining work!

I’m planning to take ownership of this issue and work through it in a few focused steps:
1. Document and test the physical-table visibility contract.
    - #2576 
    - #2577 
2. Fix the request-scoped analyzer state that currently prevents safely removing the lock.
    - #2619 
4. Remove `call_lock`, then validate same-context concurrency and throughput.
    - #2666 

I’ll share the related PRs here as they’re ready.

[...]

PR #2510, opened 2026-07-15
perf(core-py): share one Tokio runtime across session contexts

ttw225:
## Problem

Each context previously added about 18 runtime worker threads, producing
an approximately `1 + 18 × contexts` native-thread count on this machine.
With the shared runtime, the count stays flat:

Measured on an 18-core macOS machine:

| contexts | before | after |
|---:|---:|---:|
| 1 | 19 | 19 |
| 4 | 73 | 19 |
| 8 | 145 | 19 |

## Changes

- Share one process-wide Tokio runtime across all session contexts.
- Rebuild the runtime lazily when first used in a forked child.
- Preserve runtime creation errors as Python exceptions.
- Add a POSIX-only fork regression test covering contexts created both
  before and after the fork.
- Add a parent-side timeout so deadlock regressions fail instead of
  hanging CI.

A forked child inherits the runtime handle but not its worker threads.
The inherited handle is deliberately leaked because dropping it may
wait for workers that were not copied by `fork()`.

The Python API is unchanged. Runtime usage outside `wren-core-py` is
not affected.

## Related work

#2485 releases the GIL around Rust execution while retaining the
per-context `call_lock`.

#2495 fixes a separate catalog-list race exposed by concurrent
transforms. It is now merged and complements this PR: #2495 isolates
catalog state, while this PR bounds Tokio worker growth.

#2504 tracks the remaining Python-side catalog documentation,
integration coverage, and eventual `call_lock` removal. This PR does
not change catalog semantics or close that issue.

## Verification

- `cargo test` — 37 passed
- `cargo test --no-default-features` — 37 passed
- release extension built successfully with `maturin`
- Python test suite — 35 passed
- concurrent-call and fork tests passed across 10 repeated rounds
- native thread count remained flat at 19 for up to 8 contexts

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session reliability after process forking by ensuring the async runtime state is correctly handled for the child process.
  * Confirmed SQL planning/execution paths remain functional before and after a fork, avoiding potential stale-state hangs.

* **Tests**
  * Added a regression test that validates session behavior across `fork()` on POSIX systems, including deadline-based polling, timeout handling, and exit-status reporting.

### DIFF
--- a/core/wren-core-py/src/context.rs
+++ b/core/wren-core-py/src/context.rs
@@ -47,6 +47,29 @@ use wren_core::{
 };
 use wren_core_base::mdl::DataSource;
 
+/// Process-wide Tokio runtime, recreated lazily after a fork.
+///
+/// A child inherits this handle but not the runtime's worker threads, so a PID
+/// mismatch installs a new runtime. The inherited handle is leaked because
+/// dropping it may wait on nonexistent workers. The mutex is released before
+/// `block_on`; fork safety requires it to be unlocked when `fork()` occurs.
+/// Runtime creation failures are returned as Python errors.
+static RUNTIME: Mutex<Option<(u32, Arc<Runtime>)>> = Mutex::new(None);
+
+fn shared_runtime() -> PyResult<Arc<Runtime>> {
+    let mut guard = RUNTIME.lock().unwrap_or_else(PoisonError::into_inner);
+    match guard.as_ref() {
+        Some((pid, rt)) if *pid == std::process::id() => Ok(Arc::clone(rt)),
+        _ => {
+            let rt = Arc::new(Runtime::new().map_err(CoreError::from)?);
+            if let Some(stale) = guard.replace((std::process::id(), Arc::clone(&rt))) {
+                std::mem::forget(stale);
+            }
+            Ok(rt)
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


ttw225:
> 🧹 Nitpick comments (1)
> > core/wren-core-py/src/context.rs (1)> `130-152`: _📐 Maintainability & Code Quality_ | _🔵 Trivial_ | _💤 Low value_
> > > **Defer fetching the shared runtime.**
> > > The `runtime` variable is fetched at the beginning of the `new` method but remains unused in the `None` branch (lines 132-152) since `register_function_by_data_source` fetches its own runtime reference internally. To slightly reduce initialization overhead, consider deferring the `shared_runtime()?` fetch to where it is actually needed inside the `Ok` branch of `AnalyzedWrenMDL::analyze`.

Skipped. Although the local `runtime` binding is only used by the analyzed-MDL path, the no-manifest path may still require the shared runtime through `register_function_by_data_source`.

Keeping the eager lookup also preserves fail-fast runtime creation errors during Python context construction. Deferring it would only avoid initialization for a subset of paths, such as BigQuery, while changing when those errors are reported.

goldmedal (approved):
Thanks @ttw225, nice catch 👍 , thanks for working on the concurrent case.

PR #2513, opened 2026-07-16
perf(wren): defer memory model initialization

ttw225:
## Summary

- Lazily initialize the embedding model only when vector operations require it.
- Reuse the same model instance process-wide for matching model configurations.
- Make model and vector-dimension initialization thread-safe.
- Prepare embeddings before destructive query and seed writes.
- Keep existing public APIs, table schemas, and search behavior unchanged.

## Performance

Measured with five fresh CLI processes using the same environment and populated LanceDB store; values are medians.

| Command | `main` | This PR | Improvement |
|---|---:|---:|---:|
| `wren memory status` | 8.690s | 0.592s | 14.7× faster |
| `wren memory recall` | 7.140s | 3.415s | 2.09× faster |

Non-vector operations such as status and full-context retrieval no longer load the embedding model. Vector operations still initialize it on first use, then reuse the cached instance.

## Implementation

- Separate lazy initialization for the embedding function and vector dimension.
- Process-wide, thread-safe single-slot cache keyed by model name, device, and `trust_remote_code`.
- Resolve vector dimensions from existing table schemas when possible.
- Prefer locally cached model files, with an online-capable fallback when unavailable.
- Validate vector dimensions across query and seed tables before writing.
- Compute embeddings before overwrite, upsert, and seed replacement operations.

### Dependency compatibility

The memory extra now requires `sentence-transformers>=3.0.0`.
This is the first supported release that provides the
`local_files_only` constructor argument used by the local-first model
loading path. The test fake also enforces the constructor keyword
contract so unsupported argument drift is detected.

## Verification

- `just lint`
- `just test-memory`: 96 passed
- `just test-memory-slow`: 1 passed
- `just test-unit`: 585 passed, 3 skipped

## Known limitations

- Database delete-and-write operations remain non-transactional.
- The process-wide model cache holds one model configuration at a time.
- Model identity is not stored in table metadata, so different models with the same vector dimension cannot be distinguished.

## Summary by CodeRabbit

* **New Features**
  * Embedding models now prefer local loading, with automatic online fallback.
  * Memory query handling supports more efficient bulk import and seed updates.
* **Bug Fixes**
  * Improved safeguards around vector dimensionality mismatches.
  * Preserves existing memory data when embedding or rebuild/upsert operations fail.
  * Markdown exports continue to skip seed queries.
* **Tests / CI**
  * Added a test split for slow vs fast runs (including dedicated slow memory tests).
* **Chores**
  * Raised the minimum required `sentence-transformers` version for the memory feature.


goldmedal (approved):
Thanks @ttw225 👍

PR #2514, opened 2026-07-16
fix(memory): skip non-dict columns in seed query generation

Bartok9:
## Summary

- `_model_seeds` skips non-dict or nameless column entries instead of TypeErroring on `col["name"]`.
- Apache-2.0 path: `core/wren/**`.

## Motivation

Partial/hand-edited MDL column arrays can include nulls or unshaped rows. Seed generation is pure and runs during indexing; a single bad column listed should not kill the model.

## Verification

- `cd core/wren && .venv/bin/python -m pytest tests/unit/test_seed_queries.py -v` — **39 passed**

## Real behavior proof

`test_skips_non_dict_columns` PASSED (null/string/missing-name columns ignored; SUM(amount) still generated).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when processing incomplete or malformed model column metadata.
  * Seed query generation now safely ignores invalid column entries while continuing to generate listings and aggregations for valid columns.
* **Tests**
  * Added unit coverage to ensure seed queries are produced correctly even when a model’s columns list contains non-dictionary or unnamed entries.

### DIFF
--- a/core/wren/src/wren/memory/seed_queries.py
+++ b/core/wren/src/wren/memory/seed_queries.py
@@ -85,7 +85,9 @@ def _model_seeds(
         is_pk = norm_name in primary_keys
         # Identifiers are numeric by storage but not measures: summing a join
         # key (e.g. SUM(customer_id)) is semantically meaningless.
-        is_identifier = is_pk or norm_name in relationship_keys or _is_id_like(str(col_name))
+        is_identifier = (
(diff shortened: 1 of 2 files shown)


Bartok9:
Thanks @coderabbitai — pre-filtered columns into a single `valid_columns` list used by both loops, removing the duplicated `isinstance`/name checks. 39 unit tests pass.

goldmedal (approved):
Thanks @Bartok9 👍

PR #2517, opened 2026-07-16
fix(connector): coerce non-string SQL in strip_trailing_semicolon

Bartok9:
## Summary

- `strip_trailing_semicolon` returns `""` for non-string inputs instead of raising TypeError in the re engine.
- Apache-2.0: `core/wren/**`.

## Motivation

Call sites and connector wrappers occasionally pass None placeholders before SQL is fully formed. Shared strip helper should be safe for all connectors that fork through base.

## Verification

- `cd core/wren && .venv/bin/python -m pytest tests/unit/test_strip_trailing_semicolon_non_string.py -v` — **3 passed**

## Real behavior proof

none/bytes → empty; string strip still works PASSED.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQL input handling to safely process non-text values without raising errors.
  * Continued removing trailing semicolons and whitespace from valid SQL text.

* **Tests**
  * Added coverage for `None`, byte values, and standard SQL strings.

### DIFF
--- a/core/wren/src/wren/connector/base.py
+++ b/core/wren/src/wren/connector/base.py
@@ -16,7 +16,12 @@ def strip_trailing_semicolon(sql: str) -> str:
     or ``EXPLAIN SELECT 1;``). Only the *terminating* run of semicolons and
     whitespace is removed, so semicolons inside string literals
     (``SELECT 'a;b'``) are preserved.
+
+    Non-string inputs (None/bytes) coerce to ``""`` so call sites that pass
+    placeholders do not raise TypeError inside the re engine.
     """
+    if not isinstance(sql, str):
+        return ""
     return _TRAILING_SEMICOLONS_RE.sub("", sql)
 
 
--- a/core/wren/tests/unit/test_strip_trailing_semicolon_non_string.py
+++ b/core/wren/tests/unit/test_strip_trailing_semicolon_non_string.py
@@ -1,0 +1,15 @@
+"""strip_trailing_semicolon must not TypeError on non-string SQL."""
+
+from wren.connector.base import strip_trailing_semicolon
+
+
+def test_none_returns_empty():
+    assert strip_trailing_semicolon(None) == ""  # type: ignore[arg-type]
+
+
+def test_bytes_returns_empty():
+    assert strip_trailing_semicolon(b"SELECT 1;") == ""  # type: ignore[arg-type]
+
+
+def test_string_still_strips():
+    assert strip_trailing_semicolon("SELECT 1;") == "SELECT 1"


goldmedal:
## Review

Thanks for the PR. This is a small change, but I think the direction is wrong: it converts a fail-fast error into a silent downstream failure that is harder to debug. I'd suggest **requesting changes**.

### What the change does

`strip_trailing_semicolon(sql)` currently raises `TypeError` on `None`/`bytes` (the compiled pattern is a `str` regex, so it can't operate on `None` or `bytes`). This PR adds `if not isinstance(sql, str): return ""` so those inputs return `""`.

### Main concerns (blocking)

**1. This masks the bug rather than fixing it.** Every call site wraps the return value into a larger statement, e.g.:

```python
f"SELECT * FROM ({strip_trailing_semicolon(sql)}) AS _sub LIMIT {limit}"
```

With `sql=None`, the old behavior raised `TypeError` right at the helper — at the actual source of the bug (something passed `None`). After this change, `stripped == ""` produces invalid SQL like `SELECT * FROM () AS _sub LIMIT 100`, which only fails once it reaches the real database, gets caught, and is re-raised as `WrenError(INVALID_SQL)` with `DIALECT_SQL == None` in the metadata. We've traded a clear `TypeError` at the boundary for a remote, confusing `INVALID_SQL` whose original SQL is `None`. None of the assigning call sites (duckdb, clickhouse, etc.) guard against an empty stripped string, so this is a net negative for debuggability.

**2. The motivation isn't substantiated.** The description says call sites "occasionally pass None placeholders before SQL is fully formed," but there's no such call site in the diff or anywhere in `core/wren`: `query(self, sql: str, ...)` and `dry_run(self, sql: str)` are typed `str`, and every caller passes a real query string. If a `None`-passing path really exists, the fix belongs at that call site (or in the flow/type that produces it), not swallowed in the shared helper. Could you share an actual stack trace / repro where `None` or `bytes` reaches this function?

**3. `bytes → ""` is lossy, and the test locks it in.** `b"SELECT 1;"` returns `""`, silently discarding a query that is otherwise meaningful. If `bytes` SQL is a legitimate input, the right coercion is `sql.decode()`, not emptying it. If it isn't legitimate, it shouldn't be special-cased at all. As written, `test_bytes_returns_empty` turns "drop the query" into asserted behavior.

**4. Signature vs. behavior mismatch.** The signature is `sql: str`, yet the tests need `# type: ignore[arg-type]` to pass non-strings — a sign the runtime behavior contradicts the declared contract. Better to keep the contract strict.

### Suggestions

- If the goal is to reject bad input earlier and more clearly, raise explicitly — `raise TypeError(...)` or a `WrenError(INVALID_SQL, "SQL must be a non-empty string")` — rather than returning `""`.
- If `bytes` input genuinely occurs, `decode()` it instead of emptying it.
- Either way, please back this with evidence of a call site that actually passes `None`/`bytes`; without it, the change guards against a scenario that doesn't happen.

Bartok9:
Thanks for the thorough review, @goldmedal — you're right on all counts, and I appreciate the fail-fast framing.

Agreeing point by point:

1. **Masks rather than fixes.** Returning `""` does turn a clear `TypeError` at the boundary into a remote, confusing `INVALID_SQL` with `DIALECT_SQL == None`. That's a net negative for debuggability.
2. **Motivation unsubstantiated.** I can't point to a real call site that passes `None`/`bytes` — `query`/`dry_run` are typed `str` and every caller passes a real query string. Guarding against a scenario that doesn't demonstrably happen is the wrong tradeoff.
3. **`bytes → ""` is lossy** and the test locks that in as asserted behavior.
4. **Signature vs. behavior mismatch** — needing `# type: ignore[arg-type]` is the tell that the runtime contradicts the declared contract.

Since there's no evidence of a call site that actually reaches this with `None`/`bytes`, I'd rather not weaken the helper's contract speculatively. I'm closing this PR. If a genuine `None`-passing path turns up later, the fix belongs at that call site (or an explicit `raise` at the boundary), not a silent coercion here. Thanks again.

Issue #2518, opened 2026-07-16
Improve WASM SDK CI performance and cache reliability


ttw225:
## Summary

WASM SDK CI currently takes approximately **17–29 minutes** per run.

The observed latency comes from multiple independent causes rather than a
single bottleneck:

- production-oriented WASM optimization runs on every pull request;
- intermittent Cargo cache misses add a cold-compilation penalty;
- repository cache pressure increases the likelihood of those cache misses.

I plan to implement these improvements incrementally through separate,
focused PRs. This issue is intended to track that work, share the overall
context with maintainers and reviewers, and gather feedback on the proposed
direction as the PRs are opened.

## Evidence

Recent `Build WASM + TypeScript SDK` runs:

| Scenario | Cargo compile | wasm-opt | wasm-pack total |
|---|---:|---:|---:|
| [Warm cache](https://github.com/Canner/WrenAI/actions/runs/29254751463) | 1:09 | ~15:13 | 16:45 |
| [Cold cache](https://github.com/Canner/WrenAI/actions/runs/29251947757) | 10:16 | ~18:25 | 29:04 |

The measurements show two separate runtime costs:

- `wasm-opt -O` adds approximately **15–18 minutes** to every optimized build,
  even when the Cargo cache is warm.
- A Cargo cache miss adds approximately **9–10 minutes** of compilation.

The intermittent misses are consistent with repository cache pressure and
eviction. Large PR-scoped Rust caches from other workflows account for most of
the observed cache usage; the WASM workflow is primarily affected by that
pressure rather than being its main source.

Because these problems are largely independent, they can also be addressed
independently.

## Why multiple PRs?

Although the changes all concern CI performance, they address different
problems and have different risk profiles.

Keeping them in separate PRs makes each change easier to review, measure, and
revert independently. It also allows the highest-value improvement to land
without waiting for the broader cache-policy work.

## Tracked work

- [ ] **Clean up caches from closed pull requests**

  Remove caches associated with a pull request's merge ref after the PR closes,
  without checking out or executing PR-controlled code.

  This reduces long-lived repository cache pressure but does not directly
  change build behavior.

  PR: #2519

- [ ] **Add fast pull-request validation**

  Use release-mode `wasm-pack build --release --no-opt` for pull requests. This
  skips the expensive Binaryen optimization pass while retaining release
  compilation, wasm-bindgen generation, type checking, and integration tests.

  Keep fully optimized builds on `main` and all publish paths. Reuse one
[...]

PR #2519, opened 2026-07-16
ci(workflows): delete PR-scoped caches when a pull request closes

ttw225:
## Why

PR-scoped GitHub Actions caches are stored under `refs/pull/<n>/merge` and are
available only to runs of that same PR. After the PR closes, they no longer
support active PR feedback but continue occupying repository cache capacity.
This can crowd out reusable `main`-branch caches and increase cold-compile cache
misses in unrelated workflows.

Snapshot on 2026-07-16: 23 caches using approximately 11.8 GiB. `main` owns only
one reusable seed of approximately 350 MiB; all remaining entries are
PR-scoped, with several PRs retaining approximately 1–2 GiB each.

This is the first of a small, incremental series tracked in #2518 .

## What

- Add `cleanup-pr-caches.yml`, triggered by `pull_request_target: closed`, to
  delete every cache for the closing PR's merge ref:

  ```bash
  gh cache delete --all \
    --ref refs/pull/<n>/merge \
    --succeed-on-no-caches
  ```

- Use `pull_request_target` only to obtain a base-repository token with
  `actions: write`, including for fork PR closures. The job does not check out
  or execute PR-controlled code; the only PR-derived value it consumes is the
  server-assigned PR number.
- Provide a guarded `workflow_dispatch` entry point that refuses to delete
  caches for an open PR.
- Default to `permissions: {}` and grant only job-level `actions: write` and
  `pull-requests: read`, the latter being required by the open-state guard.
- Treat an empty merge ref as a successful no-op while allowing deletion,
  permission, and API failures to fail the job visibly.

## Verification

- `actionlint .github/workflows/cleanup-pr-caches.yml`
- This workflow runs from the base branch, so functional verification is
  available only after it merges to `main`.
- After another PR closes or merges, compare the workflow summary with:

  ```bash
  gh cache list \
    --repo Canner/WrenAI \
    --ref refs/pull/<n>/merge
  ```

- The manual workflow may also be tested with an already-closed PR number. It
  should succeed when the ref is empty and refuse an open PR number.

## Summary by CodeRabbit

* **Chores**
  * Added automated cleanup of GitHub Actions caches when pull requests are closed.
  * Added a manual option to clean caches for a specified pull request.
  * Added safeguards to prevent cache removal while a pull request remains open.
  * Reports cleanup results, including when no matching caches are found.

### DIFF
--- a/.github/workflows/cleanup-pr-caches.yml
+++ b/.github/workflows/cleanup-pr-caches.yml
@@ -1,0 +1,62 @@
+name: Cleanup PR Caches
+
+# Deletes GitHub Actions caches scoped to a pull request's merge ref once the
+# PR closes. PR-scoped caches (including those created by fork PRs) are stored
+# in this repository under refs/pull/<number>/merge and can only be restored
+# by re-runs of that same PR, so after close they only consume repository
+# cache capacity and crowd out reusable main-branch caches.
+#
+# Security: pull_request_target is used solely to obtain a base-repository
+# token with actions: write for fork-initiated closures. The job never checks
+# out or executes PR-controlled code — the only value consumed from the event
+# is the PR number.
+
+on:
+  pull_request_target:
+    types: [closed]
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: "Closed PR number whose merge-ref caches should be deleted"
+        required: true
+        type: number
+
+permissions: {}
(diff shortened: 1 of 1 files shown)


goldmedal (approved):
Thanks @ttw225, make sense to me 👍

PR #2521, opened 2026-07-17
fix(langchain): skip non-dict rows in format helpers

Bartok9:
## Summary

- format_fetch_context_content, format_recall_content, and format_list_models_content assumed every list element was a dict.
- Skip non-dict/null rows; tolerate non-list results/models and non-dict properties so tool envelopes never AttributeError on partial memory/search payloads.

## Motivation

Partial or malicious memory/search payloads can inject non-dict rows. Removing those edges silently is more useful than crashing the LangChain tool content formatter mid-response.

## License

Touches `sdk/**` only (Apache-2.0).

## Verification

```
cd sdk/wren-langchain && python3 -m pytest tests/unit/test_format_malformed.py -v
# 3 passed
```

## Test plan

- [x] Local unit tests
- [ ] CI green

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of search-context, recall, and model-list formatting for malformed or wrong-shaped inputs.
  * Invalid entries are skipped while preserving numbering/counters for valid items.
  * Added safer normalization for non-string summaries/descriptions and non-list columns/properties.
  * Ensures correct fallback messages when no valid content can be produced.
* **Tests**
  * Expanded unit coverage for malformed inputs across fetch-context, recall, and model listing formatters, including normalization and “all invalid” fallback cases.

### DIFF
--- a/sdk/wren-langchain/src/wren_langchain/_format.py
+++ b/sdk/wren-langchain/src/wren_langchain/_format.py
@@ -75,17 +75,27 @@ def format_fetch_context_content(result: dict[str, Any]) -> str:
         return _cap_to_bytes(text, suffix="\n\n...[truncated]")
 
     items = result.get("results", []) or []
+    if not isinstance(items, list):
+        return "_No relevant context items found._"
     if not items:
         return "_No relevant context items found._"
 
     lines = []
-    for i, item in enumerate(items, start=1):
+    n = 0
+    for item in items:
+        if not isinstance(item, dict):
+            continue
+        n += 1
         item_type = item.get("item_type", "item")
         name = item.get("name", "")
         summary = item.get("summary") or item.get("text") or ""
+        if not isinstance(summary, str):
+            summary = str(summary) if summary is not None else ""
         if len(summary) > 120:
             summary = summary[:117] + "..."
-        lines.append(f"{i}. [{item_type}] {name} — {summary}")
+        lines.append(f"{n}. [{item_type}] {name} — {summary}")
+    if not lines:
+        return "_No relevant context items found._"
     return _cap_to_bytes("\n".join(lines), suffix="\n...[truncated]")
 
 
@@ -108,12 +118,20 @@ def format_recall_content(rows: list[dict[str, Any]]) -> str:
     """Render recalled NL→SQL pairs as a numbered list with code fences."""
     if not rows:
         return "_No similar past queries found._"
+    if not isinstance(rows, list):
+        return "_No similar past queries found._"
 
     chunks = []
-    for i, row in enumerate(rows, start=1):
+    n = 0
+    for row in rows:
+        if not isinstance(row, dict):
+            continue
+        n += 1
         nl = row.get("nl_query") or row.get("nl") or ""
         sql = row.get("sql_query") or row.get("sql") or ""
-        chunks.append(f'{i}. "{nl}"\n   ```sql\n   {sql}\n   ```')
+        chunks.append(f'{n}. "{nl}"\n   ```sql\n   {sql}\n   ```')
+    if not chunks:
+        return "_No similar past queries found._"
     return "\n".join(chunks)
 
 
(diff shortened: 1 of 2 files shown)


goldmedal:
## Review

Small, well-targeted defensive fix — scope is `sdk/wren-langchain/` only (`_format.py` + one new test), no API / MDL / connector surface touched. Implementation is correct and introduces no bugs. LGTM to merge; the notes below are mostly test-coverage and consistency polish, none blocking.

### Correctness ✅
The three formatters' `isinstance` filtering, the renumbering via `n` (replacing `enumerate`'s `i`), and the "all-invalid → fallback sentinel" branches are all correct. `format_list_models_content` is the most thorough (also handles non-list `columns` → `col_count=0`, non-dict `properties`, non-str `desc`). No defects found.

### Main note (non-blocking): new branches are under-tested
The 3 new tests only cover "a non-dict element interleaved with a valid dict is skipped." The edges this PR exists to handle aren't exercised:

1. **Container itself not a list** (`results` / `rows` / `models` not a list) → should hit the new `isinstance(items, list)` guards and return the fallback. Currently uncovered.
2. **Entire list invalid** → the `if not lines` / `if not chunks` / `if not any_model` fallback branches. Currently uncovered.
3. **Field normalization** — non-str `summary`/`desc` taking the `str(...)` path, non-list `columns` → `0`. Untested.
4. **Renumbering not locked in** — the `recall` / `fetch_context` tests don't assert the leading `1.`, so the `n`-vs-`i` renumbering (the subtlest part of the change) isn't pinned. Only `list_models` uses an exact string, and that one has no numbering.

A few cheap cases would close this (non-list container → fallback, all-invalid → fallback, and asserting `'1. "List orders"'` in the recall test).

### Minor consistency (agreeing with CodeRabbit's nitpick)
The three functions use three different validation shapes. `format_recall_content` checks `if not rows` **before** `if not isinstance(rows, list)` — the reverse of the other two. Functionally equivalent for JSON types, but checking emptiness first is slightly less safe for a truthiness-ambiguous object. Suggest unifying all three on the `format_list_models_content` form:
```python
if not isinstance(x, list) or not x:
    return "<sentinel>"
```

### Minor style: test import mechanism
The new test loads `_format.py` via `importlib.util.spec_from_file_location` + `parents[2]` path-walking + a synthetic module name, whereas every other unit test in this directory does `from wren_langchain._X import ...`. CI installs `-e ".[dev]"` (deps present, and `__init__` only pulls `WrenToolkit`/exceptions), so a direct import works, matches convention, and won't break if the file moves:
```python
from wren_langchain._format import (
    format_fetch_context_content,
    format_recall_content,
    format_list_models_content,
)
```

### Non-blocking
- **Scope**: this hardens *list elements* only — the outer envelope is still assumed to be a dict (`result.get(...)` / `manifest.get(...)` would still `AttributeError` on a non-dict). Consistent with the stated motivation, but worth saying "protects rows, not the envelope" in the description.
- **Docstring coverage warning** from the pre-merge check is due to the 3 test functions lacking docstrings; repo CI (ruff + pytest) doesn't enforce it, so it won't block merge.
- Confirm `ruff format --check` passes locally before merge — it's the one hard CI gate here.

Bartok9:
Thanks for the thorough review @goldmedal — really appreciate the care. Pushed 45bd89f addressing all the non-blocking notes:

- **Under-tested branches:** added coverage for (1) non-list containers → fallback, (2) entirely-invalid lists → the `if not lines`/`if not chunks`/`if not any_model` fallbacks, (3) field normalization (non-str `summary`/`desc` via `str(...)`, non-list `columns` → `0`), and (4) locked in the renumbering by asserting the leading `1.` in both the recall and fetch_context tests.
- **Consistency:** unified `format_recall_content` and `format_fetch_context_content` on the `format_list_models_content` shape — `if not isinstance(x, list) or not x:` — so the isinstance check comes first everywhere.
- **Test import:** switched to `from wren_langchain._format import ...` to match convention (deps are present via `-e .[dev]`).

`ruff format --check` / `ruff check` pass locally and all 11 tests are green. Left the envelope-level (`.get(...)` on a non-dict) out of scope per the stated motivation — happy to note 'protects rows, not the envelope' in the description if you'd like.

Bartok9:
Pushed 5139f56 covering CodeRabbit's last nitpick — added `test_format_list_models_normalizes_non_dict_properties` (`"properties": "oops"` → falls back to top-level `description`, model still renders). All format-malformed tests green + ruff clean. Thanks again @goldmedal!

goldmedal (approved):
Thanks @Bartok9, nice catch 👍

PR #2522, opened 2026-07-17
fix(pydantic): skip non-dict models in list_models

Bartok9:
## Summary

- _run_list_models assumed every models[] element was a dict with name/columns/properties.
- Skip non-dict rows and nameless models; coerce non-list columns and non-dict properties.

## Motivation

Malformed MDL fragments should not AttributeError wren_list_models mid-comprehension. Apache-2.0: sdk/**.

## Verification

Logic exercised with an in-process fixture reproducing non-dict/name-less rows (ok model kept with column_count 0). Full pydantic-ai env not on this host; unit test added for CI:

- tests/unit/test_tools_runtime.py::test_wren_list_models_skips_non_dict_and_nameless

## Test plan

- [x] Local behavioral proof + regression test
- [ ] CI green

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model listing resilience when metadata is incomplete or malformed.
  * Invalid or unnamed models are now skipped instead of causing failures.
  * Missing or incorrectly formatted fields receive safe default values.

* **Tests**
  * Added coverage for partially invalid model metadata and expected fallback behavior.

### DIFF
--- a/sdk/wren-pydantic/src/wren_pydantic/_tools.py
+++ b/sdk/wren-pydantic/src/wren_pydantic/_tools.py
@@ -169,11 +169,27 @@ def _run_list_models(toolkit: WrenToolkit) -> list[ModelSummary]:
         raise to_model_retry(exc) from exc
 
     models = manifest.get("models") or []
-    return [
-        ModelSummary(
-            name=m["name"],
-            column_count=len(m.get("columns") or []),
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
Thanks @Bartok9

PR #2524, opened 2026-07-17
fix(langchain): coerce None in format_store_content

Bartok9:
## Summary

- format_store_content called .strip() on sql/nl without guarding None.
- Coerce non-str nl/sql; treat non-list tags as zero tags.

## Motivation

Store-tool confirmation paths can pass placeholders. formatting must not TypeError after a successful store.

## License

sdk/** Apache-2.0

## Verification

```
cd sdk/wren-langchain && python3 -m pytest tests/unit/test_format_store_content.py -v
# 3 passed
```

## Test plan

- [x] Local unit tests
- [ ] CI green

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of stored content with unexpected input types.
  * Prevented formatting failures when SQL and natural-language text are missing or not strings.
  * Hardened tag reporting by only counting tags when they are provided as a list.
* **Tests**
  * Added unit tests covering normal formatting, missing values, non-string title/SQL/natural-language inputs, and non-list tags.

### DIFF
--- a/sdk/wren-langchain/src/wren_langchain/_format.py
+++ b/sdk/wren-langchain/src/wren_langchain/_format.py
@@ -128,10 +128,7 @@ def format_store_content(nl: str, sql: str, tags: list[str] | None) -> str:
     sql_preview = sql.strip().split("\n")[0]
     if len(sql_preview) > 80:
         sql_preview = sql_preview[:77] + "..."
-    if tags is None or not isinstance(tags, list):
-        tag_count = 0
-    else:
-        tag_count = len(tags)
+    tag_count = len(tags) if isinstance(tags, list) else 0
     return f'Stored: "{nl}" → {sql_preview} ({tag_count} tags)'
 
 
--- a/sdk/wren-langchain/src/wren_langchain/_format.py
+++ b/sdk/wren-langchain/src/wren_langchain/_format.py
@@ -119,10 +119,19 @@ def format_recall_content(rows: list[dict[str, Any]]) -> str:
 
 def format_store_content(nl: str, sql: str, tags: list[str] | None) -> str:
     """One-liner ``Stored: "<nl>" → <sql preview> (N tags)``."""
+    # Store tool paths occasionally surface None/placeholders before validation.
+    # Coerce so format helpers never TypeError on .strip().
+    if not isinstance(nl, str):
+        nl = "" if nl is None else str(nl)
+    if not isinstance(sql, str):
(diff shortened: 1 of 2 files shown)


goldmedal:
Reviewed — the fix is correct and I'd be happy to see it merged. One test-quality suggestion and a small nit, neither blocking.

**Correctness ✅**
Reproduced on `main`: `format_store_content(None, ...)` raises `AttributeError: 'NoneType' object has no attribute 'strip'`. The motivation holds — a content formatter shouldn't `TypeError` *after* a store has already succeeded — so defensive coercion at this layer is a reasonable choice. The coercion logic is sound (`None → ""`, other non-str → `str(x)`, non-list tags → 0), and the three test cases cover it well.

**🟡 Suggestion — test uses `importlib` file-loading instead of a normal import**

`test_format_store_content.py` loads the module via `importlib.util.spec_from_file_location` with a hardcoded `parents[2]/"src"/"wren_langchain"/"_format.py"` path. Every existing test in `tests/unit/` imports submodules directly, e.g. `test_envelope.py`:

```python
from wren_langchain._envelope import (...)
```

I confirmed `from wren_langchain._format import format_store_content` works in this package (it's how I reproduced the bug). The importlib approach isn't needed and is more fragile — hardcoded relative path, bypasses the package, and would break if `_format.py` ever adds a relative sibling import. Suggest replacing the loader block with:

```python
from wren_langchain._format import format_store_content
```

and using `format_store_content(...)` directly in the three tests.

**🟢 Nit**
In `if tags is None or not isinstance(tags, list):`, the `tags is None or` is redundant — `not isinstance(None, list)` is already `True`, so `if not isinstance(tags, list):` suffices.

Bartok9:
Thanks for the thorough review, @goldmedal — much appreciated. Applied both:

- **Import**: dropped the `importlib` file-loader in favor of `from wren_langchain._format import format_store_content`, matching `test_envelope.py` and the rest of `tests/unit/`. Cleaner and not path-fragile.
- **Nit**: simplified to `tag_count = len(tags) if isinstance(tags, list) else 0` — the `tags is None or` was indeed redundant.

Also added a small case (from CodeRabbit) covering the non-None non-string coercion path (`str(123)`/`str(456)`) so the `str(x)` branch is exercised, not just the `None` branch.

Pushed as e60f254c.

goldmedal (approved):
Thanks @Bartok9 👍

PR #2533, opened 2026-07-18
fix(memory): skip non-dict models/relationships/views in schema_indexer

Bartok9:
## Summary

- `describe_schema` / `extract_schema_items` iterated models, relationships, and views without type checks.
- Null or string rows raised TypeError/KeyError mid-walk and aborted schema prepare entirely.
- Skip non-dict/nameless models; skip non-dict relationships/views; mirror existing cube guards.

## License

Touches `core/**` (Apache-2.0).

## Motivation

Partial/hand-edited MDL JSON can include null list elements. Prefer partial schema text over a hard fail during deploy semantics prep.

## Verification

Direct load of `schema_indexer.py`:
- `describe_schema` with `[None, {name: orders}]` → text contains `orders`
- `extract_schema_items` with mixed null models → dictionary records only

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when processing incomplete or malformed schema definitions.
  * Safely skips invalid models, relationships, and views when entries aren’t dictionaries or are missing required names.
  * Filters model column details to include only properly formed, named columns when generating descriptions and extracted schema items.
* **Tests**
  * Added unit tests to verify malformed schema entries (non-dict sections, null/unnamed items) are ignored and that extracted results never include unnamed records.

### DIFF
--- a/core/wren/src/wren/memory/cli.py
+++ b/core/wren/src/wren/memory/cli.py
@@ -220,7 +220,11 @@ def index(
             pass  # instructions are optional; never fail index because of them
 
     mem_store = _get_store(path)
-    result = mem_store.index_schema(manifest, seed_queries=not no_seed)
+    try:
+        result = mem_store.index_schema(manifest, seed_queries=not no_seed)
+    except ValueError as e:
+        typer.echo(f"Malformed manifest: {e}", err=True)
+        raise typer.Exit(1) from e
     typer.echo(
         f"Indexed {result['schema_items']} schema items"
         + (f", {result['seed_queries']} seed queries" if result["seed_queries"] else "")
--- a/core/wren/src/wren/memory/schema_indexer.py
+++ b/core/wren/src/wren/memory/schema_indexer.py
@@ -50,14 +50,23 @@ def describe_schema(manifest: dict) -> str:
         lines.append(f"Catalog: {catalog}, Schema: {schema}")
         lines.append("")
 
-    for model in manifest.get("models", []):
-        _describe_model(model, lines)
+    models = manifest.get("models") or []
+    if isinstance(models, list):
+        for model in models:
+            if isinstance(model, dict) and model.get("name"):
+                _describe_model(model, lines)
 
-    for rel in manifest.get("relationships", []):
-        _describe_relationship(rel, lines)
+    rels = manifest.get("relationships") or []
+    if isinstance(rels, list):
+        for rel in rels:
+            if isinstance(rel, dict):
+                _describe_relationship(rel, lines)
 
-    for view in manifest.get("views", []):
-        _describe_view(view, lines)
+    views = manifest.get("views") or []
+    if isinstance(views, list):
+        for view in views:
+            if isinstance(view, dict):
+                _describe_view(view, lines)
 
     cubes = manifest.get("cubes", []) or []
     if isinstance(cubes, list):
@@ -228,16 +237,28 @@ def extract_schema_items(manifest: dict) -> list[dict]:
     mdl_h = manifest_hash(manifest)
     items: list[dict] = []
 
-    for model in manifest.get("models", []):
-        items.append(_model_record(model, mdl_h, now))
-        for col in model.get("columns", []):
-            items.append(_column_record(col, model["name"], mdl_h, now))
+    models = manifest.get("models") or []
+    if isinstance(models, list):
+        for model in models:
+            if not isinstance(model, dict) or not model.get("name"):
+                continue
+            items.append(_model_record(model, mdl_h, now))
+            for col in model.get("columns", []) or []:
+                if not isinstance(col, dict) or not col.get("name"):
+                    continue
+                items.append(_column_record(col, model["name"], mdl_h, now))
 
-    for rel in manifest.get("relationships", []):
-        items.append(_relationship_record(rel, mdl_h, now))
+    rels = manifest.get("relationships") or []
+    if isinstance(rels, list):
+        for rel in rels:
+            if isinstance(rel, dict):
+                items.append(_relationship_record(rel, mdl_h, now))
 
-    for view in manifest.get("views", []):
-        items.append(_view_record(view, mdl_h, now))
+    views = manifest.get("views") or []
+    if isinstance(views, list):
+        for view in views:
+            if isinstance(view, dict):
+                items.append(_view_record(view, mdl_h, now))
 
     cubes = manifest.get("cubes", []) or []
     if isinstance(cubes, list):
--- a/core/wren/src/wren/memory/schema_indexer.py
+++ b/core/wren/src/wren/memory/schema_indexer.py
@@ -286,7 +286,11 @@ def extract_schema_items(manifest: dict) -> list[dict]:
 def _model_record(model: dict, mdl_h: str, now: datetime) -> dict:
     name = model["name"]
     cols = model.get("columns", [])
-    col_summaries = ", ".join(f"{c['name']} ({c.get('type', '?')})" for c in cols[:20])
+    col_summaries = ", ".join(
+        f"{c['name']} ({c.get('type', '?')})"
+        for c in cols[:20]
+        if isinstance(c, dict) and c.get("name")
+    )
     pk = model.get("primaryKey") or ""
 
     description = _prop_description(model)
--- a/core/wren/src/wren/memory/schema_indexer.py
+++ b/core/wren/src/wren/memory/schema_indexer.py
@@ -59,13 +59,13 @@ def describe_schema(manifest: dict) -> str:
     rels = manifest.get("relationships") or []
     if isinstance(rels, list):
         for rel in rels:
-            if isinstance(rel, dict):
+            if isinstance(rel, dict) and rel.get("name"):
                 _describe_relationship(rel, lines)
 
     views = manifest.get("views") or []
     if isinstance(views, list):
         for view in views:
-            if isinstance(view, dict):
+            if isinstance(view, dict) and view.get("name"):
                 _describe_view(view, lines)
 
     cubes = manifest.get("cubes", []) or []
@@ -96,10 +96,11 @@ def _describe_model(model: dict, lines: list[str]) -> None:
     if data_scope:
         lines.append(f"  Data scope: {data_scope}")
 
-    cols = model.get("columns", [])
-    if cols:
+    cols = model.get("columns", []) or []
+    described = [c for c in cols if isinstance(c, dict) and c.get("name")]
+    if described:
         lines.append("  Columns:")
-        for col in cols:
+        for col in described:
             _describe_column(col, lines)
     lines.append("")
 
@@ -251,13 +252,13 @@ def extract_schema_items(manifest: dict) -> list[dict]:
     rels = manifest.get("relationships") or []
     if isinstance(rels, list):
         for rel in rels:
-            if isinstance(rel, dict):
+            if isinstance(rel, dict) and rel.get("name"):
                 items.append(_relationship_record(rel, mdl_h, now))
 
     views = manifest.get("views") or []
     if isinstance(views, list):
         for view in views:
-            if isinstance(view, dict):
+            if isinstance(view, dict) and view.get("name"):
                 items.append(_view_record(view, mdl_h, now))
 
     cubes = manifest.get("cubes", []) or []
--- a/core/wren/src/wren/memory/schema_indexer.py
+++ b/core/wren/src/wren/memory/schema_indexer.py
@@ -36,6 +36,27 @@ def manifest_hash(manifest: dict) -> str:
 SCHEMA_DESCRIBE_THRESHOLD = 30_000
 
 
+def _iter_section(manifest: dict, key: str) -> list:
+    """Return ``manifest[key]`` as a list, raising if it is the wrong type.
+
+    A missing or null section is treated as empty (``[]``). A section whose
+    value is present but *not* a list (e.g. a hand-edited manifest that typos
+    ``"models"`` into an object) is a structural error in the manifest, not
+    per-element junk: silently dropping it would wipe the corresponding index
+    and report success. Raising ``ValueError`` surfaces the malformed manifest
+    while keeping this module dependency-free; the CLI already catches such
+    errors and exits non-zero.
+    """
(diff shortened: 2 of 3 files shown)


goldmedal:
## Review — skip non-dict models/relationships/views in schema_indexer

Direction is right and the change is clean, but the guards are **asymmetric and don't fully cover the case this PR claims to fix**. I ran the edge cases against the PR head; `OK` = guarded, `CRASH` = still raises:

```
CRASH TypeError | describe: model with a null column
CRASH KeyError  | describe: column dict without name
CRASH KeyError  | describe: relationship dict without name
CRASH KeyError  | describe: view dict without name
OK              | extract: model with a null column
CRASH KeyError  | extract: relationship dict without name
CRASH KeyError  | extract: view dict without name
```

### 🟠 Should-fix 1 — `describe_schema` still crashes on null/malformed columns
The `describe_schema` path only guards down to the model level. `_describe_model` iterates columns and calls `_describe_column(col)`, which does `name = col["name"]` with no dict/name guard — so `{"name": "orders", "columns": [None, ...]}` still raises `TypeError`. This is exactly the "null list element" the PR motivation calls out, and the `extract_schema_items` path (the `_column_record` loop) already guards it — so the two paths are inconsistent. Suggest mirroring the extract-path guard in the `_describe_model` column loop:

```python
for col in cols:
    if isinstance(col, dict) and col.get("name"):
        _describe_column(col, lines)
```

### 🟠 Should-fix 2 — relationship/view dicts without a `name` still raise KeyError (both paths)
The top-level guard for models is `isinstance(model, dict) and model.get("name")`, but for relationships/views it is only `isinstance(rel, dict)`. Yet `_relationship_record` / `_view_record` / `_describe_relationship` / `_describe_view` all dereference `["name"]`, so a partially hand-edited relationship/view that hasn't been given a name yet still crashes — inconsistent with the model behavior. Suggest aligning rels/views with models (in both `describe` and `extract`):

```python
if isinstance(rel, dict) and rel.get("name"):
    ...
if isinstance(view, dict) and view.get("name"):
    ...
```

### 🟡 Minor — test loads the module in a non-standard way
The new test loads the module standalone via `importlib.util.spec_from_file_location`, but the existing `tests/unit/test_memory.py` imports it normally (`from wren.memory.schema_indexer import ...`), and the module docstring states it has no LanceDB/embedding dependency, so a plain import works. Suggest using the normal import for consistency and so coverage is counted.

### 🟡 Minor — tests don't cover the cases that still crash
The test models all use `"columns": []` (empty), so they never exercise the crashing paths in #1/#2. Adding "describe + null column" and "rel/view without name" cases would lock in the regressions this PR is meant to fix.

### ✅ Nits / positives
- Minimal change that mirrors the existing cube guards (`or []` + `isinstance` filtering) — consistent style.
- `assert None not in names` is a nice negative assertion.
- The earlier `i.get("name")` → `i.get("item_name")` issue is already fixed on the head; no need to revisit.

Bartok9:
Thanks for the thorough review and for running the edge cases — you're right on all counts, the guards were asymmetric. Fixed in a80c203c:

- **Should-fix 1** — `_describe_model` now filters columns with `isinstance(col, dict) and col.get("name")` before `_describe_column`, mirroring the extract path. No more `TypeError`/`KeyError` on null or unnamed columns in describe.
- **Should-fix 2** — relationships/views now require `.get("name")` in **both** `describe_schema` and `extract_schema_items`, aligned with the model-level guard.
- **Minor (test import)** — switched to the normal `from wren.memory.schema_indexer import ...` import so coverage is counted.
- **Minor (missing coverage)** — added regression tests for describe+null/unnamed column and rel/view-without-name in both paths.

Re-ran your matrix locally — all seven cases now report `OK`, and the 5 unit tests pass. Let me know if you'd like anything else tightened.

Bartok9:
Thanks @coderabbitai — good catch. Strengthened `test_describe_schema_skips_non_dict_models_rels_views` to assert the valid records render exactly once (`### Model:`/`### Relationship:`/`### View:` section counts == 1) so the test would fail if a malformed `None`/bare-string entry were rendered as an empty heading, rather than only checking that "None" is absent. Pushed in the latest commit.

goldmedal:
Solid, well-scoped fix — the per-element guards are applied consistently, the tests assert *omission* (`text.count("### Model:") == 1`) rather than just the absence of `"None"`, and putting them in `tests/unit/` instead of `test_memory.py` is right: the latter is excluded from the `unit tests` job and runs behind the `memory` extra (wren-ci.yml:72-78, 212), while these need only stdlib.

Two requests before merge.

## 1. `_model_record` is missing the `or []` its two siblings got

`dict.get(k, default)` returns the default only when the key is *absent*; an explicit `"columns": null` yields `None` and `cols[:20]` raises. Reproduced on `80ee26c1`:

```python
extract_schema_items({"models": [{"name": "m", "columns": None}]})
# TypeError: 'NoneType' object is not subscriptable
```

`or []` is applied at `_describe_model` (L99) and in `extract_schema_items` (L247) but not L289, so `describe_schema` now survives `"columns": null` while `extract_schema_items` still aborts — and `index_schema` (store.py:122) plus both CLI callers (cli.py:223, 642) have no `try`/`except`, so it propagates exactly as the "aborts schema prepare entirely" motivation describes.

Same three lines: `cols[:20]` slices *before* filtering, so nulls among the first 20 consume summary slots — 5 leading nulls + 25 valid columns yields 15 columns in the embedding text instead of 20, a silent recall loss in the vector index. Filter, then slice:

```python
def _model_record(model: dict, mdl_h: str, now: datetime) -> dict:
    name = model["name"]
    cols = [
        c for c in (model.get("columns") or []) if isinstance(c, dict) and c.get("name")
    ]
    col_summaries = ", ".join(f"{c['name']} ({c.get('type', '?')})" for c in cols[:20])
```

Verified this fixes both and keeps all five new tests green. Please also add a `"columns": None` case — the current tests only cover a list *containing* null.

## 2. A non-list section should raise, not be skipped

These two cases are different error classes, and this PR gives them one policy:

- `"models": [None, {...}]` — junk element in a valid list. Skipping and keeping the rest is right, and is what the motivation describes.
- `"models": {...}` — the section's type is wrong. The manifest is structurally not an MDL manifest, and `isinstance(models, list)` currently drops the entire section with no signal.

The second is destructive today. `index_schema` drops the schema table when `items` is empty (store.py:126-129, `replace` defaults to `True`), so:

```
hand-edit target/mdl.json, typo "models" into an object
  → extract_schema_items returns []
  → index_schema drops the existing schema_items table
  → prints "Indexed 0 schema items.", exits 0
```

A typo silently wipes the user's index, returns success, and never mentions the manifest — after which the agent generates SQL against no schema. That's strictly worse than the `TypeError` this PR removes.

There's no upstream structural check to lean on: `validate_manifest` (context.py:1589) is semantic-only (view dry-plan, descriptions) and is reached only from `wren context validate` (context_cli.py:490), not from this path; `_load_manifest` (memory/cli.py:57) only does `json.loads`. Since `build_json` constructs the dict itself, the sole route for a malformed manifest is the hand-edited `--mdl` / `target/mdl.json` read — i.e. `schema_indexer` *is* the boundary here.

Suggest a two-tier policy: raise on a non-list section, skip per-element junk.

```python
def _iter_section(manifest: dict, key: str) -> list:
    section = manifest.get(key) or []
    if not isinstance(section, list):
        raise ValueError(
            f"manifest[{key!r}] must be a list, got {type(section).__name__}"
        )
    return section
```

`ValueError` keeps the module dependency-free, and the CLI already has the shape to catch it — `_load_manifest` catches `json.JSONDecodeError` → `typer.echo` + `Exit(1)`.

The trade-off worth naming: `describe_schema` is called from MCP tools (mcp_server.py:416, 436) and `get_context`, so raising turns a degraded response into a tool error. I'd argue that's correct — an empty schema string makes the agent hallucinate against nothing, whereas an error says "your mdl.json is malformed at `models`."

## Smaller, non-blocking

- **Adjacent, pre-existing:** `_describe_relationship` / `_relationship_record` index `models[0]` guarded only by `len(models) > 0` — `describe_schema({"relationships": [{"name": "r", "models": {"a": 1}}]})` raises `KeyError: 0`.
- **Consistency:** models/relationships/views are now skipped when unnamed, but cubes/measures/dimensions remain `isinstance`-only and still emit `item_name: ""`. The description says "mirror existing cube guards," but this goes beyond them.

Bartok9:
Thanks — both requests done in 4316695f, and thanks for reproducing the exact failure modes.

**1. `_model_record` null-column guard.** Fixed. `_model_record` now filters columns *before* slicing:
```python
cols = [c for c in (model.get("columns") or []) if isinstance(c, dict) and c.get("name")]
col_summaries = ", ".join(f"{c['name']} ({c.get('type', '?')})" for c in cols[:20])
```
This kills the `"columns": null` → `TypeError` (the `get(k, default)`-only-on-absent-key trap you pointed out) and the silent recall loss where leading nulls ate summary slots. Added `test_columns_null_does_not_crash_either_path` and `test_columns_null_slots_do_not_consume_summary_budget` (5 leading nulls + 25 valid cols → asserts exactly 20 `(int)` summaries, `c19` present / `c20` absent).

**2. Non-list section → raise, not skip.** Agreed the silent index-wipe is strictly worse than the `TypeError` this PR removes. Added a two-tier `_iter_section` helper used by both `describe_schema` and `extract_schema_items`:
```python
def _iter_section(manifest: dict, key: str) -> list:
    section = manifest.get(key)
    if section is None:
        return []          # missing / null stays permissive
    if not isinstance(section, list):
        raise ValueError(f"manifest[{key!r}] must be a list, got {type(section).__name__}")
    return section
```
So per-element junk (`[None, {...}]`) is still skipped, but a wrong-typed section (`"models": {...}`) raises `ValueError` — dependency-free. Wired the CLI `index` path to catch it (`typer.echo(..., err=True)` + `Exit(1)`) so a typo'd manifest reports the malformed key instead of dropping the table and exiting 0. Tests: `test_non_list_section_raises_valueerror` (parametrized over models/relationships/views/cubes, both paths) + `test_missing_or_null_section_is_treated_as_empty`.

On the MCP trade-off: I agree raising is correct there too — a hard "your mdl.json is malformed at `models`" beats an empty schema string that makes the agent hallucinate against nothing.

**Non-blocking two:** I left both out to keep this PR tightly scoped as you noted, happy to fold either in here or as a follow-up if you'd prefer:
- `models[0]` `KeyError` in `_describe_relationship`/`_relationship_record` when `models` is a non-list — it's adjacent/pre-existing and arguably wants the same `_iter_section`-style treatment for the rel's inner `models`.
- cubes/measures/dimensions still `isinstance`-only (emitting `item_name: ""`) — aligning those with the unnamed-skip policy is a slightly broader behavior change than "mirror cube guards," so I'd rather do it deliberately.

CI green, all new tests pass locally.

goldmedal:
Verified `4316695f` — both requests are fully addressed. Re-ran the failure modes:

- `extract_schema_items({"models": [{"name": "m", "columns": None}]})` no longer raises, on either path.
- Filter-before-slice confirmed: 5 leading nulls + 25 valid columns → exactly 20 summaries, `c19` present, `c20` absent.
- All four sections raise on a wrong type, both functions: `ValueError: manifest['models'] must be a list, got dict`.
- Missing / `null` / `[]` stay permissive; `extract_schema_items({})` unchanged, so no regression to the existing coverage in `test_memory.py`.

Nice catch on the `_iter_section` boundary — checking `section is None` *before* `isinstance` is stricter than the sketch I posted. My version used `manifest.get(key) or []`, which would have let `"models": {}` fall through the `or []` and be silently dropped, reintroducing the exact bug on the empty-object case. Yours raises there. The parametrized test with `match=section` pinning the key name in the message is the right assertion too.

## Deferred items — filed

Agreed on both, and split the way you suggested:

- **#2591** — cubes / measures / dimensions. Agree this deserves its own change: unnamed cubes are currently indexed with `item_name: ""`, so switching them to the skip policy changes behavior for manifests that index fine today. The issue frames it as "decide the policy, then apply it uniformly" rather than presuming the fix, since that call should be deliberate.
- **#2590** — `models[0]` in `_describe_relationship` / `_relationship_record`. Worth a look sooner than I implied in my last comment: writing it up turned up that the behavior splits three ways by type, and the third is the bad one.

  ```python
  {"models": {"a": 1}}   # KeyError: 0
  {"models": 5}          # TypeError: object of type 'int' has no len()
  {"models": "orders"}   # no error — indexed as  o → r
  ```

  A string is a valid sequence, so `models[0]` / `models[1]` take the first two *characters*: `"orders"` is indexed as a relationship from `o` to `r`, with `model_name: 'o'` written into the vector index. No exception, just wrong join endpoints in the text handed to the LLM as schema context. That's a silent-wrong-data path rather than a crash, which is why I'd rank it above the cube cleanup.

## One follow-up note, not for this PR

Wiring the CLI `index` path to catch `ValueError` is right, and I checked the `watch` path (`_reindex`, cli.py:642) — it's already safe: `poll_once` counts and re-raises (watch.py:126-131) and `watch_loop` catches, sleeps, and continues (watch.py:174-183) without advancing the fingerprint, so a malformed manifest can't kill the watcher and is retried.

But both handlers are bare `except Exception:` with no `as e`, and `on_event` only receives a fixed string, so the watch path prints the generic "Reindex failed; change kept pending, will retry next poll." every interval and never surfaces `manifest['models'] must be a list, got dict`. That's pre-existing `watch.py` plumbing rather than anything you introduced — but this PR is what makes that message worth seeing, and hand-editing `target/mdl.json` under `watch` is exactly where someone hits it. Left unfiled since it's a separate concern from the two above; happy to open one if you'd like it tracked.

LGTM — ready to merge from my side.

goldmedal (approved):
Thanks @Bartok9

Bartok9:
Thanks for the depth here @goldmedal — the filter-before-slice recall-loss catch and the `section is None`-before-`isinstance` note were both genuinely sharpening. Appreciate you splitting the deferred work into #2590 / #2591; the string-as-sequence `models[0]` path (`"orders"` → `o`→`r` silently) is a nasty one, agree it ranks above the cube cleanup. Happy to take #2590 next. And yes — please do file the `watch.py` bare-`except` surfacing note if you'd like it tracked; this PR is exactly what makes that generic message worth replacing.

PR #2536, opened 2026-07-18
chore(databricks): add the test for stripping trailing semicolon before query execute

Bartok9:
## Summary

- Databricks `query()` executed raw SQL including trailing `;`.
- `dry_run` already stripped for subquery wrap.
- Strip once before execute so both limited-fetch and full-fetch paths are consistent.

## License

`core/wren/**` Apache-2.0.

## Verification

```
python3 -m pytest core/wren/tests/unit/test_databricks_strip_query_semicolon.py -q
```

## Summary by CodeRabbit

* **Tests**
  * Added coverage to verify that trailing semicolons are removed from Databricks queries before execution.
  * Confirmed the behavior handles trailing whitespace alongside the semicolon.

### DIFF
--- a/core/wren/tests/unit/test_databricks_strip_query_semicolon.py
+++ b/core/wren/tests/unit/test_databricks_strip_query_semicolon.py
@@ -1,0 +1,23 @@
+"""Databricks query path strips trailing semicolons before execute."""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+
+def test_query_strips_trailing_semicolon():
+    src = (
+        Path(__file__).resolve().parents[2]
+        / "src"
+        / "wren"
+        / "connector"
+        / "databricks.py"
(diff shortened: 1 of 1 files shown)


goldmedal:
@Bartok9 there are some conflict.

Bartok9:
Thanks @goldmedal! Rebased onto latest `main` and resolved the conflict. Since the `query()` semicolon strip already landed on main (via the shared `strip_trailing_semicolon` refactor), this PR now just adds the regression test guarding that the query path keeps applying it. Ready for another look. 🙏

goldmedal:
The test case isn't necessary. Let's close this PR.

Bartok9:
No problem at all, thanks for taking a look @goldmedal! Makes sense to keep the suite lean since the strip is already covered. Appreciate the quick feedback. 🙏

PR #2539, opened 2026-07-19
fix(connector): coerce non-string SQL in strip_trailing_semicolon

Bartok9:
## Summary
- Coerce non-string `sql` arguments in `strip_trailing_semicolon` (`None` → `""`, other values via `str(...)`).
- Prevents `TypeError` inside the trailing-semicolon regex when callers pass accidental non-strings.

## Motivation
Every connector funnels user/limit/EXOLAIN composition through this helper. A `None` or int SQL argument currently raises `TypeError: expected string or bytes-like object` deep in `re.sub`, masking the real caller bug with a generic regex crash.

## License
Apache-2.0 path: `core/wren/src/wren/connector/base.py` (+ unit test).

## Verification
`cd core/wren && .venv/bin/python -m pytest tests/unit/test_strip_trailing_semicolon_coerce.py -q` → 3 passed

## Test plan
- [x] unit tests for string / None / int
- [ ] CI green

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQL statement handling when input is empty, null, or not provided as text.
  * Non-text values are now safely converted to text before trailing semicolons and whitespace are removed.

* **Tests**
  * Added coverage for standard strings, null values, and non-text inputs.

### DIFF
--- a/core/wren/src/wren/connector/base.py
+++ b/core/wren/src/wren/connector/base.py
@@ -16,7 +16,15 @@ def strip_trailing_semicolon(sql: str) -> str:
     or ``EXPLAIN SELECT 1;``). Only the *terminating* run of semicolons and
     whitespace is removed, so semicolons inside string literals
     (``SELECT 'a;b'``) are preserved.
+
+    Non-string values are coerced with ``str(...)`` so accidental ``None``/
+    numeric SQL arguments degrade instead of raising ``TypeError`` inside
+    the regex. Empty input becomes ``""``.
     """
+    if sql is None:
+        return ""
+    if not isinstance(sql, str):
+        sql = str(sql)
     return _TRAILING_SEMICOLONS_RE.sub("", sql)
 
 
--- a/core/wren/tests/unit/test_strip_trailing_semicolon_coerce.py
+++ b/core/wren/tests/unit/test_strip_trailing_semicolon_coerce.py
@@ -1,0 +1,14 @@
+from wren.connector.base import strip_trailing_semicolon
+
+
+def test_strips_string():
+    assert strip_trailing_semicolon("SELECT 1;") == "SELECT 1"
+    assert strip_trailing_semicolon("SELECT 1 ; \n") == "SELECT 1"
+
+
+def test_none_becomes_empty():
+    assert strip_trailing_semicolon(None) == ""
+
+
+def test_non_string_coerced():
+    assert strip_trailing_semicolon(42) == "42"


goldmedal:
Thanks for the patch, and for the earlier hardening work in the formatter helpers (#2521, #2522, #2524) — those sat at a real trust boundary, where the input is model/tool-produced JSON. This one is different, and I'd rather not take it. Closing, with the reasoning below.

## The input here is internally generated, not untrusted

The only production path into `strip_trailing_semicolon` is the `dialect_sql` a connector receives, and that string is produced by `Engine.dry_plan()` (wren-core + sqlglot `generate`) — see `core/wren/src/wren/engine.py:112` and `:128`. It is a `str` by construction. The PR doesn't identify a caller that can pass `None` or an `int`, and there's no reproducer for the `TypeError` it describes.

## The coercion makes the failure harder to diagnose, not easier

If `None` ever did reach the helper, `""` composes into invalid SQL that we then send to the warehouse:

| connector | resulting SQL |
| --- | --- |
| `mysql.py:41` | `"\nLIMIT 10"` |
| `postgres.py:254`, `trino.py:456`, `canner.py:248` | `SELECT * FROM () AS _sub LIMIT 10` |
| `snowflake.py:69` | `SELECT * FROM (\n\n) AS _wren_sub LIMIT 10` |

`42` becomes `SELECT * FROM (42) AS _sub LIMIT 10`. Each of these fails remotely and gets wrapped by `engine.py:117-125` as `GENERIC_USER_ERROR` with `phase=SQL_EXECUTION` — so a planner/caller bug is reported to the user as an error in *their* SQL. The current `TypeError` is ugly, but its traceback points straight at the real caller. The PR's stated goal is to stop masking the caller bug; the effect is to mask it.

## Smaller points

- The signature still reads `sql: str` while the body handles non-strings. Widening it to `Any` (as suggested in the automated review) isn't the fix either — that removes the static check that catches this class of bug at authoring time. Either the value is a `str` and the guard is dead code, or the annotation is wrong; neither resolution argues for merging as-is.
- The guard is also partial: connectors that don't route through this helper (e.g. `bigquery.py:40`) would still raise `TypeError`, so this doesn't establish a system-level invariant. The place to validate opaque input is the boundary — `Engine.query` or the MCP tool arguments (already schema-validated).
- `core/wren/tests/unit/test_strip_trailing_semicolon.py` already exists and covers the string cases via `parametrize`; `test_strips_string` in the new file duplicates two of them. Additional cases belong in that file rather than a second module. (Lint is green because ruff here only selects `E,F,I,PLC` — that isn't a signal about consistency with the surrounding test style.)

## A more useful version of this change

The empty-string path is already reachable today: `strip_trailing_semicolon("  ;  ")` returns `""` and composes exactly the broken SQL in the table above — untested and unguarded. Rejecting empty/whitespace-only SQL early with an explicit error, or raising a clearly-worded `TypeError` instead of letting `re.sub` do it, would serve the motivation in the PR description without silently degrading to invalid SQL. Happy to look at a PR along those lines.

Bartok9:
Completely fair — thanks for the thorough writeup. You're right that the production path is `str` by construction via `dry_plan()`, and coercing to `""` would turn a caller bug into invalid SQL sent to the warehouse (surfacing as a misleading `GENERIC_USER_ERROR`), which defeats the point. The `TypeError` traceback pointing at the real caller is the more useful signal.

The empty/whitespace-only observation is the interesting one — `strip_trailing_semicolon("  ;  ")` → `""` already composes the same broken SQL today, unguarded. Happy to follow up with a small PR that rejects empty/whitespace-only SQL early with an explicit error at the boundary, rather than widening this helper. Appreciate the guidance.

PR #2540, opened 2026-07-19
fix(memory): skip non-dict rows in describe_schema

Bartok9:
## Summary
- `describe_schema` skips non-dict models/relationships/views and non-dict columns.
- Mirrors defensive guards already used for cubes measures/dimensions.

## License
Apache-2.0 (`core/wren/...`).

## Verification
`cd core/wren && .venv/bin/python -m pytest tests/unit/test_schema_describe_nonduct.py -q`

## Test plan
- [x] unit test mixed malformed manifest
- [ ] CI green

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved schema description generation by safely handling missing or malformed manifest entries for models, relationships, and views.
  * Made model column rendering conditional so only valid, named columns appear.
  * Omitted empty “Columns” sections from the output.

* **Tests**
  * Added a unit test to verify schema descriptions include valid items (models/views) and skip invalid non-conforming entries.

### DIFF
--- a/core/wren/src/wren/memory/schema_indexer.py
+++ b/core/wren/src/wren/memory/schema_indexer.py
@@ -50,14 +50,23 @@ def describe_schema(manifest: dict) -> str:
         lines.append(f"Catalog: {catalog}, Schema: {schema}")
         lines.append("")
 
-    for model in manifest.get("models", []):
-        _describe_model(model, lines)
+    models = manifest.get("models", []) or []
+    if isinstance(models, list):
+        for model in models:
+            if isinstance(model, dict):
+                _describe_model(model, lines)
 
-    for rel in manifest.get("relationships", []):
-        _describe_relationship(rel, lines)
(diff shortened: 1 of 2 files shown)


goldmedal:
Closing in favor of #2586, which supersedes this.

#2586 covers the same `describe_schema` non-dict guards, plus `extract_schema_items`, and factors the repeated `x or []` / `isinstance(..., list)` dance into one `_as_list` helper — including the pre-existing cube guards, so the file ends up with less duplication rather than more. Three open PRs against `schema_indexer.py` (this one, #2541, #2586) also guarantee rebase conflicts for whichever lands second.

Review effort is better spent on #2586; nothing here is lost.

Bartok9:
Makes sense — #2586 consolidates these guards through one `_as_list` helper with a stronger test. Closing in its favor. Thanks!

PR #2541, opened 2026-07-19
fix(memory): skip non-dict models/rels/views in extract_schema_items

Bartok9:
## Summary
- `extract_schema_items` ignores non-dict models/relationships/views and nameless/non-dict columns.
- `_model_record` column summary only joins dict columns with names.

## License
Apache-2.0 path under `core/wren`.

## Verification
`cd core/wren && .venv/bin/python -m pytest tests/unit/test_schema_extract_nonduct_models.py -q`

## Test plan
- [x] unit test malformed manifest
- [ ] CI green

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved schema extraction resilience when manifests include malformed or incomplete models, columns, relationships, and views.
  * Safely skips invalid non-conforming entries to prevent extraction and summary failures.
  * Prevents malformed column data from impacting generated model summaries.
* **Tests**
  * Added unit test coverage using mixed valid and invalid schema inputs to confirm invalid entries are ignored while valid items are extracted correctly.

### DIFF
--- a/core/wren/src/wren/memory/schema_indexer.py
+++ b/core/wren/src/wren/memory/schema_indexer.py
@@ -244,13 +244,13 @@ def extract_schema_items(manifest: dict) -> list[dict]:
     rels = manifest.get("relationships", []) or []
     if isinstance(rels, list):
         for rel in rels:
-            if isinstance(rel, dict):
+            if isinstance(rel, dict) and rel.get("name") is not None:
                 items.append(_relationship_record(rel, mdl_h, now))
 
     views = manifest.get("views", []) or []
     if isinstance(views, list):
         for view in views:
-            if isinstance(view, dict):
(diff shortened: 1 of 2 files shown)


goldmedal:
Closing in favor of #2586, which supersedes this.

#2586 covers the same `extract_schema_items` non-dict guards, plus `describe_schema`, and routes all of them through one `_as_list` helper instead of repeating the `x or []` + `isinstance(..., list)` pattern at each call site. Its test also pins the exact surviving-item count, which is a stronger assertion than "does not raise". Three open PRs against `schema_indexer.py` (this one, #2540, #2586) would otherwise conflict with each other.

Nothing here is lost — it's all in #2586.

Bartok9:
Agreed — #2586 supersedes this with the shared `_as_list` helper and a tighter assertion. Closing in its favor. Thanks!

PR #2535, opened 2026-07-18
fix(athena): strip trailing semicolon on unlimited query path

Bartok9:
## Summary

- Athena only stripped `;` when wrapping with LIMIT.
- Unlimited path forwarded terminators to `cursor.execute`.
- Strip once; reuse stripped SQL for both branches.

## License

`core/wren/**` Apache-2.0.

## Verification

```
python3 -m pytest core/wren/tests/unit/test_athena_strip_unlimited_query.py -q
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Athena SQL execution by stripping any trailing semicolon before running queries.
  * Ensured the limited-query SQL wrapper isn’t affected by trailing line comments.

* **Tests**
  * Added a unit test to confirm the unlimited Athena query path removes trailing semicolons before execution.

PR #2542, opened 2026-07-19
fix(memory): skip non-dict models/columns in seed query generation

Bartok9:
## Summary
- `generate_seed_queries` skips non-dict models/relationships.
- `_model_seeds` filters non-dict / nameless columns before reading `col["name"]`.

## License
Apache-2.0 (`core/wren`).

## Verification
`cd core/wren && .venv/bin/python -m pytest tests/unit/test_seed_queries_nonduct.py -q`

## Test plan
- [x] unit test mixed manifest
- [ ] CI green

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved seed query generation to safely handle manifests with missing or unexpected structures, including malformed models, columns, and relationships.
  * Hardened column and relationship parsing to skip invalid entries and avoid malformed joins/conditions.
  * Updated Redshift “unlimited” query execution to strip trailing semicolons using the shared helper.
* **Tests**
  * Added unit coverage to verify invalid non-dict/malformed model and column definitions are skipped while still producing expected SQL output.

### DIFF
--- a/core/wren/src/wren/connector/redshift.py
+++ b/core/wren/src/wren/connector/redshift.py
@@ -52,7 +52,7 @@ class RedshiftConnector(ConnectorABC):
         else:
             # Unlimited path also rejects trailing ``;`` for single statements
             # depending on driver/session settings — strip for consistency.
-            sql = _strip_trailing_semicolon(sql)
+            sql = strip_trailing_semicolon(sql)
         with closing(self.connection.cursor()) as cursor:
             cursor.execute(sql)
             cols = [desc[0] for desc in cursor.description]
--- a/core/wren/src/wren/memory/seed_queries.py
+++ b/core/wren/src/wren/memory/seed_queries.py
@@ -71,7 +71,9 @@ def _model_seeds(
     columns = model.get("columns", []) or []
     if not isinstance(columns, list):
         columns = []
-    columns = [c for c in columns if isinstance(c, dict) and c.get("name") is not None]
+    columns = [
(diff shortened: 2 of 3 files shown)


goldmedal:
Closing this one — it needs to be re-cut against current `main`, and it overlaps another open PR.

**The Redshift hunk doesn't match `main`.** The diff rewrites `_strip_trailing_semicolon(sql)` → `strip_trailing_semicolon(sql)` inside an `else:` branch of `query()`. On `main` today, `RedshiftConnector.query()` has no `else` branch and no unlimited-path strip at all — it only strips inside the `limit is not None` wrap (`redshift.py:49`) and in `dry_run` (`redshift.py:62`), and the private alias is already gone (that part landed in #2560). So this hunk is based on a tree state that no longer exists.

**The `seed_queries.py` hunk duplicates #2584.** Both add near-identical non-dict/non-str guards to `generate_seed_queries` and `_relationship_key_columns`.

If you still want these, please split them: one PR that adds the unlimited-path strip to Redshift (with a mock-cursor test in the style of `core/wren/tests/unit/test_athena_strip_unlimited_query.py`, added with #2535 — that gap is real and worth fixing), and let the `seed_queries.py` guards land through a single PR rather than two.

Bartok9:
Thanks for the detailed read. Agreed on both counts — the Redshift hunk is stale vs current `main` (post-#2560) and the seed_queries guards overlap #2584. Closing this; I'll cut a focused PR for the real gap: the unlimited-path `strip_trailing_semicolon` in Redshift with a mock-cursor test in the style of `test_athena_strip_unlimited_query.py`.

PR #2547, opened 2026-07-20
fix(mcp): skip non-dict models/columns/rels in list and describe

Bartok9:
## Summary

- MCP `list_models` / `describe_model` called `.get` on every model/column/relationship without type checks.
- Non-dict rows in context load output crash tool handlers with `AttributeError`.
- Skip non-dict models/columns/relationships; treat non-list `columns` as empty.

## Motivation

Defensive consistency with memory/schema_indexer guards on Apache-2.0 `core/**`. MCP tools should return partial valid inventories rather than fail closed on one corrupt row.

## Verification

```text
$ cd core/wren && .venv/bin/python -m pytest tests/unit/test_mcp_model_guards.py -q
2 passed
```

(Logic mirror of the mcp_server sequence which is nested inside factory registration.)

## Apache-2.0

Touches only `core/wren/**`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved “List Models” and “Describe Model” handling for unexpected or incomplete model, column, and relationship data.
  * Non-conforming entries are now ignored instead of causing failures.
  * Column counts and column listings now remain accurate when column data is missing or malformed.
* **Tests**
  * Added unit coverage to ensure junk inputs are skipped for both model listing and model description, including malformed column entries.

### DIFF
--- a/core/wren/src/wren/mcp_server.py
+++ b/core/wren/src/wren/mcp_server.py
@@ -228,14 +228,19 @@ def _register_context_tools(mcp: FastMCP, ctx: ServeContext) -> None:
         models = load_models(ctx.project)
         result = []
         for model in models:
+            if not isinstance(model, dict):
+                continue
             description = model.get("description")
             if description is None:
                 description = (model.get("properties") or {}).get("description")
+            columns = model.get("columns") or []
+            if not isinstance(columns, list):
+                columns = []
             result.append(
                 {
                     "name": model.get("name"),
                     "description": description,
-                    "column_count": len(model.get("columns", []) or []),
+                    "column_count": len(columns),
                 }
             )
         return {"models": result}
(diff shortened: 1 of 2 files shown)


goldmedal:
Thanks for this — I'd like to pass on it, and suggest closing. Not because the underlying concern is wrong, but because of **which layer** it's fixed at, and because the reachable part is already covered by two other open PRs of yours.

**The `model` guard is dead code.** `load_models` dispatches to `_load_models_v1` / `_load_models_v2` and both already filter their items (`context.py:544`, `context.py:568`), so `isinstance(model, dict)` can't be false in either `list_models` or `describe_model`. Same conclusion as the models half of #2567.

**What genuinely leaks is a loader bug, not a consumer bug.** `load_relationships` (`context.py:693-699`):

```python
data = yaml.safe_load(...) or {}
return data.get("relationships", []) if isinstance(data, dict) else []   # outer dict checked, items not
```

That's exactly the `_load_views_v1` shape from #2567 / #2597 — the annotation promises `list[dict]` while items go unfiltered. And for the same reason as there, the guard belongs in the loader: `load_relationships` is read by `validate_project`, `build_manifest` / `build_json` and the MCP tools, so hardening `describe_model` alone leaves the other consumers exposed and implies re-validation is each caller's job.

**And that work is already in flight — yours.** #2607 puts the fix in `context.py` (the right layer), and #2608 rewrites the very same relationship loop in `describe_model` that this PR touches, so the two will conflict. #2547 is the redundant one of the three.

One note in case any of this comes back as a follow-up: the `columns` container check is legitimate (it comes straight from YAML), but it's only applied in `list_models` — `describe_model` still does `for col in model.get("columns") or []`, so `columns: 5` is still a `TypeError` there. If it's worth guarding, it should be consistent across both, along with `properties` (CodeRabbit's open comment) and `rel["models"]`, where a non-list is worse than a crash: `"orders" in "orders_extra"` is `True`, so a relationship that doesn't exist gets returned silently.

Also, `tests/unit/test_mcp_model_guards.py` re-implements the handler logic as local functions instead of importing it, so it would still pass if `mcp_server.py` were reverted. The factory-registration concern in the description is already solved in this directory — `tests/unit/test_mcp_server.py:21` has `_get_tool(mcp, name)`, and `recall_queries` / `get_context` / `run_sql` / `query_cube` are all tested through `build_server(ctx)` that way (as does your own #2608). A fixture project plus `_get_tool(build_server(ctx), "describe_model")` is the pattern to follow.

So: closing in favour of #2607 for the loader fix. Thanks for splitting these out.

goldmedal:
Correction to my comment above: I called #2607 "the loader fix" — that's wrong. #2607 changes `validate_project`, which is another consumer; **no** open PR touches `load_relationships` itself, so the loader's missing item filter is still unaddressed.

That doesn't change the recommendation, but it does change the reason. Accurately:

- #2608 supersedes this PR on the MCP side — same `describe_model` relationship loop, and its version is strictly better: it checks the container type (`isinstance(rel_models, list)`), which is what actually stops the `"orders" in "orders_extra"` substring match, and its test goes through `build_server` + `_get_tool`.
- #2607 is correct as written for what it does: `validate_project`'s job *is* to report structural problems, so raising a `ValidationError` there is the right behaviour, not something the loader should silently absorb.
- Still open, and not covered by any of the four: `load_relationships` (`context.py:693-699`) not filtering its items, with `list[dict]` in the annotation.

So: close this one in favour of #2608, and the `load_relationships` item filter is worth a separate focused PR if you want it.

Bartok9:
Thanks for the careful trace, and for the correction re: #2607 vs #2608 — that's a better map of the layers than mine. Agreed: #2608 is the strictly-better fix on the MCP side (container-type check + test through `build_server`), so closing this in its favour. The `load_relationships` item filter (`context.py:693-699`) is the genuinely-uncovered piece; happy to send that as a separate focused PR if it's wanted. Closing here.

PR #2548, opened 2026-07-20
fix(genbi): guard non-dict models/columns in inventory format

Bartok9:
## Summary

- GenBI `_format_model_inventory` iterated models/columns with unconditional `.get`.
- Non-dict MDL rows crash compose_build_instruction markdown assembly.
- Skip junk models/columns; empty inventory after filtering uses the existing empty message.

## Verification

```text
$ cd core/wren && .venv/bin/python -m pytest tests/unit/test_genbi_format_model_inventory.py -q
3 passed
```

Apache-2.0 path: `core/wren/**`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model inventory formatting to be more tolerant of missing or malformed model data.
  * Invalid/non-dictionary entries and invalid column values are now ignored while preserving valid model and column names.
  * Correctly displays the “no models found” fallback when no valid models are available.
* **Tests**
  * Added unit test coverage for formatting valid inventories, mixed invalid inputs, and empty/invalid-only inputs.

### DIFF
--- a/core/wren/src/wren/genbi/composer.py
+++ b/core/wren/src/wren/genbi/composer.py
@@ -51,14 +51,23 @@ def _data_mode_guidance(data_mode: str) -> str:
     raise ValueError(f"unknown data-mode {data_mode!r}; expected one of {DATA_MODES}")
 
 
-def _format_model_inventory(models: list[dict]) -> str:
+def _format_model_inventory(models: list[dict] | None) -> str:
     """One markdown bullet per model with its column names."""
     if not models:
         return "- (no models found — run `wren context build` first)"
     lines = []
     for model in models:
-        cols = ", ".join(c.get("name", "?") for c in model.get("columns", []))
+        if not isinstance(model, dict):
+            continue
+        raw_cols = model.get("columns") or []
+        if not isinstance(raw_cols, list):
+            raw_cols = []
+        cols = ", ".join(
+            c.get("name", "?") for c in raw_cols if isinstance(c, dict)
+        )
         lines.append(f"- **{model.get('name', '?')}**: {cols}")
+    if not lines:
+        return "- (no models found — run `wren context build` first)"
     return "\n".join(lines)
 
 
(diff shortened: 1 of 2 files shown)


goldmedal:
Suggesting we close this one too — it's the same split as #2547 and the models half of #2567.

**The model guard is dead code.** `_format_model_inventory` has exactly one production caller: `compose_build_instruction`, and the only place that is called from is `genbi/cli.py:143`:

```python
models=load_models(project_path),
```

`load_models` dispatches to `_load_models_v1` / `_load_models_v2`, and both already filter their items (`context.py:544`, `context.py:568`). So `if not isinstance(model, dict): continue` can't be false, and neither can the all-junk `[None, "x", 1]` case that `test_all_junk_falls_back_to_empty_message` covers — that input can't be produced by the only path that reaches this function.

**One line here is real.** `columns` comes straight out of YAML and nothing upstream constrains its type, so `raw_cols` being a non-list is genuinely reachable and `", ".join(...)` over it would break. That's worth having — but it's a two-line change to `cols`, not the surrounding scaffolding.

**The test doesn't go through a reachable path.** It calls the private `_format_model_inventory` directly, which is what let the unreachable inputs look meaningful. A test that goes through `compose_build_instruction` on a project fixture whose `metadata.yml` has `columns: 5` would cover the real case and would have shown that the rest can't happen.

So: close, and if you want the reachable half, a focused PR guarding the `columns` container type — tested through `compose_build_instruction` — would be welcome.

Bartok9:
Agreed — the model guard is dead code given both `_load_models_v1/v2` already filter, and the test reaches it through the private function rather than a real path. The only reachable bit is the `columns` container type straight from YAML. Closing this, and I'll follow up with a focused two-line guard on `cols` tested through `compose_build_instruction` on a fixture whose `metadata.yml` has a non-list `columns`. Thanks for the precise callsite walk.

PR #2549, opened 2026-07-20
fix(dbt): skip unusable models/columns in seed payloads

Bartok9:
## Summary

- `_seed_model_payload` / `_build_dbt_query_pairs` assumed dict models and named columns.
- Partial dbt imports with junk rows raised `TypeError`/`KeyError` during seed pair generation.
- Return `None` for unusable models/relationships, skip bad columns, harden `_camelize_props`.

## Verification

```text
$ cd core/wren && .venv/bin/python -m pytest tests/unit/test_dbt_seed_payload_guards.py -q
5 passed
```

Apache-2.0: `core/wren/**` only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dbt seed-query generation by safely handling missing or malformed model, relationship, column, and property data.
  * Invalid model/relationship inputs are now skipped, preventing crashes during manifest construction.
  * Ensures only valid entries contribute to generated seed queries, while preserving correct source and data source annotations.
* **Tests**
  * Added unit tests covering defensive payload guards and tolerance behaviors for malformed inputs.

### DIFF
--- a/core/wren/src/wren/dbt.py
+++ b/core/wren/src/wren/dbt.py
@@ -1251,8 +1251,16 @@ def _build_dbt_query_pairs(
     from wren.memory.seed_queries import generate_seed_queries  # noqa: PLC0415
 
     manifest = {
-        "models": [_seed_model_payload(model) for model in imported_models],
-        "relationships": [_seed_relationship_payload(rel) for rel in relationships],
+        "models": [
+            payload
+            for model in imported_models
+            if (payload := _seed_model_payload(model)) is not None
+        ],
+        "relationships": [
+            payload
+            for rel in relationships
+            if (payload := _seed_relationship_payload(rel)) is not None
+        ],
     }
 
     pairs = generate_seed_queries(manifest)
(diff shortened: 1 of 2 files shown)


goldmedal:
Suggesting we close this one — and here the guard doesn't just fail to fire, it would hide a bug if it ever did.

**Both inputs are produced by our own code, not by dbt.** At the only callsite (`dbt.py:351`):

```python
(imported_models, ...) = _build_imported_models(artifacts)
relationships, test_events = _apply_dbt_test_enrichment(artifacts, imported_models)
query_pairs = _build_dbt_query_pairs(imported_models, relationships, datasource=...)
```

The dbt artifacts are user input, yes — but `imported_models` and `relationships` are dicts *we construct* from them. So `isinstance(model, dict)` and `not model.get("name")` in `_seed_model_payload` are asserting that `_build_imported_models` didn't emit a malformed row. Same argument as #2572: a guard that can only fire when our own builder is broken belongs as an invariant, not as per-consumer re-validation.

**And the failure mode gets worse, not better.** Today a nameless model raises `KeyError: 'name'` at `model["name"]` — loud, with a stack trace pointing at the payload builder. After this change `_seed_model_payload` returns `None` and the model is silently dropped from the seed manifest, so a real defect in `_build_imported_models` would surface as "some seed queries are missing" long after the import. Turning a crash into silent data loss is the wrong direction for something that is by construction unreachable.

If there's an actual reproduction where `_build_imported_models` yields a row without a `name`, that's the bug worth a PR — in `_build_imported_models`, with a test driving it from a real dbt artifact fixture.

Bartok9:
Good point — `imported_models`/`relationships` are dicts we construct, so the guard can only fire if `_build_imported_models` is itself broken, and turning today's loud `KeyError: 'name'` into a silently-dropped seed model is the wrong direction. Closing. If a real dbt-artifact fixture ever produces a nameless row, the fix belongs in `_build_imported_models` with a test driving it from that artifact. Thanks.

PR #2557, opened 2026-07-21
fix(mysql): strip trailing semicolon on unlimited query path

Bartok9:
## Summary

MySQL Connector.query always strips before LIMIT so unlimited SEL CET matches dry_run (core/wren Apache-2.0).

pytest core/wren/tests/unit/test_mysql_unlimited_semicolon.py — 1 passed

Author Bartok9

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MySQL query execution by consistently removing trailing semicolons, including queries without a row limit.

* **Tests**
  * Added coverage to verify semicolon handling occurs before limit processing.

### DIFF
--- a/core/wren/src/wren/connector/mysql.py
+++ b/core/wren/src/wren/connector/mysql.py
@@ -90,6 +90,9 @@ class MySqlConnector(ConnectorABC):
 
     def query(self, sql: str, limit: int | None = None) -> pa.Table:
         limit = _coerce_limit(limit)
+        # Always strip terminator so unlimited execute matches dry_run EXPLAIN
+        # and LIMIT rewrite composition for client-pasted ``SELECT 1;``.
+        sql = strip_trailing_semicolon(sql)
         if limit is not None:
             sql = _apply_limit(sql, limit)
         with closing(self.connection.cursor()) as cursor:
--- a/core/wren/tests/unit/test_mysql_unlimited_semicolon.py
+++ b/core/wren/tests/unit/test_mysql_unlimited_semicolon.py
@@ -1,0 +1,16 @@
+"""MySQL unlimited query strips trailing semis (source pin)."""
+
+from pathlib import Path
+
+SRC = Path(__file__).resolve().parents[2] / "src" / "wren" / "connector" / "mysql.py"
+
+
+def test_query_always_strips_before_limit():
+    text = SRC.read_text(encoding="utf-8")
+    start = text.index("def query(self, sql: str, limit: int | None = None)")
+    end = text.index("def dry_run(self, sql: str)", start)
+    body = text[start:end]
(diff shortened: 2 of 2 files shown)


goldmedal:
Same note as on the sibling PRs in this series: the direction matches what already landed for Athena in #2535, but the test needs to exercise the connector rather than its source text.

`test_mysql_unlimited_semicolon.py` reads `mysql.py` as text and asserts the literal line `sql = strip_trailing_semicolon(sql)` appears before `if limit is not None:`. That passes even when the behavior is broken and fails on a harmless refactor — it pins the implementation, not the contract. Please write it as a mock-cursor test in the style of `core/wren/tests/unit/test_athena_strip_unlimited_query.py` (added with #2535): stub the cursor, call `query("SELECT 1;")`, assert on what reached `cursor.execute`.

Also worth clarifying in the description: `_apply_limit` already strips (`mysql.py:41`), so the limited path was never affected, and `mysql-connector-python` accepts a trailing semicolon for a single statement with multi-statement mode off. So what is the failure this fixes? If it's consistency with `dry_run`/limited composition rather than a driver error, that's fine — just state it that way.

goldmedal:
### Code review

Superseded — #2595 landed this change for MySQL in `5d2637c102b22c6a44911875638c52eb03e87f8b`, using an `else:` branch on the unlimited path. I'd suggest closing this one.

Worth noting because it isn't obvious from the PR page: this still reports `mergeable: true` against the updated `main`, and merging it would not conflict. It would produce a double strip:

```python
limit = _coerce_limit(limit)
# Always strip terminator so unlimited execute matches dry_run EXPLAIN ...
sql = strip_trailing_semicolon(sql)      # this PR
if limit is not None:
    sql = _apply_limit(sql, limit)
else:
    # Unlimited path: strip trailing terminators ...
    sql = strip_trailing_semicolon(sql)  # already on main via #2595
```

(verified with `git merge-tree --write-tree origin/main <head>`). Functionally harmless since the helper is idempotent, but it leaves two contradictory comments describing the same operation.

One process note that applies to the whole series: the test here pins implementation *text* rather than behaviour —

https://github.com/Canner/WrenAI/blob/2c6ef6a32da0e2bfa9ddfffbfe17ba3e628f2e38/core/wren/tests/unit/test_mysql_unlimited_semicolon.py#L8-L16

It reads `mysql.py` as a string and asserts on substrings and their relative order, so the connector is never executed. It passes whether or not the code path works, and any reformatting breaks it. The mocked-execution tests in the later PRs (#2592, #2593, #2594) are the right shape.

Bartok9:
Superseded by #2595 (already merged) — and you're right this would double-strip on top of it. Closing. Thanks.

PR #2565, opened 2026-07-22
fix(ask): guided template's recall step uses a valid CLI option

### LINKED ISSUE #2503: [Bug] The --guided instructions use an outdated wren memory recall option
### Environment

- WrenAI version: 0.13.0
- Operating system: Windows
- Command: `wren ask "test" --guided`

### Description

When running `wren ask "test" --guided`, the generated guide contains the following instruction:
wren memory recall --nl "<keywords>"
However, the current wren memory recall command accepts --query or -q. The --nl option belongs to wren memory store.

As a result, following the generated guide causes the command to fail because --nl is not a valid option for wren memory recall.
The guided workflow also does not mention wren memory fetch, even though the current CLI provides this command for retrieving schema context:
wren memory fetch --query "<question>"

Expected behavior
The guided instructions should use the current command syntax:
wren memory recall --query "<keywords>"
# or
wren memory recall -q "<keywords>"

wren memory fetch --query "<question>"
Please update the guided template so that it stays synchronized with the current CLI commands and documentation. It would also be helpful to add a regression test to prevent future mismatches between the generated guide and the CLI options.
References
https://github.com/Canner/WrenAI/blob/main/core/wren/src/wren/ask_templates/guided.md.tmpl
https://github.com/Canner/WrenAI/blob/main/core/wren/src/wren/memory/cli.py
https://github.com/Canner/WrenAI/blob/main/core/wren/docs/cli.md

AmirF194:
`guided.md.tmpl` step 2 tells the agent to run `wren memory recall --nl "<keywords>"`. `recall` only accepts `--query`/`-q` (`core/wren/src/wren/memory/cli.py`); `--nl` belongs to the unrelated `store` command. Following the guided flow as written fails at option parsing.

Reproduced against current HEAD (`f4b45edd`, `python:3.11-slim` in Docker, `uv sync`, `print(wren.__file__)` confirmed the installed package is this checkout):

```
$ wren memory recall --nl foo
Usage: wren memory recall [OPTIONS]
Try 'wren memory recall --help' for help.
Error: No such option: --nl Did you mean --help?
```

Fix: change the template's step 2 to `--query`.

Added a regression test that parses the exact `wren memory recall ...` line the guided template emits and invokes it against the real CLI, asserting it does not hit a click usage error (exit code 2). It fails on `main` (`--nl` is a usage error) and passes on this branch (`--query` reaches the command's own logic).

Ran `tests/unit/` (excluding `test_memory.py`/`test_mcp_server.py`, matching this repo's `test-unit` CI job) and `ruff format --check` / `ruff check` on `src/`: clean except a pre-existing, unrelated `NameError` in `test_redshift_semicolon.py` / `src/wren/connector/redshift.py` that also fails on unmodified `main`.

Did not add the `wren memory fetch` mention the issue also suggests; that is a template enhancement rather than the reported bug, so leaving it for a separate change if wanted.

Fixes #2503

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated guided memory recall instructions to use the correct CLI short flag for keywords.
  * Prevented guided recall workflows from failing due to an invalid CLI argument.
* **Tests**
  * Added a regression test to confirm the guided “step 2” emits a valid `wren memory recall` invocation and runs without a “no such option” error.

### DIFF
--- a/core/wren/src/wren/ask_templates/guided.md.tmpl
+++ b/core/wren/src/wren/ask_templates/guided.md.tmpl
@@ -3,7 +3,7 @@ Identify the task type, then follow the matching flow:
 
 TASK TYPE A — data question:
   1. wren context show [--path <project>]   # see MDL models
-  2. wren memory recall --nl "<keywords>"   # similar past queries (skip if no memory)
+  2. wren memory recall --query "<keywords>"  # similar past queries (skip if no memory)
   3. write SQL using model names (not raw tables)
   4. wren dry-plan --sql '...'              # validate non-trivial SQL
   5. wren query --sql '...'                 # execute
--- a/core/wren/src/wren/ask_templates/guided.md.tmpl
+++ b/core/wren/src/wren/ask_templates/guided.md.tmpl
@@ -3,7 +3,7 @@ Identify the task type, then follow the matching flow:
 
 TASK TYPE A — data question:
   1. wren context show [--path <project>]   # see MDL models
-  2. wren memory recall --query "<keywords>"  # similar past queries (skip if no memory)
+  2. wren memory recall -q "<keywords>"  # similar past queries (skip if no memory)
   3. write SQL using model names (not raw tables)
   4. wren dry-plan --sql '...'              # validate non-trivial SQL
   5. wren query --sql '...'                 # execute
--- a/core/wren/tests/unit/test_ask_cli.py
+++ b/core/wren/tests/unit/test_ask_cli.py
@@ -58,6 +58,24 @@ def test_render_unknown_mode_raises():
         ask_mod.render("auto", "anything")
 
 
+def test_guided_recall_step_uses_a_real_cli_option():
+    # The guided template's step 2 must emit an option `wren memory recall`
+    # actually accepts. Regression for #2503: the template said `--nl`, an
+    # option that belongs to `wren memory store`, not `recall`.
+    result = runner.invoke(app, ["ask", "x", "--guided"])
+    assert "--nl" not in result.output
+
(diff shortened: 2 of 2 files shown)


AmirF194:
Heads up on the two red checks (lint and unit tests): both fail on the same line, `src/wren/connector/redshift.py:55`, where `_strip_trailing_semicolon` is referenced but never defined. This file isn't touched by this PR, only `guided.md.tmpl` and `tests/unit/test_ask_cli.py` are, and I reproduced the identical `NameError` on unmodified main at `f4b45edd` in a clean container. Looks like a pre-existing issue unrelated to this change.

core/wren/src/wren/ask_templates/guided.md.tmpl
```diff
@@ -3,7 +3,7 @@ Identify the task type, then follow the matching flow:
 
 TASK TYPE A — data question:
   1. wren context show [--path <project>]   # see MDL models
-  2. wren memory recall --nl "<keywords>"   # similar past queries (skip if no memory)
+  2. wren memory recall --query "<keywords>"  # similar past queries (skip if no memory)
```
goldmedal:
```suggestion
  2. wren memory recall -q "<keywords>"  # similar past queries (skip if no memory)
```
Aligned the style with other skills. Likes:
https://github.com/Canner/WrenAI/blob/fcd8c02e329ef74c353da6a6b0fdc03c5094b266/core/wren/src/wren/skills_content/usage/SKILL.md?plain=1#L87

goldmedal (approved):
Thanks @AmirF194, make sense to me.

PR #2567, opened 2026-07-23
fix(context): reject non-dict models/views in MDL conversion

Bartok9:
## Summary
- `convert_mdl_to_project` assumed every `models`/`views` entry was a dict; list litter crashed during snake_case convert.
- Raise a clear `ValueError` naming the index and actual type.
- `validate_project` records a `ValidationError` for non-dict loaded entries instead of `AttributeError`.

## License
`core/wren/**` Apache-2.0.

## Test plan
- [x] `pytest tests/unit/test_convert_mdl.py -k non_dict -v` (2 passed)

## Real behavior proof
Non-dict model raised TypeError-ish path inside convert; now ValueError with index.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation when importing project definitions.
  * Clear errors are now reported when model or view entries have an invalid format, including the affected item’s position.
  * Missing or empty model and view lists are handled safely.

* **Tests**
  * Added coverage for invalid model and view entries during import.

### DIFF
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -197,7 +197,11 @@ def convert_mdl_to_project(mdl_json: dict) -> list[ProjectFile]:
     )
 
     # ── Models ────────────────────────────────────────────────
-    for i, model in enumerate(mdl_json.get("models", [])):
+    for i, model in enumerate(mdl_json.get("models", []) or []):
+        if not isinstance(model, dict):
+            raise ValueError(
+                f"Model at index {i} must be an object, got {type(model).__name__}"
+            )
         model_snake = _convert_keys_to_snake(model)
         if "name" not in model_snake:
             raise ValueError(f"Model at index {i} is missing required 'name' field")
@@ -226,7 +230,11 @@ def convert_mdl_to_project(mdl_json: dict) -> list[ProjectFile]:
         )
 
     # ── Views ─────────────────────────────────────────────────
-    for i, view in enumerate(mdl_json.get("views", [])):
+    for i, view in enumerate(mdl_json.get("views", []) or []):
+        if not isinstance(view, dict):
+            raise ValueError(
+                f"View at index {i} must be an object, got {type(view).__name__}"
+            )
         view_snake = _convert_keys_to_snake(view)
         if "name" not in view_snake:
             raise ValueError(f"View at index {i} is missing required 'name' field")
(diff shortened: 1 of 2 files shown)


goldmedal:
Thanks — this one landed near something real, so it's worth being specific about which parts hold up.

**`validate_project`, views half — there is a genuine bug here, and you found it.** I opened #2597 with a reproduction. A legacy (`schema_version: 1`) project whose hand-written `views.yml` contains a non-mapping entry really does crash, because `_load_views_v1` is the only loader that doesn't filter its list items:

```python
return data.get("views", []) if isinstance(data, dict) else []   # outer dict checked, items not
```

`_load_models_v1`, `_load_models_v2` and `_load_views_v2` all guard the items; that one doesn't, while its annotation still promises `list[dict]`.

**But the fix belongs in the loader, not the consumer.** `_load_views_v1` has four consumers and all of them assume mappings — `validate_project`, `build_manifest`/`build_json`, and both migration paths `_plan_v1_to_v2` / `_apply_v1_to_v2`. Patching `validate_project` alone leaves the other three crashing, and the migration one is the nastiest: `_apply_v1_to_v2` reaches the views loop only *after* it has written the new `models/<name>/` directories and `unlink()`ed the old flat files, so an `AttributeError` there leaves the project half-migrated. A one-line item filter in `_load_views_v1` fixes all four at once.

**`validate_project`, models half is dead code.** Both `_load_models_v1` and `_load_models_v2` already do `isinstance(..., dict)` filtering, so a non-dict can't reach that loop.

**`convert_mdl_to_project` — the stated motivation doesn't hold.** The description says list litter "crashed during snake_case convert", but `_convert_keys_to_snake` returns non-dict/non-list values untouched, so it doesn't crash there. Actual current behaviour:

| input | today |
| --- | --- |
| `models: ["not-a-dict"]` | already `ValueError: Model at index 0 is missing required 'name' field` |
| `models: [None]` / `[42]` | `TypeError: argument of type 'NoneType' is not iterable` |
| `models: ["my name x"]` | `TypeError: string indices must be integers` (the `"name" not in model_snake` substring check passes on a str) |

Note the first row is the exact input in `test_convert_mdl_rejects_non_dict_model` — the old code already raises a `ValueError` for it. So that half is an error-message improvement, not a crash fix. It's at a legitimate boundary (raw user JSON, before `wren-core` sees it), so I'm not opposed to it in principle, but it should be described as what it is and ideally cover the `TypeError` cases above rather than the one case that already worked.

I'm going to pass on this version. If you'd like to take #2597, a focused PR moving the guard into `_load_views_v1` — with a test that goes through `validate_project`/`build_manifest` on a real v1 project fixture rather than calling the helper directly — would be very welcome.

Bartok9:
This is a great review — thank you for pinning down exactly which half is real. You're right on all points: the models half is dead (both loaders already filter), the snake_case motivation was overstated, and the genuine bug is `_load_views_v1` not filtering its list items while its four consumers all assume mappings — with `_apply_v1_to_v2` being the dangerous one since it can leave a half-migrated project. Putting the guard in the consumer was the wrong layer. I'll take #2597: a focused one-line item filter in `_load_views_v1`, with a test that exercises `validate_project`/`build_manifest` on a real v1 project fixture rather than the helper directly. Closing this one in favor of that.

PR #2569, opened 2026-07-23
fix(sdk): coerce non-str tool descriptions in system prompts

Bartok9:
## Summary
- Shared `_build_tools_section` in langchain + pydantic assumed `tool.description` is \'`str|None`\'.
- Non-str descriptions raised `AttributeError` on `.strip()`.
- Coerce description (and missing name) before rendering markdown bullets.

## License
`sdk/**` Apache-2.0.

## Test plan
- [x] langchain: `test_prompt_tool_description.py` (1 passed)
- [x] pydantic: `test_instructions_tool_description.py` (1 passed)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved “Available tools” rendering to tolerate missing/invalid tool names and non-text descriptions, ensuring consistent bullet labels and safer description formatting (including first-line truncation).

* **Tests**
  * Expanded prompt/instruction unit test coverage for tool-name fallback behavior (missing/empty/invalid `name`, using `__name__`) and for non-string/multiline `description` handling.

### DIFF
--- a/sdk/wren-langchain/src/wren_langchain/_prompt.py
+++ b/sdk/wren-langchain/src/wren_langchain/_prompt.py
@@ -209,7 +209,9 @@ def _tool_name(tool) -> str:
     when the ``name`` attribute is missing or not a non-empty string."""
     name = getattr(tool, "name", None)
     if not isinstance(name, str) or not name:
-        name = getattr(tool, "__name__", "tool")
+        name = getattr(tool, "__name__", None)
+    if not isinstance(name, str) or not name:
+        name = "tool"
     return name
 
 
--- a/sdk/wren-langchain/src/wren_langchain/_prompt.py
+++ b/sdk/wren-langchain/src/wren_langchain/_prompt.py
@@ -209,8 +209,14 @@ def _build_tools_section(tool_list: list) -> str:
         return ""
     lines = ["## Available tools"]
     for tool in tool_list:
-        description = (tool.description or "").strip().split("\n")[0]
-        lines.append(f"- `{tool.name}`: {description}")
+        raw_desc = getattr(tool, "description", None)
+        if not isinstance(raw_desc, str):
+            raw_desc = "" if raw_desc is None else str(raw_desc)
+        description = raw_desc.strip().split("\n")[0]
+        name = getattr(tool, "name", None)
+        if not isinstance(name, str) or not name:
+            name = getattr(tool, "__name__", "tool")
+        lines.append(f"- `{name}`: {description}")
     return "\n".join(lines)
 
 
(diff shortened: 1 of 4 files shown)


Bartok9:
Thanks @coderabbitai — good call. Addressed in the latest push (76bb0509):

- Centralized validated tool-name extraction into a `_tool_name()` helper in both `sdk/wren-langchain/src/wren_langchain/_prompt.py` and `sdk/wren-pydantic/src/wren_pydantic/_instructions.py`. It now feeds both the workflow-detection `tool_names` set and the bullet rendering, so a missing/non-str `name` falls back consistently (`__name__` → `"tool"`) everywhere.
- Added regression tests covering `None`/empty/numeric names and the `__name__` fallback in both SDK test modules, alongside the existing description-coercion cases.

Ruff clean and unit tests green locally.

Bartok9:
Thanks @coderabbitai — both addressed in eb22875:

- `_tool_name()` now validates the `__name__` fallback with the same non-empty-string check as `name`, so a missing/empty/non-str `__name__` cleanly falls through to `"tool"` in both the langchain and pydantic SDKs.
- Tightened `test_build_tools_section_non_str_description` to assert the exact rendered line (`- \`wren_query\`: ` via `out.splitlines()`), so the `None`-description coercion to empty is actually covered rather than a substring match.

Ruff clean, unit tests green in both SDKs.

goldmedal:
Closing this one — the name fallback makes the prompt wrong rather than merely ugly, and the test can't validate it.

**`_tool_name` fabricates a tool name.** When `name` is missing or not a non-empty string, it falls back to the literal `"tool"`, and `_build_tools_section` then writes that into the `## Available tools` block of the system prompt. The model is being told a tool called `tool` exists and can be called; it can't. Silently listing a non-callable name is worse than the current `AttributeError`, which at least points at the malformed tool object. If a tool can't be identified, skipping it (or failing loudly) is the safer behavior.

**The helper is duplicated verbatim.** The same `_tool_name` is added to both `sdk/wren-langchain/src/wren_langchain/_prompt.py` and `sdk/wren-pydantic/src/wren_pydantic/_instructions.py`, so the two copies will drift.

**The tests can't see the real module.** Both new files `ast.parse` the source, filter out two `FunctionDef` nodes, `exec(compile(...))` them into a bare namespace, and test that. Nothing about the actual module — its imports, the rest of `build_system_prompt`, the call sites that were changed — is exercised.

The underlying concern (third-party tool objects with a non-`str` `description`) is legitimate. A version that keeps the description coercion, drops the name fallback, and tests through the public `build_system_prompt` / `build_instructions` entry points with `SimpleNamespace` tools would be welcome.

Bartok9:
Fair call — agreed. The `"tool"` fallback silently fabricates a wrong tool name in the prompt rather than surfacing the real problem, and the test can't meaningfully validate a fabricated name. Better to let a missing/invalid `name`+`__name__` fail loudly. Closing. Thanks for the careful read.

PR #2570, opened 2026-07-23
fix(wren-cli): surface skipped non-mapping rows in parse/translate-types

### LINKED ISSUE #2528: wren-core: surface skipped non-mapping rows in parse/translate-types CLI (collector + --strict)
### Background

#2508 made `parse_types` / `translate_types` (`core/wren/src/wren/type_mapping.py`) skip non-dict rows instead of crashing the whole batch on `dict(col)`. That's the right minimal fix, but it currently drops bad rows **silently** — in the CLI path (`parse-types` / `translate-types`), the output array is shorter than the input with no signal, which for a schema tool can mean a subtly incomplete result.

### Proposal

Add a `context_cli`-aligned error collector so skipped rows aren't invisible:

- **Keep the library functions pure** (`list[dict] -> list[dict]`). Put the collector at the CLI layer (`core/wren/src/wren/utils_cli.py`), where the raw input list is already available — matching how `context_cli.py` collects warnings (with a summary threshold) and honors `--strict`.
- **Accumulate skipped entries** (index + value) and report them once to `stderr`, e.g. `skipped N non-mapping row(s)`.
- **Add a `--strict` flag** that turns the warning into a non-zero exit.
- **Distinguish intent:** `None` is plausibly benign padding (skip quietly); `str` / `int` / other scalars are more likely real corruption worth surfacing. Decide and document what counts as an error vs. benign.

### Related (may land in #2508 directly)

- Guard on `collections.abc.Mapping` rather than `dict`, so mapping-likes (e.g. `RowMapping`, `types.MappingProxyType`) aren't dropped.
- Update the `parse_types` / `translate_types` docstrings to note non-mapping rows are skipped (`len(out) <= len(in)`).

### Files

- `core/wren/src/wren/utils_cli.py`
- `core/wren/src/wren/type_mapping.py`
- `core/wren/tests/unit/test_type_mapping.py`

AmirF194:
`parse_types()`/`translate_types()` (`core/wren/src/wren/type_mapping.py`, fixed in #2508) skip any row that isn't a `Mapping` instead of crashing the batch. The CLI commands built on top of them only reported a bare `skipped N` count, so a caller had no way to tell whether the dropped rows were benign padding or something worth investigating, and no way to see which rows were dropped.

Per the issue's proposal, the library functions stay pure (`list -> list`, no signature change). The CLI layer in `utils_cli.py` now recomputes which input rows are non-`Mapping` (the same check `parse_types`/`translate_types` apply internally) before calling them, then reports on that list:

- `None` rows are counted and logged once as benign padding (dynamic schema exporters routinely interleave `None`), never treated as an error.
- Any other non-mapping value (a bare string, number, list, ...) is logged individually with its index and value, since it more plausibly signals real corruption. The list is capped at 10 named entries with an `... and N more` tail so a large batch can't flood stderr.
- Both `parse-types` and `translate-types` gain `--strict`, which exits 1 only when at least one non-`None` row was dropped. A batch that only drops `None` padding still exits 0 under `--strict`, matching the benign/error split above.

Verified in Docker (`python:3.11-slim`, `uv sync`, `print(wren.__file__)` confirmed the installed package resolves to this checkout): added 7 named regression tests (`--strict` exit code on a corrupt row vs. a `None`-only batch, the note/warning wording, index+value reporting, and the 10-row truncation), each fails on unmodified `main` and passes on this branch. Ran the full `tests/unit/` tree (excluding `test_memory.py`/`test_mcp_server.py`, matching this repo's `test-unit` CI job: 1009 passed, 2 skipped) and `ruff format --check` / `ruff check` on `src/`: clean.

Did not touch the `Mapping` guard itself or the docstrings (`#2508` already covers both) and did not add a `--verbose`/grouped-summary mode like `context_cli.py`'s warning printer; the batches this CLI handles are schema-sized, not log-sized, so the flat 10-row cap seemed like the right amount of ceremony for this file.

Fixes #2528

## Summary by CodeRabbit

* **New Features**
  * Enhanced batch type-processing feedback by identifying skipped non-mapping input rows, reporting their indices/values with truncation for large batches.
  * Added a `--strict` mode to `parse-types` and `translate-types` that exits non-zero when corruption-like rows are detected while still printing valid results.
  * Improved input validation by rejecting non-array top-level JSON payloads with a clear error.
* **Bug Fixes**
  * Clarified reporting so skipped `None` rows are treated as benign padding (note) and separated from corruption warnings.
* **Tests**
  * Expanded CLI integration tests for strict/benign behavior, warning truncation limits, and non-array payload rejection.

### DIFF
--- a/core/wren/src/wren/utils_cli.py
+++ b/core/wren/src/wren/utils_cli.py
@@ -4,12 +4,63 @@ from __future__ import annotations
 
 import json
 import sys
+from collections.abc import Mapping
 from pathlib import Path
 from typing import Annotated, Optional
 
 import typer
 
 utils_app = typer.Typer(name="utils", help="Utility commands")
+
+# Cap on how many dropped rows we list individually; past this we still
+# report the total but stop naming each one, mirroring context_cli's
+# _WARNING_SUMMARY_THRESHOLD so large batches don't flood stderr.
+_SKIP_REPORT_LIMIT = 10
+
+
+def _skipped_rows(data: object) -> list[tuple[int, object]]:
+    """Return (index, value) for entries parse_types/translate_types will drop.
+
+    Mirrors the ``isinstance(col, Mapping)`` guard those functions apply
+    internally; recomputed here (rather than threaded back out of the pure
+    library functions) so the CLI layer can report which rows were skipped
+    and why.
+    """
+    if not isinstance(data, list):
+        return []
+    return [(i, row) for i, row in enumerate(data) if not isinstance(row, Mapping)]
+
+
+def _report_skipped(skipped: list[tuple[int, object]]) -> None:
+    """Warn about dropped rows, splitting benign padding from likely corruption.
+
+    ``None`` is treated as benign padding from dynamic schema exporters and
+    only counted. Any other non-mapping value (a bare string, number, list,
+    ...) is treated as more likely a real data problem and listed with its
+    index and value so it isn't invisible in the output.
+    """
+    if not skipped:
+        return
+    benign = [i for i, v in skipped if v is None]
+    corrupt = [(i, v) for i, v in skipped if v is not None]
+    if benign:
+        typer.echo(
+            f"Note: skipped {len(benign)} None row(s) (benign padding)", err=True
+        )
+    if corrupt:
+        typer.echo(
+            f"Warning: skipped {len(corrupt)} non-mapping row(s) with unexpected values:",
+            err=True,
+        )
+        for i, v in corrupt[:_SKIP_REPORT_LIMIT]:
+            typer.echo(f"  [{i}] {type(v).__name__}: {v!r}", err=True)
+        remaining = len(corrupt) - _SKIP_REPORT_LIMIT
+        if remaining > 0:
+            typer.echo(f"  ... and {remaining} more", err=True)
+
+
+def _has_corrupt_skips(skipped: list[tuple[int, object]]) -> bool:
+    return any(v is not None for _, v in skipped)
 
 
 @utils_app.command(name="parse-type")
@@ -37,6 +88,16 @@ def parse_types_cmd(
         Optional[str],
         typer.Option("--input", "-i", help="Input JSON file (default: stdin)"),
     ] = None,
+    strict: Annotated[
+        bool,
+        typer.Option(
+            "--strict",
+            help=(
+                "Exit 1 if any dropped row held a non-None value "
+                "(likely corruption, not benign padding)."
+            ),
+        ),
+    ] = False,
 ):
     """Batch-normalize types. Reads JSON array from stdin or file, writes JSON to stdout.
 
@@ -63,13 +124,12 @@ def parse_types_cmd(
         typer.echo(f"Error: invalid JSON input: {e}", err=True)
         raise typer.Exit(1)
 
+    skipped = _skipped_rows(data)
     results = parse_types(data, dialect, type_field=type_field)
-    if isinstance(data, list) and len(results) < len(data):
-        typer.echo(
-            f"Note: skipped {len(data) - len(results)} non-mapping row(s)",
-            err=True,
-        )
+    _report_skipped(skipped)
     typer.echo(json.dumps(results, indent=2))
+    if strict and _has_corrupt_skips(skipped):
+        raise typer.Exit(1)
 
 
 @utils_app.command(name="translate-type")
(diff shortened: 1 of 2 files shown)


AmirF194:
The failing lint check is `ruff format --check src/` flagging two markdown files, `src/wren/skills_content/dlt-connector/SKILL.md` and `src/wren/skills_content/dlt-connector/references/dlt_sources.md`, for reformatting. This PR only touches `core/wren/src/wren/utils_cli.py` and its test file, neither of those markdown files is part of the diff, and both have been unchanged since #2399 landed on 2026-06-26. So this looks like pre-existing formatting drift, not something introduced here.

goldmedal:
Reviewed this locally (checked out the PR head and `main` side by side, `uv sync`, ran the suite and probed the CLI edges). The design lands where the issue asked: library functions stay pure, the collector sits at the CLI layer, and the benign/corrupt split is drawn at `None`. Nice touch that `--strict` stays consistent with `context_cli.py`'s "treat warnings as errors" — `None` padding only emits a `Note:`, never a `Warning:`, so it correctly doesn't trip the exit. `tests/unit/test_type_mapping.py` passes (59) and `ruff format --check src/` / `ruff check src/` are clean on `utils_cli.py`.

Two things I'd like to see before merge.

### 1. The skip predicate is now duplicated with no drift guard

`_skipped_rows()` hand-copies the `isinstance(col, Mapping)` rule from `parse_types` / `translate_types`. That's the right call given the issue asked to keep the library pure — but it means if the skip condition in `type_mapping.py` ever changes (say it also starts skipping mappings missing `type_field`), the CLI report goes **silently wrong**, and no existing test fails.

Cheap way to pin the invariant — either assert it in the command, or add a regression:

```python
assert len(results) + len(skipped) == len(data)
```

That keeps the two sites honest without giving up the pure `list -> list` signature.

### 2. `_SKIP_REPORT_LIMIT` caps the row count but not the output volume

The comment says the cap exists "so large batches don't flood stderr", but `{v!r}` is unbounded, so a batch of *few but large* rows still floods. Measured on this branch:

| input | stderr |
| --- | --- |
| 2 corrupt rows (3 KB string + 400-element list) | 4,366 bytes |
| 10 corrupt rows × 50 KB string | **500,204 bytes** |

Ten rows is exactly the cap, so the guard doesn't engage at all. Bounding the per-value repr closes it:

```python
typer.echo(f"  [{i}] {type(v).__name__}: {v!r:.120}", err=True)
```

(A corrupt row's *type* and *position* are the actionable parts; the first ~120 chars of the value is plenty to identify it.)

AmirF194:
Both addressed, pushed to the branch (85c1fd4c):

1. Added the invariant assertion at both call sites: `assert len(results) + len(skipped) == len(data)`, right where you suggested. If the mirrored `Mapping` check in `utils_cli.py` ever drifts from the real one in `type_mapping.py`, this now fails loudly instead of silently under-reporting.

2. Bounded each corrupt value's repr to 120 chars (`{v!r:.120}`). Confirmed your measurement: a batch under the 10-row cap with a 50 KB value still flooded stderr before this. Added a regression test with a 5000-char value, fails on the prior code and passes now.

Full `tests/unit/` (excluding the two modules with a missing optional dep, `test_memory.py`/`test_mcp_server.py`, same as this repo's `test-unit` CI job) plus `ruff check`/`ruff format --check` on `src/`: all green.

goldmedal (changes requested):
The new invariant assumes the decoded input is a list, but JSON decoding can return any JSON type. A top-level object now reaches the assertion and produces an internal traceback instead of a clean input error.

Reproduction from `core/wren`:

```bash
printf '%s\n' '{"column":"id","raw_type":"int8"}' \
  | uv run --no-sync wren utils parse-types --dialect postgres
```

This exits 1 with:

```text
AssertionError: parse_types dropped a different set of rows than _skipped_rows predicted
```

Because the command documents a JSON array as its input, please validate the top-level shape before calling `parse_types`/`translate_types` and return a clean error such as `Error: input must be a JSON array`. The same issue applies to `translate-types`.

AmirF194:
Good catch, and a clean repro. Pushed 68f6e99: both commands now validate the top-level JSON shape before calling `parse_types`/`translate_types`, so an object or a string exits 1 with `Error: input must be a JSON array (got dict)` instead of the internal assertion. Added regression tests for object and string payloads on both commands (`test_cli_parse_types_rejects_non_list_object_payload` and friends).

goldmedal (approved):
Thanks @AmirF194, LGTM. I filed a follow-up issue https://github.com/Canner/WrenAI/issues/2629 for a non-blocking change.

AmirF194:
Thanks for the thorough review, both points made the fix better.

PR #2572, opened 2026-07-24
fix(mdl): skip non-dict models/columns/views in CTERewriter

Bartok9:
## Summary

- Skip non-dict / nameless models, columns, and views when building CTERewriter indexes
- Harden `_iter_model_column_names` and case-collision scans the same way

## License

Apache-2.0 (`core/**`).

## Motivation

A single non-object row in `manifest.models` / `columns` / `views` raised `TypeError` / `KeyError` during rewriter construction and blocked planning for the whole catalog. Match defensive guards already used in memory/mcp formatters.

## Verification

```text
cd core/wren && .venv/bin/python -m pytest tests/unit/test_cte_rewriter_nonduct.py -v
```

— 2 passed

## Duplicate check

- No open PR targeting CTERewriter non-dict model/view gates (engine name collection is #2546)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when processing malformed MDL manifests.
  * Invalid models, views, and column definitions are now safely ignored instead of causing processing errors.
  * Existing handling for case-only column name collisions is preserved.

* **Tests**
  * Added coverage for malformed manifest entries and invalid column definitions.

### DIFF
--- a/core/wren/src/wren/mdl/cte_rewriter.py
+++ b/core/wren/src/wren/mdl/cte_rewriter.py
@@ -136,16 +136,27 @@ class CTERewriter:
         self._model_cols: dict[str, list[str]] = {}
 
         for model in self.manifest.get("models", []):
-            name = model["name"]
+            if not isinstance(model, dict):
+                continue
+            name = model.get("name")
+            if not isinstance(name, str) or not name:
+                continue
             self.model_dict[name] = model
             cols: dict[str, str] = {}
             orig: dict[str, str] = {}
-            for col in model.get("columns", []):
+            raw_cols = model.get("columns", []) or []
+            if not isinstance(raw_cols, list):
+                raw_cols = []
+            for col in raw_cols:
+                if not isinstance(col, dict):
+                    continue
                 if col.get("isHidden"):
                     continue
                 if col.get("relationship"):
                     continue
-                col_name = col["name"]
+                col_name = col.get("name")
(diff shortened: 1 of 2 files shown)


goldmedal:
Passing on this one too, for the same reason as #2546 — but the argument is actually stronger here.

`CTERewriter` has exactly one production construction site, `engine.py:222`, and the line immediately before it is `get_session_context(effective_manifest, ...)` — same manifest string, same `try` block. Every shape this PR guards against (`null` / string / int rows, missing or non-str `name`) is rejected by `wren-core`'s serde deserialization at that call, so the rewriter is never constructed with a junk manifest.

On top of that, in the normal path `effective_manifest` is not user input at all: it is `extractor.extract_by(tables)` re-serialized via `to_json_base64()`, i.e. output produced *by* `wren-core` from an already-validated Rust `Manifest`. A non-dict model row there would mean `wren-core` emitted invalid JSON of its own.

So the guards can't fire, and the tests only exercise the guarded functions directly rather than any reachable path. I'd rather keep the invariant explicit — manifest schema is validated once, in `wren-core` — than add per-callsite re-validation that implies otherwise.

Bartok9:
Agreed — with `get_session_context(effective_manifest, ...)` on the line right before construction, and `effective_manifest` being `wren-core` output in the normal path, every shape guarded here is already rejected upstream. The invariant is better kept explicit in one place. Closing. Thanks for walking through it.

PR #2573, opened 2026-07-24
fix(policy): harden resolve_model_name against non-str names

Bartok9:
## Summary

- Filter non-str model names before exact/CI matching in `resolve_model_name`
- Return None when the lookup name is not a non-empty string

## License

Apache-2.0 (`core/**`).

## Motivation

Corrupt or partially-migrated MDL name collections could include `None`/ints. Calling `.lower()` on those raised and aborted strict-mode planning even when a usable string candidate existed.

## Verification

`cd core/wren && .venv/bin/python -m pytest tests/unit/test_resolve_model_name_nonstr.py -v` — 3 passed

## Duplicate check

- No open PR on `policy.resolve_model_name` non-str filtering

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model name resolution by safely handling invalid or empty lookup inputs.
  * Candidate model names that aren’t non-empty strings are now ignored to prevent lookup errors.
  * Preserves existing exact-match behavior for quoted model names (case-sensitive).

* **Tests**
  * Added unit test coverage for non-string, empty, and quoted lookup values, including scenarios that previously could raise errors.

### DIFF
--- a/core/wren/src/wren/policy.py
+++ b/core/wren/src/wren/policy.py
@@ -159,10 +159,7 @@ def resolve_model_name(
     # Drop non-str names so a corrupt models collection cannot crash .lower().
     if not isinstance(name, str) or not name:
         return None
-    if isinstance(model_names, (set, frozenset)):
-        model_set = {n for n in model_names if isinstance(n, str) and n}
-    else:
-        model_set = {n for n in model_names if isinstance(n, str) and n}
+    model_set = {n for n in model_names if isinstance(n, str) and n}
     if name in model_set:
         return name
     if quoted:
--- a/core/wren/src/wren/policy.py
+++ b/core/wren/src/wren/policy.py
@@ -156,9 +156,13 @@ def resolve_model_name(
     falls back to a case-insensitive scan. Returns ``None`` if no model
     matches.
     """
-    model_set = (
-        model_names if isinstance(model_names, (set, frozenset)) else set(model_names)
-    )
+    # Drop non-str names so a corrupt models collection cannot crash .lower().
+    if not isinstance(name, str) or not name:
+        return None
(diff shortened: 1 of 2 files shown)


Bartok9:
Good catch — collapsed the redundant set/frozenset branch into a single comprehension in 26e80d8. Behavior unchanged.

goldmedal:
Passing on this one — it's unreachable, and it costs something on the planning hot path.

**Unreachable.** For a non-str entry to reach `resolve_model_name`, the manifest would need something like `{"name": 99}`. `wren-core` rejects that (`invalid type: integer 99, expected a string`) at `get_manifest_extractor`, a few lines later in the same `try` block in `_plan` — so the request aborts either way. The `name` parameter guard is dead too: the only caller passes `t.name` from sqlglot, and `_plan` already does `if not t.name: continue` immediately above.

**Regression.** The line being replaced is a deliberate fast path:

```python
model_set = model_names if isinstance(model_names, (set, frozenset)) else set(model_names)
```

Callers pass a `set`, so today there is no copy. The patch replaces it with an unconditional set comprehension, so every call rebuilds the full name set — once per table reference, per query, scaling with the number of models in the manifest. That's a real cost in exchange for a guard that cannot fire.

Same reasoning as #2546 and #2572: manifest schema is validated once, in `wren-core`; I'd rather not re-validate it per callsite in Python.

Bartok9:
Fair call — you're right that `wren-core` rejects a non-str `name` at `get_manifest_extractor` in the same `try`, so the guard can't fire, and I missed that the comprehension drops the deliberate set/frozenset fast path. Not worth a per-callsite cost on the planning path for a dead branch. Closing this one. Thanks for the detailed read.

PR #2574, opened 2026-07-24
fix(spark): apply limit via DataFrame.limit before toPandas

Bartok9:
## Summary

- Apply `limit` with Spark `DataFrame.limit(n)` **before** `toPandas()`, so the engine gets a server-side `CollectLimit` instead of materializing the full result and slicing the Arrow table on the client
- Drop SQL subquery wrapping and the non-subqueryable classifier — unnecessary on Spark’s DataFrame API (same physical plan as wrap; works for `SHOW`/`DESCRIBE`/SELECT alike)
- Use shared `coerce_limit` from `connector/base.py` (#2624) so Spark matches the other connectors (rejects negatives, bools, and non-integral values — on `main`, `limit=-1` silently returned an empty table via `pa.Table.slice(0, -1)`)
- `dry_run` unchanged vs `main` (trailing `;` strip + `.limit(0).count()`)

## License

Apache-2.0 (`core/**`).

## Motivation

On `main`, Spark applied `limit` only after `toPandas()` via Arrow `slice`, so bounded asks still fully materialized. Maintainer measurements on PySpark 4.1.1 show `DataFrame.limit()` is already plan-level pushdown:

```
spark.sql("SELECT * FROM t").limit(3).explain()
== Physical Plan ==
CollectLimit 3
+- *(1) Range (0, 1000, step=1, splits=1)

spark.sql("SELECT * FROM (\nSELECT * FROM t\n) AS _q LIMIT 3").explain()
== Physical Plan ==
CollectLimit 3
+- *(1) Range (0, 1000, step=1, splits=1)
```

Identical plans — string-wrapping LIMIT is the right tool for DBAPI connectors, not Spark.

Bounded Spark queries (MCP always passes a default limit) fully materialize the result set, then client-slice with Arrow. Negative limits on `main` did not error: `pa.Table.slice(0, -1)` returned zero rows silently.

## Verification

`cd core/wren && .venv/bin/python -m pytest tests/unit/test_spark_semicolon.py -v` — 8 passed

## Duplicate check

- No open PR for Spark DataFrame limit-before-toPandas

### DIFF
--- a/core/wren/src/wren/connector/spark.py
+++ b/core/wren/src/wren/connector/spark.py
@@ -22,20 +22,26 @@ class SparkConnector(ConnectorABC):
         )
 
     def query(self, sql: str, limit: int | None = None) -> pa.Table:
-        df = self.connection.sql(strip_trailing_semicolon(sql)).toPandas()
+        # Push LIMIT into Spark SQL so the engine does not materialize the full
+        # result only for a client-side slice. Strip trailing ``;`` first so the
+        # outer subscript form stays valid.
+        cleaned = strip_trailing_semicolon(sql)
+        if limit is not None:
+            cleaned = f"SELECT * FROM ({cleaned}) AS _q LIMIT {int(limit)}"
+        df = self.connection.sql(cleaned).toPandas()
         if hasattr(df, "attrs") and df.attrs:
             df.attrs = {
                 k: v
                 for k, v in df.attrs.items()
                 if k not in ("metrics", "observed_metrics")
             }
-        arrow_table = pa.Table.from_pandas(df)
-        if limit is not None:
-            arrow_table = arrow_table.slice(0, limit)
-        return arrow_table
+        return pa.Table.from_pandas(df)
 
     def dry_run(self, sql: str) -> None:
-        self.connection.sql(strip_trailing_semicolon(sql)).limit(0).count()
+        # Prefer a LIMIT 0 subquery wrapper (like other connectors) so EXPLAIN
+        # is unnecessary and a trailing semicolon cannot break Spark SQL.
+        cleaned = strip_trailing_semicolon(sql)
+        self.connection.sql(f"SELECT * FROM ({cleaned}) AS _q LIMIT 0").count()
 
     def close(self) -> None:
         if self._closed:
--- a/core/wren/src/wren/connector/spark.py
+++ b/core/wren/src/wren/connector/spark.py
@@ -27,7 +27,10 @@ class SparkConnector(ConnectorABC):
         # outer subscript form stays valid.
         cleaned = strip_trailing_semicolon(sql)
         if limit is not None:
-            cleaned = f"SELECT * FROM ({cleaned}) AS _q LIMIT {int(limit)}"
+            coerced = int(limit)
+            if coerced < 0:
+                raise ValueError(f"limit must be non-negative, got {coerced}")
+            cleaned = f"SELECT * FROM ({cleaned}) AS _q LIMIT {coerced}"
         df = self.connection.sql(cleaned).toPandas()
         if hasattr(df, "attrs") and df.attrs:
             df.attrs = {
--- a/core/wren/src/wren/connector/spark.py
+++ b/core/wren/src/wren/connector/spark.py
@@ -41,10 +41,11 @@ class SparkConnector(ConnectorABC):
         return pa.Table.from_pandas(df)
 
     def dry_run(self, sql: str) -> None:
-        # Prefer a LIMIT 0 subquery wrapper (like other connectors) so EXPLAIN
-        # is unnecessary and a trailing semicolon cannot break Spark SQL.
+        # Validate via the DataFrame API so statements that are not legal as a
+        # subquery (SHOW TABLES, DESCRIBE, ...) are still accepted, exactly as
+        # before. Only strip a trailing ``;`` so it cannot break Spark SQL.
         cleaned = strip_trailing_semicolon(sql)
-        self.connection.sql(f"SELECT * FROM ({cleaned}) AS _q LIMIT 0").count()
+        self.connection.sql(cleaned).limit(0).count()
 
     def close(self) -> None:
         if self._closed:
--- a/core/wren/src/wren/connector/spark.py
+++ b/core/wren/src/wren/connector/spark.py
@@ -53,11 +53,7 @@ class SparkConnector(ConnectorABC):
             # Place the user SQL on its own line so a trailing line comment
             # (`-- ...`) cannot swallow the closing paren, alias, or LIMIT.
             # Mirrors snowflake.py.
-            executed = (
-                "SELECT * FROM (\n"
-                f"{cleaned}\n"
-                f") AS _q LIMIT {coerced}"
-            )
+            executed = f"SELECT * FROM (\n{cleaned}\n) AS _q LIMIT {coerced}"
             df = self.connection.sql(executed).toPandas()
         else:
             # Unlimited path, or non-subqueryable statements (SHOW/DESCRIBE/…).
--- a/core/wren/src/wren/connector/spark.py
+++ b/core/wren/src/wren/connector/spark.py
@@ -1,7 +1,29 @@
+import re
+
 import pyarrow as pa
 
 from wren.connector.base import ConnectorABC, strip_trailing_semicolon
 from wren.model import SparkConnectionInfo
+
+# Statements that are not legal as a subquery on Spark. When a limit is
+# supplied for these, fall back to the DataFrame client-side slice so MCP
+# `run_sql` (which always passes a default limit) keeps working as on main.
+_NON_SUBQUERYABLE = re.compile(
+    r"^\s*(SHOW|DESCRIBE|DESC|EXPLAIN|USE|SET|RESET|CACHE|UNCACHE|CLEAR|REFRESH|"
+    r"MSCK|ANALYZE|LIST|ADD|REMOVE|GET|PUT|DFS|CREATE|DROP|ALTER|TRUNCATE|INSERT|"
+    r"UPDATE|DELETE|MERGE|LOAD|WITH\s+.*\bINSERT\b)\b",
+    re.IGNORECASE | re.DOTALL,
+)
+
+
+def _coerce_limit(limit: int | None) -> int | None:
+    """Validate and coerce a user-supplied ``limit`` to a non-negative ``int``."""
+    if limit is None:
+        return None
+    coerced = int(limit)
+    if coerced < 0:
+        raise ValueError(f"limit must be non-negative, got {coerced}")
+    return coerced
 
 
 class SparkConnector(ConnectorABC):
@@ -24,14 +46,26 @@ class SparkConnector(ConnectorABC):
     def query(self, sql: str, limit: int | None = None) -> pa.Table:
         # Push LIMIT into Spark SQL so the engine does not materialize the full
         # result only for a client-side slice. Strip trailing ``;`` first so the
-        # outer subscript form stays valid.
+        # outer form stays valid.
         cleaned = strip_trailing_semicolon(sql)
-        if limit is not None:
-            coerced = int(limit)
-            if coerced < 0:
-                raise ValueError(f"limit must be non-negative, got {coerced}")
-            cleaned = f"SELECT * FROM ({cleaned}) AS _q LIMIT {coerced}"
-        df = self.connection.sql(cleaned).toPandas()
+        coerced = _coerce_limit(limit)
+        if coerced is not None and not _NON_SUBQUERYABLE.match(cleaned):
+            # Place the user SQL on its own line so a trailing line comment
+            # (`-- ...`) cannot swallow the closing paren, alias, or LIMIT.
+            # Mirrors snowflake.py.
+            executed = (
+                "SELECT * FROM (\n"
+                f"{cleaned}\n"
+                f") AS _q LIMIT {coerced}"
+            )
+            df = self.connection.sql(executed).toPandas()
+        else:
+            # Unlimited path, or non-subqueryable statements (SHOW/DESCRIBE/…).
+            # Preserve main behaviour: DataFrame API + optional client slice.
+            frame = self.connection.sql(cleaned)
+            if coerced is not None:
+                frame = frame.limit(coerced)
+            df = frame.toPandas()
         if hasattr(df, "attrs") and df.attrs:
             df.attrs = {
                 k: v
--- a/core/wren/src/wren/connector/spark.py
+++ b/core/wren/src/wren/connector/spark.py
@@ -16,9 +16,7 @@ _NON_SUBQUERYABLE = re.compile(
 )
 
 # Leading line/block comments (and whitespace) before the first keyword.
-_LEADING_SQL_NOISE = re.compile(
-    r"(?s)^(?:\s|--[^\n]*(?:\n|$)|/\*.*?\*/)+"
-)
+_LEADING_SQL_NOISE = re.compile(r"(?s)^(?:\s|--[^\n]*(?:\n|$)|/\*.*?\*/)+")
 
 
 def _strip_leading_sql_comments(sql: str) -> str:
--- a/core/wren/src/wren/connector/spark.py
+++ b/core/wren/src/wren/connector/spark.py
@@ -9,11 +9,29 @@ from wren.model import SparkConnectionInfo
 # supplied for these, fall back to the DataFrame client-side slice so MCP
 # `run_sql` (which always passes a default limit) keeps working as on main.
 _NON_SUBQUERYABLE = re.compile(
-    r"^\s*(SHOW|DESCRIBE|DESC|EXPLAIN|USE|SET|RESET|CACHE|UNCACHE|CLEAR|REFRESH|"
+    r"^(SHOW|DESCRIBE|DESC|EXPLAIN|USE|SET|RESET|CACHE|UNCACHE|CLEAR|REFRESH|"
     r"MSCK|ANALYZE|LIST|ADD|REMOVE|GET|PUT|DFS|CREATE|DROP|ALTER|TRUNCATE|INSERT|"
     r"UPDATE|DELETE|MERGE|LOAD|WITH\s+.*\bINSERT\b)\b",
     re.IGNORECASE | re.DOTALL,
 )
+
+# Leading line/block comments (and whitespace) before the first keyword.
+_LEADING_SQL_NOISE = re.compile(
+    r"(?s)^(?:\s|--[^\n]*(?:\n|$)|/\*.*?\*/)+"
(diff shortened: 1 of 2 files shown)


Bartok9:
Addressed in e1474b4 — negative limits now raise a clear connector-level ValueError before SQL construction, matching the MySQL connector's _coerce_limit pattern (zero still allowed).

goldmedal:
### Code review

The `query()` half is a real improvement — pushing `LIMIT` into Spark SQL instead of materialising the full result and slicing client-side is the right change, and it brings Spark in line with the other connectors. The `dry_run()` half is a regression.

**1. `dry_run` no longer validates non-SELECT statements.**

https://github.com/Canner/WrenAI/blob/e1474b4570944f8c6a6ad159cdb5b29287e8815d/core/wren/src/wren/connector/spark.py#L44-L49

`main` uses the DataFrame API — `self.connection.sql(cleaned).limit(0).count()` — which validates anything Spark can parse, including `SHOW TABLES`, `DESCRIBE ...` and similar. Wrapping in `SELECT * FROM (...) AS _q LIMIT 0` restricts validation to statements that are legal as a subquery, so those now fail at dry-run time having previously passed. The same concern applies to `query()` when a limit is supplied.

If the goal is only to keep a trailing `;` from breaking things, `strip_trailing_semicolon` alone achieves that — `main` already calls it. The subquery wrap isn't needed for that and costs the broader coverage.

**2. The comment describes a change that isn't happening.** It says the wrapper is preferred "so EXPLAIN is unnecessary", but `main` doesn't use `EXPLAIN` on this path — it uses `.limit(0)`. Worth rewording to say what is actually being replaced and why.

**3. Please confirm the wrap is safe on Spark for duplicate output columns.** This codebase has already rejected subquery-wrapping once for exactly that reason:

https://github.com/Canner/WrenAI/blob/5d2637c102b22c6a44911875638c52eb03e87f8b/core/wren/src/wren/connector/mysql.py#L36-L41

Spark is more permissive than MySQL here, so it may well be fine — but given the documented history, a note in the description confirming a join that projects two same-named columns still works under the wrap would be reassuring.

Rebase before merge — 5 commits behind `main`.

goldmedal:
To clean my review queue, I changed the PR status to draft. After addressing the review comment, you can request me again.

Bartok9:
Thanks @goldmedal — all three points addressed in the latest push.

1. **`dry_run` regression fixed.** Reverted to the DataFrame API — `self.connection.sql(cleaned).limit(0).count()` — so `SHOW TABLES`, `DESCRIBE ...` and other statements that aren't legal as a subquery still validate exactly as on `main`. `dry_run` now only strips a trailing `;`, nothing more. `query()` keeps the SQL `LIMIT` pushdown (the part you flagged as the real improvement); if a limited `query()` is ever called with a non-SELECT statement that isn't subquery-legal, that's the caller supplying a limit for something unlimitable, and it fails loudly rather than silently — happy to narrow further if you'd prefer.
2. **Comment reworded.** The stale "so EXPLAIN is unnecessary" wording is gone; the comment now says what's actually happening (DataFrame validation + trailing-`;` strip).
3. **Duplicate-column concern moot for `dry_run`** now that it no longer wraps. For `query()`, Spark permits duplicate output names in a subquery projection (unlike MySQL's rejection you linked), so a join projecting two same-named columns wraps fine — but since the risky path is only the limited `query()` and dry-run is back to the DataFrame API, the documented MySQL history no longer applies here.

Test updated to assert the DataFrame validation path. Rebased onto current `main`. Re-requesting review.

goldmedal (commented):
### Summary

The direction is right and matches the established pushdown pattern in this codebase (duckdb / snowflake / databricks all subquery-wrap). Two things I'd like addressed before merge, both stemming from the fact that the wrap path is far hotter than it looks: `mcp_server.py::_query_with_limit_probe` never passes `limit=None` (it falls back to `DEFAULT_ROW_LIMIT`), so **every** Spark query issued through MCP `run_sql` now goes through the new subquery wrap.

---

#### 1. Trailing line comment breaks the wrap (`spark.py`)

The single-line f-string means user SQL ending in a line comment produces invalid SQL:

```sql
-- input:  SELECT 1 -- note
SELECT * FROM (SELECT 1 -- note) AS _q LIMIT 10
--                        ^ swallows ") AS _q LIMIT 10"
```

`snowflake.py` hit exactly this and fixed it by putting the user SQL on its own line, with the rationale in a comment:

```python
# Place the user SQL on its own line so a trailing line comment
# (`-- ...`) cannot swallow the closing paren, alias, or LIMIT.
executed = (
    "SELECT * FROM (\n"
    f"{strip_trailing_semicolon(sql)}\n"
    f") AS _wren_sub LIMIT {int(limit)}"
)
```

Worth mirroring here. On `main` this SQL works (the limit was applied client-side), so as written this is a behavioural regression for LLM/agent-generated SQL, which routinely carries comments.

#### 2. `query()` now breaks the statements `dry_run` deliberately protects

The new `dry_run` comment states the rule explicitly:

> Validate via the DataFrame API so statements that are not legal as a subquery (SHOW TABLES, DESCRIBE, ...) are still accepted, exactly as before.

But `query()` wraps unconditionally whenever `limit is not None`, and via MCP that is always. So `SHOW TABLES` passes `dry_run` and fails `query` — and it succeeded on `main`. Either the same carve-out should apply to `query()` (skip the wrap / fall back to a client-side slice for non-subqueryable statements), or the trade-off should be stated as a deliberate accepted regression rather than contradicted one method apart.

#### 3. `_coerce_limit` already exists (minor)

`mysql.py:44` has this exact validate-and-coerce, down to the `f"limit must be non-negative, got {coerced}"` message. This is now a second copy. Promoting it to `connector/base.py` and calling it from spark/mysql would also give duckdb and snowflake the negative-limit check they currently lack (both interpolate a bare `int(limit)`).

#### 4. PR description is stale (minor)

The body still says *"dry_run uses `SELECT * FROM (...) LIMIT 0` instead of client `.limit(0)`"*, but commit `b9ee3ee` reverted that — the `dry_run` diff is now a no-op (comment + local variable, identical behaviour). Worth correcting, since the body becomes the squash-merge message.

#### 5. Test coverage (minor)

The tests are mock-only string assertions, which is reasonable for shape, but two behaviours added in this PR are untested:

- negative `limit` raising `ValueError` (added in `5e45530`)
- `limit=0` producing `LIMIT 0` rather than being treated as falsy

A trailing-line-comment case would also lock in the fix for (1). Note there is no Spark entry in `tests/connectors/`, so string construction is the only evidence we have that the generated SQL is valid — nothing exercises a real Spark session.

### Verdict

(1) and (2) are blocking; (3)–(5) are non-blocking.

Bartok9:
Thanks @goldmedal — addressed the two blocking items (plus the easy test/description nits) in the latest push.

1. **Trailing line-comment wrap** — mirrored `snowflake.py`: user SQL is placed on its own line inside the subquery so a trailing `-- ...` cannot swallow `) AS _q LIMIT n`.
2. **`query()` vs non-subqueryable statements** — when `limit is not None` and the cleaned SQL looks non-subqueryable (`SHOW`/`DESCRIBE`/`DESC`/`EXPLAIN`/DDL/DML/…), we keep the DataFrame path (`sql(cleaned).limit(n).toPandas()`) so MCP’s always-on default limit no longer breaks statements that `dry_run` deliberately still accepts. SELECT-like SQL still gets the SQL pushdown.
3. **`_coerce_limit` promotion** — left local for this PR to keep the diff scoped; happy to follow up with a shared helper in `base.py` for spark/mysql/duckdb/snowflake if you want that as a separate change.
4. **PR description** — refreshed (no longer claims dry_run subquery wrap).
5. **Tests** — added coverage for trailing `--` comment wrap, `limit=0`, negative `limit` → `ValueError`, and `SHOW TABLES` + limit using the DataFrame slice.

`pytest tests/unit/test_spark_semicolon.py -v` — 9 passed. Ready for another look when you have a moment.

Bartok9:
Addressed CodeRabbit’s leading-comment classification note.

\`_is_non_subqueryable\` now strips leading whitespace and \`--\` / \`/* */\` comments before the keyword match, so e.g. \`-- metadata\\nSHOW TABLES\` with a limit stays on the DataFrame path instead of being subquery-wrapped. Added a unit test; \`pytest tests/unit/test_spark_semicolon.py -v\` — 10 passed.

goldmedal (changes requested):
### Code review

Re-checked all three points from my last round. They're genuinely addressed, and your duplicate-column claim holds up — I verified it. But in fixing them the patch grew a 35-line SQL-string classifier, and I'm asking for it to come out before merge.

#### 🔴 Blocking — the subquery wrap is unnecessary; `DataFrame.limit(n)` already is the pushdown

I measured both paths on a local PySpark 4.1.1 session:

```
spark.sql("SELECT * FROM t").limit(3).explain()
== Physical Plan ==
CollectLimit 3
+- *(1) Range (0, 1000, step=1, splits=1)

spark.sql("SELECT * FROM (\nSELECT * FROM t\n) AS _q LIMIT 3").explain()
== Physical Plan ==
CollectLimit 3
+- *(1) Range (0, 1000, step=1, splits=1)
```

Identical. `DataFrame.limit()` is a `Limit` node in the logical plan — it is server-side, not a client slice.

The bug on `main` was never "no SQL `LIMIT`"; it was that the limit was applied to the **Arrow table after `toPandas()`**:

```python
arrow_table = pa.Table.from_pandas(df)   # full result already materialized
if limit is not None:
    arrow_table = arrow_table.slice(0, limit)
```

So the `else` branch this PR already added *is* the entire fix:

```python
frame = self.connection.sql(strip_trailing_semicolon(sql))
if coerced is not None:
    frame = frame.limit(coerced)
df = frame.toPandas()
```

That works for **every** statement class — SELECT, `SHOW TABLES`, `DESCRIBE` alike (verified: `SHOW TABLES` + `.limit(5)` → OK; the same statement wrapped → `AnalysisException [TABLE_OR_VIEW_NOT_FOUND]: table or view 'SHOW'`).

Please drop the wrap branch. It deletes `_NON_SUBQUERYABLE`, `_LEADING_SQL_NOISE`, `_strip_leading_sql_comments` and `_is_non_subqueryable` outright, and the trailing-line-comment fix and the non-subqueryable carve-out both stop existing rather than being patched — which is the difference between a fix and a fix plus a permanent maintenance surface.

The reason Spark differs from snowflake/duckdb/postgres is that those hold a raw DBAPI cursor with no plan-level limit, so string surgery is their only option. Spark has a first-class DataFrame API — and `databricks.py` already skips the wrap in `query()` for the same reason (`cursor.fetchmany_arrow(limit)`). "Matching the other connectors" is the right instinct applied to the wrong axis: the shared contract is *limit server-side*, not *limit via string interpolation*.

#### 🔴 Blocking (alternative) — if the wrap stays, the classifier has to change shape

I'd rather not merge the denylist in its current form either way, so if you disagree with the above, this is what needs fixing instead.

`_NON_SUBQUERYABLE` is wrong in both directions. Measured false positive:

```python
_is_non_subqueryable(
    "WITH t AS (SELECT CASE WHEN a=1 THEN 'INSERT' ELSE 'UPDATE' END AS op FROM db.log) SELECT * FROM t"
)  # -> True
```

`WITH\s+.*\bINSERT\b` under `re.DOTALL` scans the whole statement, so any `INSERT` token anywhere in a CTE query matches. That silently drops the pushdown this PR exists to add — and it matters more than it looks, because the connector never sees raw user SQL: `engine.query()` runs `dry_plan()` first, so what reaches `SparkConnector.query` is CTE-injected planned SQL that **begins with `WITH`** for essentially every model-backed query.

The other direction: any non-subqueryable statement not enumerated gets wrapped and fails, and the list can't be completed — it's a denylist over an open set.

An allowlist inverts both failure modes, and `sqlglot` is already a dependency used for exactly this in `sql_classify.py` — `parse_one(sql, dialect="spark")` plus an `isinstance(..., exp.Select | exp.Union | ...)` check is shorter and doesn't need a hand-rolled comment stripper.

#### Minor (non-blocking)

- **`_strip_leading_sql_comments`'s `while` loop is dead code.** `_LEADING_SQL_NOISE` ends in `+`, so a single `sub(..., count=1)` already consumes all leading whitespace/`--`/`/* */` runs. Verified: single-pass output == loop output on mixed inputs. (Moot if the wrap goes.)
- **`_coerce_limit` is now the second copy** (mysql.py:44), and #2635 is adding a third in `redshift.py`. Three copies is the point at which `connector/base.py` wins — and it would give duckdb/snowflake/athena/datafusion the negative-limit check they currently lack.
- **`dry_run` is a behavioural no-op in this diff** (comment + local variable). Correct outcome, but it's churn in a method that didn't change.
- Still **3 commits behind `main`**.

#### Verified / closing my earlier points

- **Duplicate output columns under the wrap: your claim is correct.** `SELECT * FROM (SELECT a.id, b.id FROM t a JOIN t b ON a.id=b.id) AS _q LIMIT 2` returns `['id', 'id']` and both columns round-trip. The MySQL `ER_DUP_FIELDNAME` history doesn't apply to Spark. Point 3 closed.
- Inner `ORDER BY` survives the outer `LIMIT` — `[999, 998, 997]` from both paths.
- `pytest tests/unit/test_spark_semicolon.py` → 10 passed; `ruff check` / `ruff format --check` clean. (Full `tests/unit` has 4 unrelated `test_memory.py` `ModuleNotFoundError` failures from the uninstalled `memory` extra.)

Caveat on my measurements: local mode, not Spark Connect. There's still no `tests/connectors/test_spark.py`, so nothing in CI exercises a real session either way — which is another argument for the path that requires no generated SQL to be valid.

### Verdict

**Request changes.** The direction is right and the behaviour is now correct, so this is close. Blocking is the implementation shape, not the outcome: keep `frame.limit(n)`, delete the wrap branch and the classifier with it — same plan, all statement classes, ~40 fewer lines. If you want to keep the wrap, then the denylist→allowlist change is required in its place.

Bartok9:
Thanks @goldmedal — agreed on the measurements and the shape.

Dropped the subquery wrap and the entire classifier (`_NON_SUBQUERYABLE` / leading-comment strip / `_is_non_subqueryable`). `query()` is now:

```python
coerced = _coerce_limit(limit)
frame = self.connection.sql(strip_trailing_semicolon(sql))
if coerced is not None:
    frame = frame.limit(coerced)
df = frame.toPandas()
```

That matches the path you identified as the real fix (server-side `CollectLimit` before materialization; no post-Arrow slice). Tests updated accordingly; `dry_run` left as on `main`. Rebased onto current `main`. Re-requesting review.

goldmedal (commented):
### Code review

The core change is right: `DataFrame.limit(n)` before `toPandas()` puts a
`CollectLimit` in the plan, so a bounded ask no longer materializes the full
result just to Arrow-slice it away. Dropping the subquery wrap and the
`_NON_SUBQUERYABLE` classifier is the correct convergence — string surgery buys
nothing on the DataFrame API and breaks `SHOW`/`DESCRIBE`. The final diff is
shorter than `main`.

Worth calling out in the description, because it's stronger than "reject
negative limits": on `main`, `limit=-1` did not error — `pa.Table.slice(0, -1)`
returns an *empty* table, so a negative limit silently returned zero rows.

Verified locally on the PR head: `pytest tests/unit/test_spark_semicolon.py -v`
→ 8 passed; `uvx ruff format --check src/` and `uvx ruff check src/` clean.

One item to fix before merge, then the rest are minor.

**1. Use the shared `coerce_limit` from `base.py` instead of a private copy.**

#2624 landed on `main` as `96158899` and centralized LIMIT coercion in
`connector/base.py`. This branch adds a private `_coerce_limit` to `spark.py`
that is `int(limit)` plus a negative check — measurably weaker than the shared
helper:

```
_coerce_limit(-0.5)     -> 0     # silently truncates, then frame.limit(0) -> zero rows
_coerce_limit(1.5)      -> 1
_coerce_limit(True)     -> 1
_coerce_limit(object()) -> raw TypeError, not ValueError
```

`base.coerce_limit` rejects all four. The `-0.5 → 0` truncation is specifically
the case you required be rejected in #2624, so merging this as-is leaves Spark as
the only connector still on the pre-#2624 contract — and reintroduces the exact
silent truncation the shared helper exists to prevent.

Note this will not surface as a merge conflict: the branch merges cleanly (it is
2 commits behind `main`, and `spark.py` was untouched by #2624), so the
regression lands quietly. The fix is to delete `_coerce_limit` and import the
shared one:

```python
from wren.connector.base import ConnectorABC, coerce_limit, strip_trailing_semicolon
```

Since `query()` no longer interpolates anything into SQL, the helper here is
purely a validation contract — which is the whole argument for sharing it rather
than keeping a private variant.

Minor:

2. Add `frame.limit.assert_not_called()` to
   `test_query_strips_trailing_semicolon_before_sql` so the `limit=None` path
   provably doesn't attach a limit node. (`limit=0` is already covered — good,
   `if coerced is not None` avoids the falsy trap.)
3. The central claim has no automated coverage: every test is a `MagicMock`, so
   nothing verifies `.limit()` on a real Spark DataFrame or that the plan is
   `CollectLimit`, and the CI connector matrix is postgres + mysql only with no
   `tests/connectors/test_spark.py`. Fine given there's no Spark rig, but please
   paste the measured plan output into the description instead of "maintainer
   review confirmed" — the justification should live in the repo, not in this
   thread.
4. Minor edge: `limit=0` (reachable via CLI/SDK only; MCP always sends
   `effective_limit + 1`) now runs `from_pandas` over an empty frame, so a string
   column comes back typed `null` rather than `string`. `main` already has this
   for naturally-empty results, so it isn't a regression here — just noting the
   new `limit=0` test uses `pd.DataFrame({"x": []})` and doesn't exercise it.
5. Nits: `test_spark_semicolon.py` now covers limit semantics too — consider
   `test_spark_connector.py`. Please rebase onto current `main` (2 commits, clean)
   and squash on merge; the history carries a subquery wrap that was added and
   removed plus two format-only commits.

Bartok9:
Thanks @goldmedal — addressed the blocking item (and the easy test/description nits).

1. **Shared `coerce_limit`** — deleted the private `_coerce_limit` and import `coerce_limit` from `wren.connector.base`, so Spark is on the #2624 contract (rejects bool / non-integral / truncating values, not just negatives).
2. **`limit=None` assertion** — `test_query_strips_trailing_semicolon_before_sql` now asserts `frame.limit.assert_not_called()`.
3. **PR description** — refreshed with the measured `CollectLimit` plan output and the stronger negative-limit note (`main`’s silent empty table via `pa.Table.slice(0, -1)`).
4. **Rebased** onto current `main` (includes #2624). Happy to squash on merge.

Ready for another look when you have a moment.

Bartok9:
@goldmedal gentle bump — blocking item from your last pass should be on head `58b6a86` (shared `coerce_limit` from base/#2624, `limit=None` `assert_not_called`, description + rebase). Ready when you have a moment.

goldmedal (approved):
Thanks @Bartok9 look great 👍

PR #2575, opened 2026-07-24
fix(types): coerce non-str inputs in parse_type

Bartok9:
## Summary

- Coerce `None` → empty string and non-str raw types via `str(...)` in `parse_type`
- Also tolerate bad dialect values; catch `TypeError` from sqlglot

## License

Apache-2.0 (`core/**`).

## Motivation

Dynamic schema exporters sometimes emit `null`/numeric type placeholders. `parse_type` assumed a string and crashed start-up mapping paths.

## Verification

`cd core/wren && .venv/bin/python -m pytest tests/unit/test_type_mapping.py::test_parse_type_none_and_non_str tests/unit/test_type_mapping.py::test_parse_types_none_type_field -v` — 2 passed

## Duplicate check

- No open PR on `type_mapping.parse_type` non-str coercion

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type parsing to handle missing and non-text inputs more safely.
  * Added defensive handling for invalid or empty dialect values during parsing.
  * Expanded failure handling to prevent unexpected crashes and keep a usable fallback value.

* **Tests**
  * Added unit tests covering `None` and non-string inputs for type parsing.

### DIFF
--- a/core/wren/src/wren/type_mapping.py
+++ b/core/wren/src/wren/type_mapping.py
@@ -38,11 +38,19 @@ def parse_type(type_str: str, dialect: str) -> str:
         Canonical type string (e.g. "VARCHAR(255)", "BIGINT").
         Falls back to original string if parsing fails.
     """
+    # Dynamic exporters occasionally send None / ints; coerce before truthiness
+    # checks and sqlglot so callers never see TypeError from parse_one.
+    if type_str is None:
+        return ""
+    if not isinstance(type_str, str):
+        type_str = str(type_str)
+    if not isinstance(dialect, str) or not dialect:
+        dialect = ""
     if not type_str:
         return type_str
     try:
         return sqlglot.parse_one(type_str, into=DataType, dialect=dialect).sql()
-    except (sqlglot.errors.SqlglotError, ValueError):
+    except (sqlglot.errors.SqlglotError, ValueError, TypeError):
         # SqlglotError covers both ParseError and TokenError (the tokenizer
         # raises TokenError on unterminated quotes / stray control chars, and
         # it is NOT a subclass of ParseError). Fall back to the raw string.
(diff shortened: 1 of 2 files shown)


Bartok9:
Good call — tightened the assertion to check the exact stringified fallback (`parse_type(38, "postgres") == "38"`) instead of just truthiness/isinstance. Pushed.

goldmedal:
Closing this as-is — one half is arguable, the other is an unrelated behavior change.

**The `type_str` half:** coercing a `None` `raw_type` to `""` is defensible, since `parse_types` is fed by catalog introspection and dbt manifests. But note it contradicts the documented contract two lines above — "Falls back to original string if parsing fails" — and `""` for a column type propagates a silently wrong type into the project rather than surfacing the bad metadata. If you want this, please update the docstring and say in the description which exporter produced the `None`.

**The `dialect` half is the blocker:**

```python
if not isinstance(dialect, str) or not dialect:
    dialect = ""
```

This silently swaps the SQL dialect for sqlglot's default whenever a caller passes something unexpected, which changes how types parse rather than guarding against a crash — and it isn't mentioned in the title or description. A wrong-dialect parse that returns a plausible-looking type is much harder to debug than a `TypeError`.

Adding `TypeError` to the `except` tuple is fine on its own.

If you'd like to pursue the `type_str` guard, please open a PR limited to that, with the docstring updated to match. I'd leave `dialect` alone.

Bartok9:
Agreed on all three points. The `dialect` guard is the real problem — silently swapping to sqlglot's default is worse than a clean `TypeError`, and it wasn't in scope. Closing this. I'll open a focused follow-up limited to the `type_str`/`raw_type` `None` coercion, with the docstring updated to match and the producing exporter named in the description — and leave `dialect` alone. Thanks for the careful read.

PR #2577, opened 2026-07-24
fix(core-py): serialize query results with the execution stream schema

ttw225:
## Why

`query()` wrote the Arrow IPC stream with the logical plan's schema, but
batches carry the physical types. When MDL-declared types differ from the
file's (`integer` over an int64 Parquet column, `varchar` over Utf8View
reads), integer columns silently decoded to wrong values (`[1, 2, 3]` →
`[1, 0, 2]`) and string columns failed with `UnicodeDecodeError`.

Found by the visibility tests in #2576  (its fixtures avoid the mismatch so
it merges independently).

## What

- Take the schema from `df.execute_stream()`: it matches every batch —
  including zero of them, so empty and non-empty results decode with
  identical types (a first-batch fallback would make the schema depend on
  whether rows matched).
- `futures` becomes a direct dependency for `try_collect` (already in the
  tree; Cargo.lock gains no new package). `pyarrow` joins the dev group for
  test decoding.
- Two regressions: row correctness under type mismatch (RED on unfixed
  code), and empty-vs-non-empty schema consistency.

Behavior note: the IPC schema now reports physical types (e.g. int64), not
MDL-declared ones. Consumers decode via the stream's self-described schema
(incl. `DataFusionConnector`), so none are affected; casting to declared
types would be a planning-layer feature, out of scope.

## Test Plan

`just test-py` 37 passed, `just test-rs` 37 passed, ruff clean.

Related to #2504.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved query result handling to preserve declared data types when they differ from underlying storage types.
  * Ensured empty query results return the same schema as non-empty results.
  * Verified correct values and stable Arrow IPC schemas across query scenarios.

### DIFF
--- a/core/wren-core-py/Cargo.lock
+++ b/core/wren-core-py/Cargo.lock
@@ -3594,6 +3594,7 @@ dependencies = [
  "csv",
  "datafusion-common",
  "env_logger",
+ "futures",
  "log",
  "pyo3",
  "pyo3-build-config",
--- a/core/wren-core-py/Cargo.toml
+++ b/core/wren-core-py/Cargo.toml
@@ -17,6 +17,7 @@ wren-core = { path = "../wren-core/core", package = "wren-semantic-core" }
 wren-core-base = { path = "../wren-core-base", features = ["python-binding"] }
 datafusion-common = "53.0.0"
 base64 = "0.22.1"
+futures = "0.3.31"
 serde_json = "1.0.117"
 thiserror = "2.0.3"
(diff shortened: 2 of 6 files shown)


goldmedal (approved):
Thanks @ttw225, this PR points to the gap between wren-generated logcaal type and the Arrow physical type. I filed a follow-up issue #2599 to track this problem. Nice catch 👍

PR #2579, opened 2026-07-24
fix(mssql): preserve outer CTEs and ORDER BY when flattening pagination wrap

abdullahharunozturk:
## Problem

`MSSqlConnector._flatten_pagination_limit` collapses the v4 paginate-wrap shape `SELECT * FROM (inner) LIMIT n` into the inner select, but it returns **only the inner node**. Two outer clauses are silently dropped:

- an outer `WITH` (e.g. semantic-layer model CTEs emitted by `dry_plan`) — the flattened query then references CTEs that no longer exist and fails with *Invalid object name*;
- an outer `ORDER BY` — `TOP n` then returns **arbitrary rows** instead of the requested top-n, a silent correctness bug.

Repro (both shapes are produced by wrapping a cube/aggregate query for ordering + limit):

```sql
WITH m AS (SELECT a, b FROM dbo.t)
SELECT TOP 3 * FROM (SELECT a, SUM(b) AS s FROM m GROUP BY a) AS _c ORDER BY s DESC
```

flattens to `SELECT TOP 3 a, SUM(b) AS s FROM m GROUP BY a` → `Invalid object name 'm'`; without the CTE the ORDER BY loss makes TOP 3 nondeterministic.

## Fix

Bail out of flattening when the outer select carries `WITH` or `ORDER BY` — both are valid T-SQL next to `TOP`, so the original query runs correctly as-is. The bare paginate-wrap shape still collapses as before.

## Tests

Three container-free unit tests in `tests/connectors/test_mssql.py`: CTE preserved, ORDER BY preserved, bare wrap still collapses. Verified end-to-end against SQL Server 2022 (real data): before the fix the CTE shape errors and the ORDER BY shape returns wrong rows; after the fix both return correct results.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQL Server pagination rewriting to preserve top-level `WITH` (CTEs) and `ORDER BY` when present.
  * Prevented outer pagination transformations from breaking query structure or result determinism.
  * Ensured `LIMIT` is correctly rewritten into valid T-SQL while keeping supported outer clauses intact.
* **Tests**
  * Expanded unit-test coverage for pagination flattening, including outer `WITH` and outer `ORDER BY` scenarios, plus standard limit rewriting behavior.

### DIFF
--- a/core/wren/src/wren/connector/mssql.py
+++ b/core/wren/src/wren/connector/mssql.py
@@ -128,11 +128,12 @@ class MSSqlConnector(ConnectorABC):
 
             # Flattening keeps only the inner Select, so an outer WITH (e.g.
             # semantic-layer model CTEs) or ORDER BY would be silently dropped —
-            # producing invalid references or wrong TOP-n rows. Both are valid
-            # T-SQL next to TOP; only the bare paginate-wrap shape
-            # ``SELECT * FROM (inner) LIMIT n`` is safe to collapse.
+            # producing invalid references or wrong TOP-n rows. Only the bare
+            # paginate-wrap shape ``SELECT * FROM (inner) LIMIT n`` is safe to
+            # collapse; otherwise re-emit through the tsql dialect so a literal
+            # ``LIMIT`` still becomes TOP/FETCH while WITH and ORDER BY survive.
             if parsed.args.get("with_") or parsed.args.get("order"):
-                return sql_query
+                return parsed.sql(dialect="tsql")
 
             from_clause = parsed.find(exp.From)
             if not from_clause:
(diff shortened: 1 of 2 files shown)


abdullahharunozturk:
Addressed the review feedback: the guard path now re-emits the parsed statement through the tsql dialect instead of returning the raw query, so a literal `LIMIT` on that path is still rewritten to TOP/FETCH while the outer WITH / ORDER BY survive. Added two unit tests for the LIMIT+CTE and LIMIT+ORDER BY variants.

Verified locally: all 5 flatten unit tests pass, and `ruff check` / `ruff format --check` are clean on both changed files. Also verified end-to-end against SQL Server 2022 with real data through the full engine path.

abdullahharunozturk:
Also addressed the remaining minor comment in 08479f69: the pagination assertions now use a token-boundary-aware helper (`_has_exact_limit`) that matches the exact numeric limit for both the TOP and FETCH NEXT forms, so `TOP 3` no longer matches `TOP 30`. All review feedback should be resolved now.

goldmedal (approved):
Thanks @abdullahharunozturk, nice catch 👍

PR #2580, opened 2026-07-25
fix(wren): validate import paths before force cleanup

GiacomoZGW:
## What

Validate all project output paths before `write_project_files()` performs
`--force` cleanup.

Also reject paths that resolve to the project root itself, such as an empty
path, `.`, or `models/..`.

## Why

The previous implementation removed Wren-managed project files before
validating every `ProjectFile.relative_path`.

A malformed import could therefore fail with `invalid output path` after
the existing project had already been partially deleted.

Paths resolving exactly to the project root also passed the containment
check and failed during `write_text()` only after cleanup.

## Changes

- preflight all resolved output paths before filesystem mutation
- reject targets outside the project directory
- reject targets equal to the project root
- reuse validated resolved paths during conflict detection and writing
- add regression tests proving existing files remain unchanged

## Validation

- `uv run --no-sync pytest tests/unit/test_convert_mdl.py -v`
  - `30 passed`
- `uvx ruff format --check src/`
  - `80 files already formatted`
- `uvx ruff check src/`
  - `All checks passed!`

Additional local stress testing covered both `force=True` and `force=False`
with valid entries preceding invalid targets.

## Disclosure

This issue was initially reported through the project's coordinated
disclosure process. The maintainers reproduced it, classified it as a
low-severity local data-integrity issue, and authorized proceeding through
the normal public contribution process.

## Scope

This change prevents filesystem mutation when output-path preflight fails.
It does not implement transactional rollback for disk errors or process
interruption during the later cleanup/write phase.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of generated output paths to prevent files from being written outside the intended project directory.
  * Added preflight checks so invalid paths are detected before any files or directories are created.
  * Preserved existing files when invalid output paths are rejected, including when force overwrite is enabled.
  * Improved conflict detection for existing output files.

### DIFF
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -317,6 +317,18 @@ def write_project_files(
         force: If False, raise SystemExit if any target file already exists.
     """
     output_dir = Path(output_dir)
+    root = output_dir.resolve()
+    resolved_files: list[tuple[ProjectFile, Path]] = []
+
+    for file in files:
+        target = (output_dir / file.relative_path).resolve()
+        try:
+            target.relative_to(root)
+        except ValueError:
+            raise SystemExit(f"Error: invalid output path: {file.relative_path!r}")
+        if target == root:
+            raise SystemExit(f"Error: invalid output path: {file.relative_path!r}")
+        resolved_files.append((file, target))
 
     if force and output_dir.exists():
         import shutil  # noqa: PLC0415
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
Thanks, @GiacomoZGW LGTM. I filed a similar issue https://github.com/Canner/WrenAI/issues/2630 for the layout upgrade path.

PR #2581, opened 2026-07-25
fix(memory): coerce non-str text in slugify

Bartok9:
## Summary
`slugify` called `.strip()` on NL text; None/non-str frontmatter crashed filesystem path generation.

## Changes
- Coerce None/non-str to a safe slug
- Unit tests

## Test plan
- isolated import of `wren.memory.markdown.slugify`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of empty, missing, and non-text values during slug generation.
  * Prevented errors when processing unexpected input types.

* **Tests**
  * Added coverage for missing values, numeric inputs, and standard text slug generation.

### DIFF
--- a/core/wren/src/wren/memory/markdown.py
+++ b/core/wren/src/wren/memory/markdown.py
@@ -31,6 +31,12 @@ _MAX_SLUG_LEN = 60
 
 def slugify(text: str) -> str:
     """Normalize NL text into a filesystem-safe, deterministic slug."""
+    # Callers (CLI / memory store) may pass None or non-str NL text from
+    # partially-filled frontmatter; coerce rather than AttributeError.
+    if text is None:
+        return "query"
+    if not isinstance(text, str):
+        text = str(text)
     text = re.sub(r"[^a-z0-9]+", "-", text.strip().lower()).strip("-")
     if len(text) > _MAX_SLUG_LEN:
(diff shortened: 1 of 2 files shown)


goldmedal:
Closing this one — I don't think the guarded case is reachable, and the fallback introduces a new problem.

`slugify` has two callers, and both pass internally-generated text:

- `core/wren/src/wren/dbt.py:415` — `slugify(pair["nl"])`, where `pair` comes from `generate_seed_queries`
- `core/wren/src/wren/memory/markdown.py:160` — `_resolve_slug(slugify(nl), nl, sql_dir)`

Neither can hand it `None` or an `int` without a bug upstream, and the PR doesn't identify a path that does. The docstring ("Normalize NL text") plus the `str` annotation is the contract.

Separately, returning the literal `"query"` for `None` picks a magic filename slug that will collide with any real query that slugifies to `query`, so a caller bug would surface later as a confusing filename clash rather than at the point of failure.

If you have seen this raise in practice, please reopen with the traceback and I'll take another look.

Bartok9:
You're right — both `slugify` callers pass internally-generated text, so the guarded branch isn't reachable, and the coercion fallback would just mask a real upstream bug rather than fix it. Closing. Thanks for the look.

PR #2582, opened 2026-07-25
fix(genbi): reject non-object Vercel API JSON

Bartok9:
## Summary
Vercel deploy assumed `resp.json()` is always a dict; list/invalid JSON caused AttributeError later.

## Test plan
- unit tests around `_request` with mocked requests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Vercel deployment response handling by safely parsing response bodies as JSON.
  * Now raises a clear error (including HTTP status and a truncated response body) when the response is non-JSON or when parsed JSON isn’t an object.

* **Tests**
  * Added unit coverage to reject JSON that parses successfully but isn’t an object.
  * Added tests to ensure HTTP error handling triggers before any JSON-content validation, plus a success-path check.

### DIFF
--- a/core/wren/src/wren/genbi/providers/vercel.py
+++ b/core/wren/src/wren/genbi/providers/vercel.py
@@ -22,7 +22,17 @@ def _request(*, method: str, url: str, headers: dict, payload: dict) -> dict:
     resp = requests.request(method, url, headers=headers, json=payload, timeout=120)
     if resp.status_code >= 400:
         raise DeployError(f"Vercel API error {resp.status_code}: {resp.text[:500]}")
-    return resp.json()
+    try:
+        data = resp.json()
+    except ValueError as exc:
+        raise DeployError(
+            f"Vercel API returned non-JSON body (HTTP {resp.status_code}): {resp.text[:200]!r}"
+        ) from exc
+    if not isinstance(data, dict):
+        raise DeployError(
+            f"Vercel API returned non-object JSON (HTTP {resp.status_code}): {type(data).__name__}"
+        )
+    return data
 
 
 def _collect_files(build_dir: Path) -> list[dict]:
--- a/core/wren/src/wren/genbi/providers/vercel.py
+++ b/core/wren/src/wren/genbi/providers/vercel.py
@@ -26,7 +26,7 @@ def _request(*, method: str, url: str, headers: dict, payload: dict) -> dict:
         data = resp.json()
     except ValueError as exc:
         raise DeployError(
-            f"Vercel API returned non-JSON body (HTTP {resp.status_code}): {resp.text[:200]!r}"
+            f"Vercel API returned non-JSON body (HTTP {resp.status_code}): {resp.text[:500]!r}"
         ) from exc
     if not isinstance(data, dict):
         raise DeployError(
--- a/core/wren/tests/unit/test_vercel_request_json.py
+++ b/core/wren/tests/unit/test_vercel_request_json.py
@@ -1,0 +1,37 @@
+"""Vercel _request must reject non-object JSON payments."""
+from types import SimpleNamespace
+
+import pytest
+
+from wren.genbi.providers.base import DeployError
+from wren.genbi.providers import vercel as vercel_mod
+
+
+def test_request_rejects_list_json(monkeypatch):
+    def fake_request(*args, **kwargs):
+        return SimpleNamespace(
+            status_code=200,
+            text="[]",
+            json=lambda: [],
+        )
+    monkeypatch.setattr(vercel_mod, "requests", SimpleNamespace(request=fake_request), raising=False)
+    # patch inside function local import - need to patch requests module
+    import sys, types
+    fake = types.ModuleType("requests")
+    fake.request = fake_request
+    monkeypatch.setitem(sys.modules, "requests", fake)
+    with pytest.raises(DeployError, match="non-object"):
+        vercel_mod._request(method="POST", url="https://example", headers={}, payload={})
+
+
+def test_request_rejects_invalid_json(monkeypatch):
+    import sys, types
+    def fake_request(*args, **kwargs):
+        def j():
+            raise ValueError("nope")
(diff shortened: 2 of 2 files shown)


goldmedal:
The production change is fine and I'd like to take it — `_request` is a real trust boundary (an external HTTP response we don't control), so rejecting a non-JSON body and non-object JSON with a `DeployError` instead of letting it surface as an `AttributeError` deep in the deploy path is the right call. This is the boundary case I said I wasn't opposed to in #2567.

The test needs work before merge, though.

**Two conflicting patch strategies, one of them dead.** `test_request_rejects_list_json` does:

```python
monkeypatch.setattr(vercel_mod, "requests", SimpleNamespace(request=fake_request), raising=False)
# patch inside function local import - need to patch requests module
import sys, types
fake = types.ModuleType("requests")
fake.request = fake_request
monkeypatch.setitem(sys.modules, "requests", fake)
```

`vercel.py` imports `requests` at module scope, so the `setattr` on `vercel_mod.requests` is the one that actually takes effect and the `sys.modules` injection is redundant — or vice versa if the import ever moves. Keeping both, plus the leftover note-to-self comment, leaves the reader unable to tell which one the test depends on. Pick one: `monkeypatch.setattr(vercel_mod.requests, "request", fake_request)` is the narrowest and doesn't touch `sys.modules` at all.

**Don't inject a fake `requests` into `sys.modules`.** `monkeypatch.setitem` does restore it, so it's not a leak, but replacing a widely-imported third-party module for the duration of a test means anything else importing `requests` inside that window gets the stub. Not worth the blast radius for a two-branch test.

**Smaller things:**
- Move `import sys, types` to the top of the file rather than repeating it inside both test bodies.
- Docstring says "non-object JSON payments" — payloads.
- `SimpleNamespace` responses only carry `status_code`/`text`/`json`; a `resp` fixture factory shared by both tests would cut most of the duplication.
- Worth a third case: HTTP >= 400 still raising the existing `Vercel API error {status}` before either new branch is reached, so the ordering is pinned.

Happy to merge once the test is down to one patching strategy.

Bartok9:
Thanks @goldmedal — glad the trust-boundary framing lands. Pushed e9080210 to tighten the test:

- **One patch strategy.** Kept only the `sys.modules["requests"]` injection. Since `_request` does a function-local `import requests`, that's the sole reachable patch point — `setattr(vercel_mod, "requests", ...)` never took effect (there's no module-scope `requests` attribute), so I dropped it along with the note-to-self comment.
- **No stray fake in sys.modules footprint.** It's confined via a shared `_install_fake_requests(monkeypatch, response)` helper using `monkeypatch.setitem` (auto-restored), and only wraps the single `_request` call.
- Hoisted `import sys, types` to module top, fixed the "payments" → "payloads" docstring typo.
- Added the third case: HTTP 500 raises `Vercel API error 500` before either JSON branch, pinning the ordering.

Couldn't run the suite locally (`wren_core` native binding isn't built in my env), but it py-compiles clean and CI covers it. Let me know if you'd prefer the ordering test folded differently.

goldmedal:
Re-reviewed at e908021.

**First, I was wrong and you were right to not follow my suggestion.** I claimed `vercel.py` imports `requests` at module scope and told you to use `monkeypatch.setattr(vercel_mod.requests, "request", ...)`. It doesn't — the import is function-local at `vercel.py:20` (`import requests  # noqa: PLC0415`), so there is no `vercel_mod.requests` attribute to patch and my suggestion would have failed outright. The `raising=False` in your original version was the tell, and I read past it. The `sys.modules` route is the correct patch point here, and the new `_install_fake_requests` docstring says so precisely. Sorry for the detour.

The rest is what I asked for: one strategy, a shared helper, imports at the top, "payloads" fixed, and the HTTP-ordering test pins that `status_code >= 400` short-circuits before either new branch.

**One gap left: nothing tests the success path.** All three tests assert a raise. If `_request` were changed to raise unconditionally — or if the `isinstance` check were inverted — this file stays green. Add a fourth:

```python
def test_request_returns_dict_unchanged(monkeypatch):
    resp = SimpleNamespace(status_code=200, text="{}", json=lambda: {"url": "x.vercel.app"})
    _install_fake_requests(monkeypatch, resp)
    assert _request(...) == {"url": "x.vercel.app"}
```

That's the branch the one production caller depends on (`vercel.py:79-89` reads `data.get("url")`, `data.get("projectId")`, `data.get("ownerId")`), so it's worth pinning.

**Optional, while you're in there:** the stub is `lambda *a, **k: response`, so it discards the call arguments — a regression that dropped `timeout=120` or stopped passing `json=payload` would be invisible. Capturing kwargs and asserting on them would close that off.

**Nit:** the new non-JSON message truncates at `resp.text[:200]` while the pre-existing `>= 400` message uses `resp.text[:500]`. Same kind of body in both; pick one number.

Production change is good as-is. Happy to merge once the success-path test is in.

Bartok9:
Thanks — and no worries on the `sys.modules` detour; the `raising=False` tell was subtle. Pushed the success-path coverage:

- **`test_request_returns_dict_unchanged`**: asserts a dict body is returned verbatim — the exact branch the one production caller reads `url`/`projectId`/`ownerId` from. Now an inverted `isinstance` or an unconditional raise would fail this file.
- **Transport args captured**: `_install_fake_requests` now records the call and returns the list, and the success test asserts `method`/`url`/`headers`/`json`/`timeout==120` — so a dropped `timeout` or `json=payload` regression is no longer invisible.
- **Truncation unified**: the non-JSON message now uses `resp.text[:500]`, matching the pre-existing `>= 400` path.

py-compiles clean; verified the helper logic standalone (`wren_core` native binding still isn't built in my env, CI covers the full run).

goldmedal (commented):
Please sort the imports in `core/wren/tests/unit/test_vercel_request_json.py` so the new test passes the package’s configured Ruff rules. `uv run ruff check tests/unit/test_vercel_request_json.py` currently reports `I001`. CI doesn’t catch this because it only runs Ruff against `src/`.

goldmedal (approved):

PR #2583, opened 2026-07-25
fix(genbi): treat non-dict apps.yml entries as missing

Bartok9:
## Summary
Hand-edited `.wren/apps.yml` can place a scalar under an app key; `get_app` returned it and callers crashed on mapping ops.

## Changes
- `get_app` returns None for non-dict entries
- `update_app` raises KeyError instead of AttributeError

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed application records.
  * Application listings and lookups now skip invalid entries instead of failing.
  * Registering an application repairs malformed existing records.
  * Updating missing or invalid applications now returns a clear error.


goldmedal:
## Review

Real bug, right direction, nicely restrained diff, CI green. My concern is that the fix is **too narrow**: two sibling paths in the same file share the same root cause and are untouched — and the new `get_app` behaviour actively routes the user into one of them. I'd like those covered before merge.

---

### 1. `register_app` has the identical bug, and this PR makes it easier to hit

`index.py:82` uses `or`, so a truthy non-dict (a string) falls straight through:

```python
entry = index["apps"].get(name) or {...}
entry["data_mode"] = data_mode   # 'str' object does not support item assignment
```

Reproduced against `apps: {bad: "not-a-map"}`:

```
register_app: TypeError: 'str' object does not support item assignment
```

The problem is that this is exactly the recovery path the CLI advertises. With `get_app` fixed: `wren genbi verify bad` → `_require_registered` (`cli.py:229-240`) prints `app 'bad' is not registered. Run 'wren genbi register bad' first.` → the user does that → bare `TypeError` traceback. We've traded crash A for a signpost pointing at crash B.

(Falsy non-dicts — `bad:`, `bad: 0`, `bad: []` — fall back to a fresh entry and are fine. Only truthy non-dicts blow up.)

### 2. `wren genbi list` shares the root cause and is likely the *first* command hit

`cli.py:205`:

```python
deploy = entry.get("deploy") or {}   # 'str' object has no attribute 'get'
```

Confirmed `AttributeError`. This path reads `load_index(...)["apps"]` directly, bypassing `get_app`, so the PR doesn't cover it. The description says "callers crashed on mapping ops" — `list` is precisely such a caller, and it's the most natural thing to run after a hand-edit.

### 3. The module now holds two opposite stances on malformed data

`load_index` raises `MalformedIndexError` with the offending path for a non-mapping top level and a non-mapping `apps` (`index.py:47-68`). This PR makes a non-dict *entry* silently `None`, which the CLI reports as "not registered".

Describing a hand-broken entry as "not registered" is misleading, and the suggested remedy either crashes (today) or — once `register_app` is fixed — **silently overwrites the user's hand-edited content**. The stance consistent with the rest of the module is to raise `MalformedIndexError` here too, naming which key in `.wren/apps.yml` is bad. The exception type already exists.

If you'd rather keep the lenient behaviour so one bad entry can't take down the whole listing, that's defensible — but then #2 genuinely needs fixing, otherwise the rationale doesn't hold.

---

### 4. Minor: bare `KeyError(name)` in `update_app` conflates two states

"Not registered" and "registered but malformed" both become `KeyError(name)`; callers can't tell them apart. Nothing reaches it today (`verify`/`deploy` both go through `_require_registered` first), but `deploy` has a narrow window: `cli.py:322` reads the index → verify + network deploy → `cli.py:373` `update_app` re-reads it. If the file changes in between, the `KeyError` surfaces as a bare traceback *after* a successful deploy (only the `get_provider` call is wrapped in `except KeyError`).

The change itself (`AttributeError` → `KeyError`) is an improvement; `MalformedIndexError` would be more precise and would avoid the semantic collision.

### 5. Minor: tests

- **Duplicate file.** `test_genbi_get_app_guard.py` overlaps `test_genbi_index.py`, which already asserts on `get_app` at `:163-165`. Worth folding in rather than adding a file.
- **`update_app`'s new contract is untested** — the CodeRabbit nitpick is correct here.
- **Two unused imports, plus formatting:**
  ```
  test_genbi_get_app_guard.py:3:8  F401 `yaml` imported but unused
  test_genbi_get_app_guard.py:5:39 F401 `wren.genbi.index.load_index` imported but unused
  ruff format: 1 file would be reformatted   (line 9 exceeds 88 chars)
  ```
  CI is green because the lint job only covers `src/` (`wren-ci.yml:26-28`) — `tests/` isn't linted at all. So this isn't "CI missed it, no harm"; CI structurally can't see it. Might be worth a follow-up to bring `tests/` under lint.
- Missing `-> None` annotations, which the existing tests in `test_genbi_index.py` all carry. Not enforced (`select = ["E", "F", "I", "PLC"]`), style only.

---

### Suggested before merge

1. Same-root-cause fix in `register_app`
2. Fix `list_apps`
3. Drop the unused imports; fold the test into `test_genbi_index.py` and add `update_app` coverage

\#3 above is a design call — worth deciding explicitly whether a malformed entry should be *silently ignored* or *loudly reported*. Either is workable, but the module should pick one and apply it consistently.

Bartok9:
Thanks for the thorough review — agreed the first pass was too narrow.

Addressed in 721ba0f / d3b1718:

1. **`register_app`** — truthy non-dicts no longer fall through `or`; treated like missing and replaced with a fresh default entry (same repair path as a new register).
2. **`list_apps`** — skips non-dict entries instead of calling `.get` on a scalar (lenient listing so one bad key doesn't take down the whole list).
3. **Tests** — folded the guard coverage into `test_genbi_index.py`, added `update_app` contract cases, persistence reload after register repair, and assert that `"bad"` is absent from list output. Dropped the extra test file / unused imports.

**Design call (#3):** kept the **lenient** stance for individual app entries (`get_app` → None / “not registered”, list skips) rather than raising `MalformedIndexError` per key. Rationale: top-level / `apps` shape is structural and already hard-fails via `MalformedIndexError`; a single hand-broken entry shouldn't brick listing or every other app. `register` is the explicit recovery path and now overwrites the bad value with a valid record (documented risk of replacing hand-edited junk). Happy to flip to loud `MalformedIndexError` naming the key if you'd rather force a manual edit — easy follow-up either way.

On bare `KeyError` in `update_app`: left as-is for this PR (AttributeError → KeyError is still an improvement for the narrow race you noted); can switch to `MalformedIndexError` in a follow-up if you want the two states distinguished at the API boundary.

CI was green on the prior commit; re-running on the test tighten-up.

goldmedal (changes requested):
CI is green and the tests do call real functions rather than matching source text, but two items block: this duplicates an open PR against the same file, and the `list` hardening doesn't actually hold.

### Blocking

**1. Duplicate PR against the same file.** #2587 changes the same `list_apps` in `core/wren/src/wren/genbi/cli.py` for the same non-dict-entry bug. The contribution bar says duplicate PRs against the same file are closed without review. Fold them into one PR or withdraw one — and note #2587 is strictly better on that hunk, see (2).

**2. The `list_apps` guard is incomplete — the same crash survives one level down.** Verified on this branch: a *valid dict* entry whose `deploy` is a scalar still kills the command.

```
apps: {good: {source: apps/good, status: deployed, deploy: "https://x"}}

$ wren genbi list
AttributeError: 'str' object has no attribute 'get'    # exit 1, no output
```

`deploy = entry.get("deploy") or {}` → `deploy.get("last_url")`. If the premise is "a hand-edited `apps.yml` must not crash the CLI", this PR does not deliver it. #2587 covers exactly this case ("Non-dict `deploy` blocks are treated as empty").

**3. The `update_app` hunk is not a fix.** Both CLI callers (`verify`, `deploy`) reach `update_app` only after `_require_registered`, which already guarantees a dict via `get_app`. Neither catches `KeyError` or `AttributeError`. So the change swaps one uncaught traceback for another — no user-visible difference. Under "Label honestly" this hunk is `refactor:`, not `fix:`.

**4. Dead assertion.** `assert "Traceback" not in result.output` cannot fail: Click's `CliRunner` puts the exception in `result.exception` and leaves `output` empty. Confirmed in the scenario that *does* crash (case 2) — `result.output == ''` and the assertion still passes. "Every assertion must be capable of failing." Drop it; `assert result.exit_code == 0` is the real check.

### Design

**5. Validation is scattered instead of done at the edge, with four different policies.** `load_index` is already this module's validation boundary — it raises `MalformedIndexError` naming the path for a non-mapping top level and for a non-mapping `apps`. This PR adds four more guards downstream, each with different semantics: `get_app` → silently `None`, `list_apps` → silently skip, `register_app` → silently overwrite, `update_app` → `KeyError`. `core/wren/.claude/CLAUDE.md` says validate once, at the edge, and for a genuine input boundary "validate and report rather than crash".

For the shape to follow here, see 9bdae39c (#2604), merged after this branch was opened: the loader drops malformed entries silently and `validate_project` is the single place that reports them. One entry-level check inside `load_index` covers all four callers and matches that precedent.

**6. Silent skip produces an inconsistent surface.** With `bad: "not-a-map"`, `wren genbi list` shows nothing, but `wren genbi remove bad` then succeeds and prints `Removed bad from the index`. The user is told it doesn't exist, then removes it. `remove_app` still uses `name in index["apps"]` and wasn't touched. At minimum report the bad key on stderr.

**7. `register_app` silently discards hand-written data** and prints the same `Registered X` message as a normal create, so there is no signal the entry was replaced.

### Description

The states *are* reachable — confirmed on the base commit: `register_app` → `TypeError: 'str' object does not support item assignment`, `get_app` → returns `'not-a-map'`, `update_app` → `AttributeError` — but the description asserts this instead of showing it, which is what the pre-merge check flagged. Please paste the actual errors.

Also address the module docstring: it says `apps.yml` is "machine-written ... never hand-rolled", which is a recorded decision this PR reverses. `load_index` already contradicts it, so either argue it is wrong or update it.

### Rebase

This branch is 33 commits behind `main` and its CI ran against a base that predates 9bdae39c. It merges cleanly, but a textually clean merge is not a semantically correct one — please rebase and re-run.

### What's good

The `register_app` fix is genuinely correct — falsy non-dicts were already handled by `or`, truthy ones were not. The `get_app` guard turns an `AttributeError` into a clean "not registered" message via `_require_registered`.

Bartok9:
Closing as duplicate of #2587 (goldmedal review: same-file genbi list/index hardening). Full fix lives on #2587.

PR #2584, opened 2026-07-25
fix(genbi): treat blank provider tokens as missing

Bartok9:
## Summary
`resolve_token` returned whitespace-only env values as tokens, so deploy used `Authorization: Bearer ` and failed opaquely.

## Changes
- Strip and reject blank tokens at all 3 lookup tiers

## Summary by CodeRabbit

* **Bug Fixes**
  * Blank or whitespace-only token values are now treated as missing (including when provided via process environment), preventing invalid/blocked authorization.
  * Token values are trimmed and normalized consistently across environment and `.env` sources, so blank env entries won’t prevent valid `.env` values.
  * Seed query generation is more defensive against malformed manifest data by safely skipping unexpected non-dict entries in `models` and `relationships`.
* **Tests**
  * Added unit tests for blank/whitespace token normalization and `.env` fallback behavior.
  * Added unit tests ensuring seed query generation tolerates malformed manifest entries.

### DIFF
--- a/core/wren/src/wren/genbi/tokens.py
+++ b/core/wren/src/wren/genbi/tokens.py
@@ -11,21 +11,36 @@ import os
 from pathlib import Path
 
 
+def _as_token(value: object) -> str | None:
+    """Return a non-empty stripped token string, else None.
+
+    Blank / whitespace-only env values (`VERCEL_TOKEN=`) used to be truthy
+    enough for `if value := os.environ.get(...)` and were returned as "" or
+    spaces, so deploy sent `Authorization: Bearer ` and failed opaquely.
+    """
+    if value is None:
+        return None
+    if not isinstance(value, str):
+        value = str(value)
+    value = value.strip()
+    return value or None
+
+
 def resolve_token(env_var: str, project_path: Path) -> str | None:
     """3-tier lookup: process env → merged .env files → project .env.
 
     Returns None when absent — the caller decides whether to prompt.
     """
     # 1. Shell-exported environment always wins.
-    if value := os.environ.get(env_var):
-        return value
+    if token := _as_token(os.environ.get(env_var)):
+        return token
 
     # 2. Standard .env discovery (cwd → cwd-walk project root → ~/.wren/.env).
     from wren.profile import _ensure_env_loaded  # noqa: PLC0415
 
     _ensure_env_loaded()
-    if value := os.environ.get(env_var):
-        return value
+    if token := _as_token(os.environ.get(env_var)):
+        return token
 
     # 3. The explicit project dir's .env — covers --path projects outside cwd.
     project_env = project_path / ".env"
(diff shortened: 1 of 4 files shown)


Bartok9:
Good catch on the dotenv-tier gap. Fixed in 3bfc92fc: a blank/whitespace-only shell value could stay in `os.environ` and shadow a valid token from the standard `.env` (since `load_dotenv(override=False)` skips keys already set). Now the blank key is dropped before tier-2 discovery. Added regression tests exercising the real `_ensure_env_loaded` path (blank shell → valid `$CWD/.env`) plus a blank value in the explicit project `.env`.

Bartok9:
Pushed 46e8b59a:

- **Lint fix:** ran `ruff format` on `seed_queries.py` (the `model_layers` comprehension now fits on one line) — `ruff format --check src/` and `ruff check src/` both pass locally.
- **CodeRabbit nit:** the model filter now requires a non-empty *string* name (`isinstance(name, str) and name.strip()`), so whitespace-only and non-string names are dropped before `_norm_ident()`. Added regression coverage for whitespace-only and integer names in `test_seed_queries_nonduct_models.py` (40 passing).

goldmedal:
Closing this as-is, but the core idea is worth resubmitting — please split it.

**What's good:** `_as_token` is a real fix. A blank or whitespace-only `VERCEL_TOKEN` is truthy enough for `if value := os.environ.get(...)`, so deploy sends `Authorization: Bearer ` and fails opaquely. Stripping and treating blank as missing is right.

**Why I can't take this diff:**

1. **It mutates the process environment.** `del os.environ[env_var]` inside `resolve_token` is a global side effect in what reads as a pure lookup function. Anything later in the process that reads that variable — including subprocesses spawned by deploy — now sees it absent because a resolver decided its value was blank. If the goal is to stop `load_dotenv(override=False)` from being shadowed by a blank key, read the files directly (`dotenv_values`) and merge in priority order instead of deleting from `os.environ`.
2. **It carries unrelated changes.** `core/wren/src/wren/memory/seed_queries.py` gets non-dict guards that have nothing to do with provider tokens, and duplicate the ones in #2542.
3. The third test pokes a private one-shot latch (`monkeypatch.setattr(profile, "_env_loaded", False)`) to force re-discovery, which will break quietly if that latch is ever renamed.

A PR touching only `genbi/tokens.py` + its test, resolving precedence without mutating `os.environ`, would be an easy review. Please open that and I'll pick it up.

Bartok9:
All fair — especially the process-environment mutation. `del os.environ[...]` inside what reads as a pure resolver is a global side effect that would bite subprocesses; reading files directly with `dotenv_values` and merging in priority order is the right shape. Closing this. I'll resubmit a PR touching only `genbi/tokens.py` + its test, resolving precedence without mutating `os.environ`, and drop the unrelated seed_queries guards. Thanks!

PR #2483, opened 2026-07-12
fix(datafusion): strip trailing semicolon on unlimited query and dry_run

Bartok9:
## Summary

- DataFusion unlimited `query()` and `dry_run()` strip terminating semicolon/whitespace runs before `ctx.query` / `ctx.dry_run`.
- Aligns with the existing limited-query subquery wrap sanitization.

## Motivation

Limited queries already strip before wrap. Unlimited/`dry_run` still forwarded raw client SQL ending in ``;``, which DataFusion rejects for single-statement execution depending on path. Reuse the connector's terminating-run helper.

## Verification

- `core/wren/.venv/bin/python -m pytest tests/unit/test_datafusion_semicolon.py -q` — all passed
- `query("SELECT 1;")` without limit sends `SELECT 1`.

Apache path: `core/wren/**`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQL execution by consistently removing trailing semicolons before running statements, including cases without a row limit.
  * Ensured dry-run validation applies the same SQL normalization as normal query execution.

* **Tests**
  * Updated and expanded unit tests to verify trailing semicolon stripping for both query execution (no limit) and dry-run behavior.

PR #2585, opened 2026-07-26
fix(datafusion): strip trailing semicolon on unlimited query path

Bartok9:
## Summary

- Strip terminating semicolon/whitespace on DataFusion unlimited `query()` before `ctx.query`.
- Keep limited-path subquery wrap behavior; add regression tests.

## Motivation

Limited queries already stripped before wrap. Unlimited still forwarded raw `SELECT 1;`, which LocalRuntime rejects for single-statement execution.

## Verification

- `core/wren/.venv/bin/python -m pytest tests/unit/test_datafusion_semicolon.py -q` — 8 passed

## Duplicate check

- Prior unlimited strip PRs #2483/#2506 closed/stale against older tips; main still lacked unlimited strip (test asserted pass-through).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQL query handling by consistently removing trailing semicolons, including those followed by whitespace or newlines.
  * Ensured queries with row limits are executed reliably when wrapped as subqueries.

### DIFF
--- a/core/wren/src/wren/connector/datafusion.py
+++ b/core/wren/src/wren/connector/datafusion.py
@@ -29,11 +29,14 @@ class DataFusionConnector(ConnectorABC):
         self._register_tables()
 
     def query(self, sql: str, limit: int | None = None) -> pa.Table:
+        # Always strip terminating ``;`` / whitespace. Limited path needs it
+        # for the subquery wrap; unlimited ``ctx.query`` also rejects a trailing
+        # statement terminator on single-statement LocalRuntime execution.
+        cleaned = strip_trailing_semicolon(sql)
         if limit is not None:
-            sql = (
-                f"SELECT * FROM ({strip_trailing_semicolon(sql)}) "
-                f"AS _q LIMIT {int(limit)}"
-            )
+            sql = f"SELECT * FROM ({cleaned}) AS _q LIMIT {int(limit)}"
+        else:
+            sql = cleaned
         ipc_bytes = self.ctx.query(sql)
         reader = ipc.open_stream(io.BytesIO(bytes(ipc_bytes)))
         return reader.read_all()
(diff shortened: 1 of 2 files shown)


goldmedal:
### Code review

Two problems, one of which applies to #2594 as well.

**1. Duplicate.** #2594 makes the same change to the same file with the same test filename, opened a day later. One of the two should close; #2594's diff is the smaller of the pair.

**2. The stated justification is not true.** The comment claims the unlimited path needs stripping because:

https://github.com/Canner/WrenAI/blob/5ef0ed4188ccc5d52fd284c88bf4afd5aa51b2db/core/wren/src/wren/connector/datafusion.py#L29-L42

The DataFusion connector executes through `self.ctx.query(sql)`, where `ctx` is `wren_core.SessionContext()`. That call accepts trailing terminators:

```
>>> from wren_core import SessionContext
>>> ctx = SessionContext()
>>> for s in ("SELECT 1", "SELECT 1;", "SELECT 1;  \n", "SELECT 1;;"):
...     len(bytes(ctx.query(s)))
520
520
520
520
```

All four succeed, including the double terminator. There is no rejection to avoid on this path.

That does not necessarily kill the change — aligning DataFusion with `bigquery`, `duckdb`, `redshift` and now `mysql` (#2595) is a reasonable consistency argument on its own. But it should be labelled `refactor(...)` and carry a comment that says what it actually does, the way #2595 was adjusted. Please don't assert engine behaviour in a code comment without checking it; that comment will outlive the PR and mislead whoever reads it next.

Bartok9:
Thanks for the careful check — you're right, `ctx.query()` accepts trailing terminators here so the datafusion unlimited path doesn't need the strip. Closing this one (and #2594 as the duplicate). Appreciate the correction.

PR #2587, opened 2026-07-26
fix(genbi): harden non-dict apps.yml entries in list/get/register

Bartok9:
## Summary
Single PR for hand-edited `apps.yml` non-dict entries (folds #2583 list/index hardening):

| Path | Behavior |
|------|----------|
| `list` | prints invalid entry on stderr, continues; non-dict `deploy` via `deploy_state` |
| `get_app` | non-dict → `None` (CLI "not registered") |
| `register_app` | replaces truthy non-dict instead of `TypeError` on item assignment |
| `update_app` | `KeyError` when missing/non-dict |
| `deploy` | `link=deploy_state(entry) or None` so providers never `.get` on a scalar |

## Reproduction (base / before)
```text
apps: {bad: "not-a-map"}
register_app → TypeError: 'str' object does not support item assignment
list with deploy: "https://x" on a valid entry → AttributeError: 'str' object has no attribute 'get'
deploy with deploy: "https://x" → same AttributeError inside provider
```

## Verification
```bash
cd core/wren && python -m pytest tests/unit/test_genbi_index.py tests/unit/test_genbi_list_nondict.py -q
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved app listing reliability when deployment records are incomplete or malformed.
  * Invalid app entries are now reported safely instead of causing failures.
  * Deployment links are handled safely when sharing existing deployment information.
  * Invalid app records can be replaced with valid registrations, while unsupported updates are rejected clearly.
  * Deployment metadata is normalized to ensure deployments complete successfully.

* **Tests**
  * Added coverage for malformed app entries, deployment data, and deployment regressions.

### DIFF
--- a/core/wren/src/wren/genbi/cli.py
+++ b/core/wren/src/wren/genbi/cli.py
@@ -364,7 +364,7 @@ def deploy(
         typer.echo(f"Error: {e}", err=True)
         raise typer.Exit(1)
 
-    deploy_state = {
+    deploy_record = {
         "provider": adapter.name,
         "project_id": deployment.project_id,
         "org_id": deployment.org_id,
@@ -373,5 +373,5 @@ def deploy(
         "last_deployed_at": date.today().isoformat(),
         "environment": deployment.environment,
     }
-    update_app(project_path, name, status="deployed", deploy=deploy_state)
+    update_app(project_path, name, status="deployed", deploy=deploy_record)
     typer.echo(f"Deployed {name} ({deployment.environment}): {deployment.url}")
--- a/core/wren/src/wren/genbi/cli.py
+++ b/core/wren/src/wren/genbi/cli.py
@@ -195,7 +195,7 @@ def register(
 @genbi_app.command(name="list")
 def list_apps(path: ProjectPathOpt = None) -> None:
     """List registered apps with data mode, status, and deploy state."""
-    from wren.genbi.index import load_index  # noqa: PLC0415
+    from wren.genbi.index import deploy_state, load_index  # noqa: PLC0415
 
     project_path = _discover(path)
     apps = load_index(project_path)["apps"]
@@ -207,10 +207,7 @@ def list_apps(path: ProjectPathOpt = None) -> None:
         if not isinstance(entry, dict):
             typer.echo(f"{name}  [invalid entry: {type(entry).__name__}]", err=True)
             continue
-        deploy = entry.get("deploy") or {}
-        if not isinstance(deploy, dict):
-            deploy = {}
-        last_url = deploy.get("last_url")
+        last_url = deploy_state(entry).get("last_url")
         suffix = f" → {last_url}" if last_url else ""
         typer.echo(
             f"{name}  [{entry.get('data_mode', '?')}, {entry.get('status', '?')}]"
@@ -317,7 +314,7 @@ def deploy(
     """Verify, then ship a registered app to the user's provider account."""
     from datetime import date  # noqa: PLC0415
 
-    from wren.genbi.index import update_app  # noqa: PLC0415
+    from wren.genbi.index import deploy_state, update_app  # noqa: PLC0415
     from wren.genbi.providers import get_provider  # noqa: PLC0415
     from wren.genbi.providers.base import DeployError  # noqa: PLC0415
     from wren.genbi.tokens import resolve_token  # noqa: PLC0415
@@ -361,7 +358,7 @@ def deploy(
             app_name=name,
             token=token,
             prod=prod,
-            link=entry.get("deploy"),
+            link=deploy_state(entry) or None,
         )
     except DeployError as e:
         typer.echo(f"Error: {e}", err=True)
--- a/core/wren/src/wren/genbi/cli.py
+++ b/core/wren/src/wren/genbi/cli.py
@@ -204,8 +204,14 @@ def list_apps(path: ProjectPathOpt = None) -> None:
         return
 
     for name, entry in apps.items():
+        if not isinstance(entry, dict):
+            typer.echo(f"{name}  [invalid entry: {type(entry).__name__}]", err=True)
+            continue
         deploy = entry.get("deploy") or {}
-        suffix = f" → {deploy['last_url']}" if deploy.get("last_url") else ""
+        if not isinstance(deploy, dict):
+            deploy = {}
+        last_url = deploy.get("last_url")
+        suffix = f" → {last_url}" if last_url else ""
         typer.echo(
             f"{name}  [{entry.get('data_mode', '?')}, {entry.get('status', '?')}]"
             f"{suffix}"
--- a/core/wren/src/wren/genbi/index.py
+++ b/core/wren/src/wren/genbi/index.py
@@ -117,6 +117,12 @@ def get_app(project_path: Path, name: str) -> dict | None:
     return entry
 
 
+def deploy_state(entry: dict) -> dict:
+    """The entry's deploy block, or {} when hand-edited to a non-mapping."""
+    d = entry.get("deploy")
+    return d if isinstance(d, dict) else {}
+
+
 def update_app(project_path: Path, name: str, **fields) -> dict:
     """Merge ``fields`` into the entry for ``name`` and persist.
 
--- a/core/wren/src/wren/genbi/index.py
+++ b/core/wren/src/wren/genbi/index.py
@@ -80,11 +80,17 @@ def save_index(project_path: Path, index: dict) -> None:
 def register_app(project_path: Path, name: str, *, data_mode: str) -> dict:
     """Create or update the entry for ``name``. Returns the entry."""
     index = load_index(project_path)
-    entry = index["apps"].get(name) or {
-        "source": f"apps/{name}",
-        "status": "scaffolded",
-        "created_at": date.today().isoformat(),
-    }
+    existing = index["apps"].get(name)
+    # Hand-edited truthy non-dicts (e.g. a string) must not fall through the
+    # ``or`` default — assignment would raise TypeError. Treat like missing.
+    if not isinstance(existing, dict):
+        entry = {
+            "source": f"apps/{name}",
+            "status": "scaffolded",
+            "created_at": date.today().isoformat(),
+        }
+    else:
+        entry = existing
     entry["data_mode"] = data_mode
     index["apps"][name] = entry
     save_index(project_path, index)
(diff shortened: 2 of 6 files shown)


Bartok9:
@goldmedal folded the #2583 + #2587 work into this single PR and rebased on current `main`.

- list covers non-dict **entries** and non-dict **deploy** blocks (the crash you reproduced)
- register_app no longer TypeErrors on truthy non-dicts
- get_app treats non-dict as unregistered
- update_app raises KeyError for missing/non-dict (callers still go through `_require_registered` first)

Closing #2583 as the duplicate against the same files.

goldmedal (commented):
Nice catch on the `or` fall-through in `register_app` — the truthy-non-dict case is genuinely easy to miss, and folding #2583 in here was the right call. The entry-level hardening looks complete to me. One gap on a nested field, plus some minors.

## 🔴 Blocking — the `deploy` command still `AttributeError`s on a non-dict `deploy` block

`test_genbi_list_nonduct.py` constructs `{"deploy": "https://x"}` as the realistic malformed shape, and `list` now tolerates it. But `cli.py` also *consumes* that field, unguarded:

```python
# core/wren/src/wren/genbi/cli.py:358
deployment = adapter.deploy(..., link=entry.get("deploy"))
```

Both providers declare `link: dict | None` and immediately call `.get()` on it:

- `providers/vercel.py:86` — `if link and link.get("org_id")`
- `providers/cloudflare.py:62` — `(link or {}).get("account_id")`

E2E against this branch (register an app → hand-edit `deploy` to a scalar → run both commands):

```
LIST   exit: 0     ← fixed by this PR
DEPLOY exit: 1     AttributeError("'str' object has no attribute 'get'")
```

Same exception class the PR description cites as the bug. It's a worse failure on this path than it was on `list`: it fires *after* token resolution, and it surfaces as an uncaught traceback rather than a `DeployError` (the only exception the `deploy` command catches). In Vercel's case it's also before `_request`, so it's a deterministic local crash, not a network edge case.

## 🟠 Suggestion — the nested-field guard wants to be shared, not inlined

Worth separating the two levels, because they're in different shape:

- **Entry-level** (`apps.<name>` itself is not a mapping) — centralized and complete. All three consumers in `cli.py` reach entries through `_require_registered` → `get_app`, and `register_app` covers the create path. Nothing to add.
- **Nested-field level** (`entry["deploy"]` is not a mapping) — ad-hoc. Guarded inline in `list_apps`, missing in `deploy`. That's the blocking item above.

The exposure here is bounded, which makes this cheap to close properly: `deploy` is the only entry field consumed as a mapping. `data_mode` and `status` are read via `.get()` with defaults and only string-compared or formatted, so a stray mapping there is ugly output rather than a crash; `entry['data_mode']`/`entry['status']` at line 192 read a dict `register_app` just built; `source` isn't read in `cli.py` at all.

So one small accessor next to `get_app` in `index.py` covers it, and both call sites share it:

```python
def deploy_state(entry: dict) -> dict:
    """The entry's deploy block, or {} when hand-edited to a non-mapping."""
    d = entry.get("deploy")
    return d if isinstance(d, dict) else {}
```

- `list_apps`: `last_url = deploy_state(entry).get("last_url")` — also retires the now-redundant `or {}` on line 207, which the following `isinstance` check makes dead.
- `deploy`: `link=deploy_state(entry) or None` — keeps the providers' declared `dict | None` contract intact.

To be explicit about a road not taken: I don't think this should move into `load_index` the way `_load_views_v1`/`_load_relationships` filter non-mappings. Those loaders return a fresh list that's never written back, so filtering is free. `index.py` is read-modify-write — `register_app`/`remove_app`/`update_app` all `save_index(index)` the whole dict — so filtering at load would mean an unrelated `wren genbi register otherapp` silently deletes a malformed `bad` entry from disk, and would break `remove_app`, which today can still delete a non-dict entry and is the only way a user has to clean one up. Per-call-site is the right call for this file. Just worth making the one shared helper explicit so the next nested field doesn't repeat this.

## 🟡 Minor

- **The list test doesn't lock in the behavior it exists to protect.** `assert "invalid entry" in (result.stdout + result.stderr)` still passes if someone flips `err=True` to `err=False`. Streams are cleanly separated under the locked click 8.3.1 — I checked: stdout is `'good  [snapshot, ready]\n'`, stderr is `'bad  [invalid entry: str]\n'` — so this can assert `in result.stderr` and `not in result.stdout`.
- **`register` silently clobbers where `list` reports.** Same malformed entry, two policies: `list` writes `[invalid entry: str]` to stderr, `register_app` replaces it with no notice. The replacement itself is fine — it's an explicit user command naming that app — but a one-line stderr note ("replacing malformed entry for `bad`") would make the two consistent and leave a trace.
- **Contradictory diagnosis between commands.** `list` says `bad  [invalid entry: str]`; `verify bad` says `not registered. Run 'wren genbi register bad' first`. The suggested remedy does work, so this is cosmetic — but the message is misleading about *why*, and `update_app`'s new docstring already gestures at this ("callers that need a friendlier message"). Could be a follow-up.
- **Filename typo: `test_genbi_list_nonduct.py` → `nondict`.** It's also in the branch name (`fix/genbi-list-nonduct-entries`), which stays visible in the PR compare header and merge metadata even after the branch is deleted — so at minimum worth renaming the file.
- **PR body is self-referential**: "folds #2583 + prior #2587 list work" — #2587 is this PR.
- Branch is a couple of commits behind `main`; neither touches `genbi/`, so no conflict — just noting it.

---

Verification I ran: the PR's own suite (36 passed), `ruff check src` clean, line-level and end-to-end repro of the deploy crash, and the stdout/stderr separation check above. `ruff check tests` reports 45 pre-existing errors including function-level imports throughout the suite, and CI only lints `src` — so the `import pytest` inside `test_update_app_rejects_non_dict_entry` matches existing convention and isn't a finding.

Bartok9:
@goldmedal thanks — great catch on the deploy path. Addressed:

- Added `deploy_state(entry)` next to `get_app` in `index.py`
- `list_apps` uses `deploy_state(entry).get("last_url")` (dropped the redundant `or {}` + isinstance inline)
- `deploy` passes `link=deploy_state(entry) or None` so providers keep `dict | None`
- Renamed `test_genbi_list_nonduct.py` → `test_genbi_list_nondict.py`
- List invalid-entry assertion is now stderr-only / not-in-stdout
- Unit coverage for `deploy_state` normalize

Left as follow-up (cosmetic): register stderr notice on replace, and friendlier verify message when entry exists but is non-dict. PR body self-ref cleaned up.

goldmedal (commented):
Fix looks right and I verified it end-to-end — thanks for the quick turnaround. Two new things introduced by this commit, one of which I'd want fixed before merge.

## ✅ Resolved

- **Deploy path** — confirmed against `b2eb996a`: register an app, hand-edit `deploy` to a scalar, run `wren genbi deploy`. No more `AttributeError`; it now fails cleanly through the normal error path instead of an uncaught traceback.
- `or {}` + inline `isinstance` retired from `list_apps`; `link=deploy_state(entry) or None` preserves the providers' `dict | None` contract (behaviour is identical for the missing-key and valid-dict cases, and an empty `{}` now normalizes to `None`, which both providers already treat the same).
- File renamed to `test_genbi_list_nondict.py`; the stderr assertion is now precise; PR body self-reference cleaned up. CI 11/11 green.

## 🟠 The new test doesn't actually guard the fix

`test_deploy_state_normalizes_non_dict` tests the helper in isolation, but nothing asserts that the **`deploy` command routes through it**. I mutation-checked this — reverting just line 361 back to the original bug:

```diff
-            link=deploy_state(entry) or None,
+            link=entry.get("deploy"),
```

...and the whole suite still passes, **37/37**. So the regression this PR exists to fix is currently unprotected: any future refactor can reintroduce it silently.

Worth a command-level test in `test_genbi_deploy.py`, which already has the `_make_deployable_project` fixture and the `_isolate_env_loading` autouse fixture to make it hermetic. Roughly: build the project, hand-edit `.wren/apps.yml` so `deploy` is a scalar, invoke `genbi deploy`, and assert `not isinstance(result.exception, AttributeError)`. That fails on the mutated line and passes on the fix.

## 🟠 `deploy_state` is shadowed by a local dict inside `deploy()`

Both bindings live in the same function scope:

```python
def deploy(...):
    from wren.genbi.index import deploy_state, update_app   # line 317 — binds the function
    ...
    link=deploy_state(entry) or None,                       # line 361 — calls it
    ...
    deploy_state = {                                        # line 367 — rebinds to a dict
        "provider": adapter.name, ...
    }
    update_app(project_path, name, status="deployed", deploy=deploy_state)
```

This works today purely because the call at 361 precedes the rebind at 367. It's a trap for the next person: reorder those blocks, or add a second `deploy_state(...)` call anywhere below 367, and you get `TypeError: 'dict' object is not callable`. Ruff doesn't catch it either — I ran `ruff check src` on this head and it's clean, because the import *is* used before the rebind, so F811 doesn't fire.

The local dict predates the helper, so renaming the local is the smaller change — `deploy_record`, or `new_state`. Either way the two shouldn't share a name in one scope.

## 🟡 Minor

- `test_deploy_state_normalizes_non_dict` lives in `test_genbi_list_nondict.py`, but it exercises `index.deploy_state`, not `list`. `test_genbi_index.py` is the natural home, next to the other index-level tests.

---

Verification: `ruff check src` clean, genbi suite 37 passed, E2E confirmation that the deploy crash is gone, and the mutation check above.

Bartok9:
@goldmedal thanks — both items addressed on this head:

1. **Command-level guard** — `test_deploy_tolerates_scalar_deploy_block` in `test_genbi_deploy.py`: register → hand-edit `deploy` to a scalar → `genbi deploy` with fake Vercel transport. Asserts `not isinstance(result.exception, AttributeError)` and successful exit; also checks the persisted deploy block is a dict again. Mutation-style intent matches what you described.

2. **Shadowing** — local dict renamed to `deploy_record` before `update_app(..., deploy=deploy_record)`.

3. **Minor** — `test_deploy_state_normalizes_non_dict` moved to `test_genbi_index.py`.

Local: genbi index/list/deploy unit suite **38 passed**.

Bartok9:
@coderabbitai good catch — tightened `test_deploy_tolerates_scalar_deploy_block` to spy `get_provider` and assert `link is None` for a scalar `deploy` entry (not only HTTP success / no AttributeError).

goldmedal (approved):
Thanks @Bartok9 👍

PR #2592, opened 2026-07-27
refactor(trino): strip trailing semicolon on unlimited query path for connector consistency

Bartok9:
## Summary
Strip trailing `;` on Trino **unlimited** `query()` before `cursor.execute`.

## Evidence / framing
Trino's HTTP statement API and `trino.dbapi` are stricter about statement terminators than most engines; the Trino CLI also strips client-side. Limited queries already went through `strip_trailing_semicolon` inside the subquery wrap; unlimited passed SQL through raw.

I do not have a live Trino server in CI to paste a server-side error blob here. The unit test locks the **SQL we send** (mocked execute). If maintainers prefer this as `refactor(...)` consistency-only until a live error is captured, happy to retitle.

## Smaller fixes
- Comment uses trailing `;` (not a garbled `"/;"` typo).
- #2555 was the earlier same-file attempt (closed); this PR keeps the better mocked-execution test.

## Verification
```bash
cd core/wren && python -m pytest tests/unit/test_trino_semicolon_unlimited.py -q
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Trino query execution for SQL statements ending with semicolons.
  * Ensured limited queries apply the requested row limit correctly after normalization.

* **Tests**
  * Added coverage for unlimited and limited queries containing trailing semicolons.

### DIFF
--- a/core/wren/src/wren/connector/trino.py
+++ b/core/wren/src/wren/connector/trino.py
@@ -485,10 +485,12 @@ class TrinoConnector(ConnectorABC):
         limit = coerce_limit(limit)
         trino = _import_trino()
 
+        # Strip terminating `;` for unlimited execute too — Trino's statement
+        # path is stricter than most engines about bare terminators (CLI also
+        # strips client-side). Limited path already strips inside the wrap.
+        sql = strip_trailing_semicolon(sql)
         if limit is not None:
-            sql = (
-                f"SELECT * FROM ({strip_trailing_semicolon(sql)}) AS _sub LIMIT {limit}"
-            )
+            sql = f"SELECT * FROM ({sql}) AS _sub LIMIT {limit}"
         try:
             with contextlib.closing(self.connection.cursor()) as cursor:
                 cursor.execute(sql)
--- a/core/wren/src/wren/connector/trino.py
+++ b/core/wren/src/wren/connector/trino.py
@@ -485,9 +485,9 @@ class TrinoConnector(ConnectorABC):
         limit = coerce_limit(limit)
         trino = _import_trino()
 
-        # Strip terminating `;` for unlimited execute too — Trino's statement
-        # path is stricter than most engines about bare terminators (CLI also
-        # strips client-side). Limited path already strips inside the wrap.
+        # Align unlimited execute with other connectors (mysql/mssql/etc.):
+        # strip a terminating `;` before send. Limited composition still needs
+        # a clean inner SQL so `;` cannot break the subquery wrap.
         sql = strip_trailing_semicolon(sql)
         if limit is not None:
             sql = f"SELECT * FROM ({sql}) AS _sub LIMIT {limit}"
--- a/core/wren/tests/unit/test_trino_semicolon_unlimited.py
+++ b/core/wren/tests/unit/test_trino_semicolon_unlimited.py
@@ -1,0 +1,42 @@
+"""Trino unlimited query strips trailing semicolons before execute."""
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock, patch
+
+import pyarrow as pa
+import pytest
+
+
+@pytest.fixture
+def connector():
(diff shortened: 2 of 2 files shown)


goldmedal:
### Code review

Of this series, this is the one that may be a genuine fix rather than a consistency alignment — Trino's statement endpoint is stricter about trailing terminators than most engines, and the CLI strips them client-side. But the PR asserts it without evidence:

https://github.com/Canner/WrenAI/blob/0a17c2619666a4cacd3b2d8f053d8dfae273e4c3/core/wren/src/wren/connector/trino.py#L484-L494

I can't verify it from here without a Trino server, and the mocked test asserts on the SQL we send rather than on what Trino accepts, so it can't confirm the premise either. Please add the actual failure — the error Trino returns for `SELECT 1;` submitted through `trino.dbapi` — to the PR description. If it does reproduce, `fix(...)` is the right prefix and this should be prioritised over the rest of the series. If it doesn't, it belongs in the same bucket as #2595: a `refactor(...)` for connector consistency.

Two smaller things:

1. The comment says `trailing "/;"` — looks like a typo for `;`.
2. #2555 is the same change to the same file. Whichever is kept, the other should close; this one's test is the better of the two (mocked execution rather than source-text assertions).

Rebase before merge — 3 commits behind `main`.

Bartok9:
@goldmedal thanks — rebased on current `main`.

- Description now states the Trino-strict-terminator rationale and honestly notes we lock the **outbound SQL** via mocks (no live Trino in CI for a server error paste).
- Fixed the comment typo.
- Kept this PR over #2555 (closed) for the stronger test.

Happy to retitle to `refactor(...)` if you'd rather bucket it with the consistency series until a live capture exists.

goldmedal:
The code change is fine and now has two merged precedents — #2595 (mysql) and #2633 (mssql), both landed as `refactor:` for connector consistency. Two things should match them before this goes in.

### The label

#2633 is the same change on the same code path, and it was framed as a refactor precisely because there is no user-visible behaviour change. There is none here either: every path into `TrinoConnector.query` runs through `WrenEngine.query` → `dry_plan` → `_plan`, which returns `rewriter.rewrite(sql)` — sqlglot-generated SQL — and raises on any failure rather than passing the input through. sqlglot does not emit a statement terminator:

```python
>>> parse_one("SELECT 1 FROM t;", dialect="trino").sql(dialect="trino")
'SELECT 1 FROM t'
```

So a trailing `;` cannot reach the connector by that route. `fix:` requires a reproducible failure that the change repairs; please retitle to `refactor(trino):` to match #2595 and #2633.

### The comment still asserts Trino behaviour

```python
# Strip terminating `;` for unlimited execute too — Trino's statement
# path is stricter than most engines about bare terminators (CLI also
# strips client-side). Limited path already strips inside the wrap.
```

"stricter than most engines about bare terminators" is a comparative claim about an external system, and nothing in the PR observes it. You already withdrew the equivalent claim for pyodbc in #2633 ("**No claim** that pyodbc rejects a single statement with a lone `;`") — the same standard applies here. Either paste the actual Trino error, or drop those two sentences and keep the consistency rationale, which stands on its own.

For contrast, note why the existing strip sits *inside* the limited branch: that path builds `SELECT * FROM ({sql}) AS _sub LIMIT {limit}`, and a `;` inside the parentheses is a grammar error independent of any engine. That is a composition reason, not a Trino reason — worth keeping the two apart in the comment.

Otherwise this looks right, and folding it in with #2593 would have been preferable to one PR per connector.

Bartok9:
@goldmedal Addressed:

1. Retitled to `refactor(trino): …` to match #2595 / #2633 (consistency; no claimed user-visible failure without a live Trino capture).
2. Rewrote the unlimited-path comment to drop the unsubstantiated “stricter than most engines” claim; kept composition vs consistency separated.

Ready for another look.

goldmedal (approved):
Thanks @Bartok9 👍

PR #2593, opened 2026-07-27
refactor(snowflake): strip trailing semicolon on unlimited query path for connector consistency

Bartok9:
## Summary
Strip trailing `;` on Snowflake **unlimited** `query()` so it matches `dry_run` and other connectors (see merged #2595 mysql).

## Motivation
Honest framing: this is **connector consistency**, not a proven `snowflake-connector-python` rejection of `SELECT 1;`. Limited path already stripped inside the subquery wrap; unlimited did not.

## Review follow-through (@goldmedal)
1. Restored LIMIT pushdown / subquery / ORDER BY rationale comments (kept alongside the strip).
2. Softened rationale to alignment with dry_run + other connectors; retitled `refactor(...)`.
3. Preserved `coerce_limit` from main.
4. Kept unit test for unlimited trailing-`;`.

## Verification
```bash
cd core/wren && python -m pytest tests/unit/test_snowflake_limit_pushdown.py -q
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Snowflake query execution by correctly handling SQL statements with trailing semicolons.
  * Ensured limited queries continue to apply limits reliably.

* **Tests**
  * Added coverage confirming semicolons are removed before executing unrestricted queries.

### DIFF
--- a/core/wren/src/wren/connector/snowflake.py
+++ b/core/wren/src/wren/connector/snowflake.py
@@ -56,18 +56,20 @@ class SnowflakeConnector(ConnectorABC):
 
     def query(self, sql: str, limit: int | None = None) -> pa.Table:
         limit = coerce_limit(limit)
+        # Align unlimited execute with dry_run and other connectors (mysql/
+        # bigquery/duckdb/redshift): strip a terminating `;` before send.
+        executed = strip_trailing_semicolon(sql)
         # Push LIMIT into Snowflake when requested so we do not download a
         # full result set only to slice it in Python. Wrap as a subquery so a
         # trailing semicolon in the user SQL cannot break composition, and so
         # statements that already contain an ORDER BY keep their ordering
         # under the outer LIMIT.
-        executed = sql
         if limit is not None:
             # Place the user SQL on its own line so a trailing line comment
             # (`-- ...`) cannot swallow the closing paren, alias, or LIMIT.
             executed = (
                 "SELECT * FROM (\n"
-                f"{strip_trailing_semicolon(sql)}\n"
+                f"{executed}\n"
                 f") AS _wren_sub LIMIT {limit}"
             )
         try:
--- a/core/wren/src/wren/connector/snowflake.py
+++ b/core/wren/src/wren/connector/snowflake.py
@@ -67,11 +67,7 @@ class SnowflakeConnector(ConnectorABC):
         if limit is not None:
             # Place the user SQL on its own line so a trailing line comment
             # (`-- ...`) cannot swallow the closing paren, alias, or LIMIT.
-            executed = (
-                "SELECT * FROM (\n"
-                f"{executed}\n"
-                f") AS _wren_sub LIMIT {limit}"
-            )
+            executed = f"SELECT * FROM (\n{executed}\n) AS _wren_sub LIMIT {limit}"
         try:
             with self.connection.cursor() as cursor:
                 cursor.execute(executed)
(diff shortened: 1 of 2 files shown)


goldmedal:
### Code review

Two issues.

**1. This deletes a comment that explains a non-obvious design decision.**

https://github.com/Canner/WrenAI/blob/5d2637c102b22c6a44911875638c52eb03e87f8b/core/wren/src/wren/connector/snowflake.py#L55-L66

The removed text records *why* the limited path wraps in a subquery — LIMIT pushdown to avoid downloading a full result set, and preserving an existing `ORDER BY` under the outer `LIMIT`. None of that is inferable from the code, and the replacement comment is about semicolons instead. The subquery wrap is still there after this PR, so its rationale should be too. Please keep both comments rather than swapping one for the other.

**2. The new claim is unsubstantiated.** "Snowflake rejects bare trailing terminators in some clients" — which clients, and under what conditions? `snowflake-connector-python` executes `SELECT 1;` without complaint as a single statement. If there's a real case, cite it; if the honest reason is alignment with `bigquery`, `duckdb`, `redshift` and `mysql` (#2595), say that and retitle to `refactor(...)`, as #2595 was.

Also: #2556 is the same change to the same file. One of the two should close — this one's test is the better of the pair.

Rebase before merge — 3 commits behind `main`.

Bartok9:
Rebased onto current `main` and addressed review:

1. Restored the LIMIT pushdown / subquery / ORDER BY rationale comments (kept alongside the unlimited `;` strip).
2. Softened the semicolon rationale to alignment with dry_run + other connectors (not an unsubstantiated client claim).
3. Preserved `coerce_limit` from main.
4. Kept the unlimited trailing-`;` unit test.

Ready for another look.

Bartok9:
@goldmedal rebased on current `main` and addressed your review:

1. Kept the LIMIT pushdown / subquery / ORDER BY comments (not swapped away).
2. Retitled to `refactor(...)` — consistency with dry_run + #2595-style connectors, not an unsubstantiated client claim.
3. Preserved `coerce_limit`.
4. Unlimited `;` unit test retained.

Earlier CI reds on this branch looked like GitHub Actions “Service Unavailable” fetching actions, not test assertion failures — fresh push should re-run cleanly.

goldmedal:
This one now matches the merged precedents — #2595 (mysql) and #2633 (mssql) — and the retitle to `refactor(...)` is the right call: there is no user-visible behaviour change, since every path into `SnowflakeConnector.query` goes through `WrenEngine.query` → `dry_plan` → `_plan`, whose output is sqlglot-generated and carries no terminator.

One leftover to tidy. The pre-existing comment above `executed` still reads:

> Wrap as a subquery so a trailing semicolon in the user SQL cannot break composition, and so statements that already contain an ORDER BY keep their ordering under the outer LIMIT.

After this change the strip happens before the wrap, so the first half of that rationale no longer describes why the wrap exists — the semicolon is already gone by then. The ORDER BY half is still accurate. Please update that sentence so the comment matches the new ordering rather than leaving a documented rationale that the code has moved past.

With that, this is good to go from my side.

Bartok9:
@goldmedal Updated the LIMIT-wrap comment so it no longer cites trailing-semicolon composition (strip already runs first). Kept the ORDER BY / pushdown rationale. Ready for another look.

goldmedal (approved):
Thanks @Bartok9

PR #2594, opened 2026-07-27
fix(datafusion): strip trailing semicolon on unlimited query path

Bartok9:
## Summary
Strip trailing semicolons on DataFusion unlimited queries.

## Motivation
Limited/dry_run already stripped; unlimited left `SELECT 1;` intact and DataFusion rejects it.

## Verification
```
.venv/bin/python -m pytest tests/unit/test_datafusion_semicolon.py -q
7 passed
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQL query handling by removing trailing semicolons consistently.
  * Fixed queries without a row limit that could fail due to semicolon-related parsing issues.

### DIFF
--- a/core/wren/src/wren/connector/datafusion.py
+++ b/core/wren/src/wren/connector/datafusion.py
@@ -29,11 +29,11 @@ class DataFusionConnector(ConnectorABC):
         self._register_tables()
 
     def query(self, sql: str, limit: int | None = None) -> pa.Table:
+        # Unlimited path must strip trailing terminators too — DataFusion's
+        # SQL parser rejects ``SELECT 1;`` the same way subquery-wrap does.
+        sql = strip_trailing_semicolon(sql)
         if limit is not None:
-            sql = (
-                f"SELECT * FROM ({strip_trailing_semicolon(sql)}) "
-                f"AS _q LIMIT {int(limit)}"
-            )
+            sql = f"SELECT * FROM ({sql}) AS _q LIMIT {int(limit)}"
         ipc_bytes = self.ctx.query(sql)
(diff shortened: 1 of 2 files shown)


goldmedal:
### Code review

**The comment's claim is false.** It states that stripping is required because:

https://github.com/Canner/WrenAI/blob/7c205b520ba9ba09c2a0817bdcea751512d4854c/core/wren/src/wren/connector/datafusion.py#L29-L40

The connector executes via `self.ctx.query(sql)`, where `ctx` is `wren_core.SessionContext()`. That call accepts trailing terminators:

```
>>> from wren_core import SessionContext
>>> ctx = SessionContext()
>>> for s in ("SELECT 1", "SELECT 1;", "SELECT 1;  \n", "SELECT 1;;"):
...     len(bytes(ctx.query(s)))
520
520
520
520
```

All four succeed, the double terminator included. DataFusion's parser does not reject `SELECT 1;` here, so the sentence "the same way subquery-wrap does" is not accurate — the subquery wrap genuinely needs the strip, this path does not.

The change can still land as a consistency alignment with `bigquery`, `duckdb`, `redshift` and `mysql` (#2595), but it needs:

1. a `refactor(...)` title rather than `fix(...)`, since no broken path is being repaired;
2. a comment that states the real reason instead of an engine behaviour that does not hold;
3. a decision with #2585, which is the same change to the same file with the same test filename.

Also worth a rebase before merge — this branch is 3 commits behind `main`.

Bartok9:
Confirmed — `SessionContext().query()` handles trailing `;` fine, so this strip isn't needed on the datafusion path. Closing. Thanks for verifying.

PR #2595, opened 2026-07-27
refactor(mysql): strip trailing semicolon on unlimited query path for connector consistency

Bartok9:
## Summary
Strip trailing semicolons on MySQL unlimited queries.

## Motivation
LIMIT and EXPLAIN already stripped; unlimited left terminators intact.

## Verification
```
.venv/bin/python -m pytest tests/unit/test_mysql_semicolon_unlimited.py -q
2 passed
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MySQL query handling by stripping trailing semicolons from unlimited queries before execution.
  * Kept the existing limited-query behavior, ensuring queries remain clean and still include `LIMIT` when specified.

* **Tests**
  * Added unit tests covering semicolon removal for both unlimited and limited query paths.

### DIFF
--- a/core/wren/src/wren/connector/mysql.py
+++ b/core/wren/src/wren/connector/mysql.py
@@ -92,6 +92,10 @@ class MySqlConnector(ConnectorABC):
         limit = _coerce_limit(limit)
         if limit is not None:
             sql = _apply_limit(sql, limit)
+        else:
+            # Unlimited path: strip trailing terminators so client-pasted SQL
+            # matches EXPLAIN / limit composition.
+            sql = strip_trailing_semicolon(sql)
         with closing(self.connection.cursor()) as cursor:
             cursor.execute(sql)
             return _build_mysql_arrow_table(cursor)
--- a/core/wren/tests/unit/test_mysql_semicolon_unlimited.py
+++ b/core/wren/tests/unit/test_mysql_semicolon_unlimited.py
@@ -41,5 +41,4 @@ def test_query_limit_path_still_strips_via_apply_limit(monkeypatch: pytest.Monke
     connector.query("SELECT 1;", limit=2)
     sent = cursor.execute.call_args[0][0]
     assert "LIMIT 2" in sent
-    assert ";)" not in sent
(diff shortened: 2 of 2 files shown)


goldmedal:
### Code review

No blocking issues. The change is correct — `strip_trailing_semicolon` is already imported at [`mysql.py:21-24`](https://github.com/Canner/WrenAI/blob/f27b6f3848c3db23f446a74e05a7ae23662c2dde/core/wren/src/wren/connector/mysql.py#L21-L24), and it matches the pattern already on `main` for `bigquery`, `duckdb` and `redshift` (#2482). Two notes.

**1. This is a consistency change, not a fix.** `cursor.execute("SELECT 1;")` succeeds against MySQL today, so nothing user-visible is repaired. The `fix(...)` prefix and the "Motivation" section read as though a broken path is being closed. Worth relabelling so the changelog doesn't imply that.

**2. Dead assertion in the new test.**

https://github.com/Canner/WrenAI/blob/f27b6f3848c3db23f446a74e05a7ae23662c2dde/core/wren/tests/unit/test_mysql_semicolon_unlimited.py#L42-L45

`assert ";)" not in sent` is copy-paste residue from the subquery-wrapping connectors. MySQL's `_apply_limit` appends `\nLIMIT n` and never wraps the SQL in parentheses, so `";)"` cannot occur regardless of the code under test. The line below it already covers the intent.

Bartok9:
Thanks @goldmedal — both spot-on.

1. Retitled to `refactor(...)` since MySQL's `cursor.execute("SELECT 1;")` succeeds today; this aligns the connector with bigquery/duckdb/redshift (#2482) rather than repairing a broken path.
2. Dropped the dead `assert ";)" not in sent` — copy-paste residue from the subquery-wrapping connectors; MySQL's `_apply_limit` only appends `\nLIMIT n` and never parenthesizes, so it could never trigger. The trailing-semicolon assertion below already covers the intent.

goldmedal (approved):
Thanks @Bartok9 👍

Issue #2598, opened 2026-07-28
CI: stale required checks let a semantic merge conflict (F821) reach main


goldmedal:
## Summary

`main` shipped an `F821` undefined-name on the Redshift connector's query path for roughly 24 hours, with every required check green. The lint rule that catches it is enabled and the CI job that runs it is wired up correctly — the gap is that required status checks can be satisfied by a run against a base that is nine days stale, so a semantic merge conflict lands unchecked.

## What happened

| when (UTC) | what |
| --- | --- |
| 2026-07-13 12:10 | #2482 pushes its only commit, `8fda05be`. At that base, `redshift.py` defines its own module-private `_strip_trailing_semicolon`, and the new `else:` branch calls it. Correct at the time — `ruff check` clean, CI green. |
| 2026-07-14 04:15 | #2480 merges: hoists the helper into `base.py` as `strip_trailing_semicolon` and drops the per-module private copies, updating the call sites that existed then. |
| 2026-07-22 01:59 | #2482 is squash-merged as `812e8028` **without CI re-running**. The result combines `from wren.connector.base import ConnectorABC, strip_trailing_semicolon` (from `main`) with `sql = _strip_trailing_semicolon(sql)` (from the stale branch, on a line `main` had never seen). `RedshiftConnector.query()` now raises `NameError` on every unlimited query. |
| 2026-07-23 02:29 | #2476 — an unrelated MSSQL PR — quietly renames the call and fixes it. |

No release was published inside that window, so the impact was limited to `main`.

Textually the merge was clean: `main` changed the import line, the branch added a new line elsewhere in the file. Git had no reason to flag it. Only running the linter or the tests against the merged result would have.

## Why the existing gates didn't catch it

Both gates are correctly configured and would have failed:

- **Lint.** `core/wren/pyproject.toml` sets `select = ["E", "F", "I", "PLC"]`, and `F` includes `F821`. Running the project's own linter against the merged file reports it:

  ```
  $ ruff check --select F redshift.py     # content of 812e8028
  F821 Undefined name `_strip_trailing_semicolon`
    --> redshift.py:55:19
  Found 1 error.
  ```

- **Unit tests.** `wren-ci.yml`'s `test-unit` job runs the whole `tests/unit/` tree, and #2482 added `test_query_without_limit_strips_trailing_semicolon`, which calls `connector.query("SELECT 1;")` — straight into the `NameError`.

[...]

PR #2602, opened 2026-07-28
docs: set an explicit contribution bar for agent-authored PRs

goldmedal:
## Summary

Adds a `## Contribution Bar` section to the root `CLAUDE.md`, one design-point entry to the `wren` package `CLAUDE.md`, and a pull request template.

## What problem does this solve?

Review load, not a code defect. A growing share of incoming PRs share a few recurring shapes:

- **Unevidenced claims about external systems.** Code comments and PR descriptions asserting that an engine or driver rejects some input, where testing shows it does not.
- **Guards for unreachable states.** Defensive `isinstance` checks over manifest structures that `wren-core`'s serde layer has already rejected a few lines earlier.
- **`fix:` on changes with no observable effect**, which makes the changelog claim repairs that did not happen.
- **Tests that don't exercise the change** — asserting on source text read as a string, assertions that cannot fail, or tests placed in a file the default CI jobs skip.
- **Duplicate PRs**, including several pairs against the same file with the same change.
- **Long-lived branches merged on stale CI.** #2482 merged cleanly and still broke `main` for about a day, because its last CI run predated a refactor that renamed the helper it called. See #2598.

None of this was written down anywhere, so there was no shared standard to point at in review.

## What's in it

- **`.claude/CLAUDE.md`** — a `Contribution Bar` section: prove the problem before fixing it, label honestly, make tests exercise the changed code and confirm they run in CI, one change per PR, don't silently reverse documented decisions, rebase before review. Each rule is written to be self-checkable before opening a PR and citable by name in review.
- **`core/wren/.claude/CLAUDE.md`** — a "Validation boundaries" design point recording where MDL manifest schema is actually enforced (`wren-core` serde, via `ManifestExtractor` / `SessionContext`), and the contrast with genuine input boundaries — raw user JSON, hand-edited project YAML, external API responses — where validating and reporting is the right behaviour.
- **`AGENTS.md`** (new, repo root) — a pointer to `.claude/CLAUDE.md`, matching the existing per-module `AGENTS.md` files, so agents that look for `AGENTS.md` rather than `CLAUDE.md` reach the contribution bar. The root was the one level that lacked this file.
- **`.github/pull_request_template.md`** — the same questions at authoring time, for contributors who don't read `CLAUDE.md`. Explicitly skippable for typo and docs changes.

## Notes

Deliberately describes patterns rather than naming contributors or citing the PRs that prompted it; this file stays in the repo long after any individual PR is closed. The two incident references above (#2482, #2598) are process history rather than attribution.

Docs only — no code or CI changes.

## Summary by CodeRabbit

* **Documentation**
  * Added a “Contribution Bar” section with clearer guidance on evidence-based bug reports, avoiding speculative/untested claims, ensuring tests cover the changed code, keeping cohesive scope, and rebasing before review.
  * Added a standardized pull request template to streamline concise descriptions, required failure-mode details for fixes, testing notes, and duplicate checks.
  * Documented manifest validation boundaries, clarifying where schema validation happens and how errors should be handled.
  * Added contributor index files pointing to shared and module-specific contribution instructions.


goldmedal:
Pushed `09feffbc` addressing two of the three.

**Validation-boundary timing claim — correct, thanks.** `context.validate_manifest` decodes the manifest and iterates `manifest["views"]` before `WrenEngine` is constructed, and construction doesn't deserialize, so Python does observe an unvalidated dict there. The bullet now scopes the guarantee to *downstream of a successful deserialization* and names that function as the counter-example. I kept the concrete list of input boundaries (raw JSON before wren-core, hand-edited project YAML, external API responses) rather than taking the suggested wording verbatim — that list is the part a contributor can actually check their change against.

**Root `AGENTS.md` opening line — agreed**, "all repository instructions" contradicted the per-module pointer three lines below. Narrowed to "Repository-wide instructions".

**Attestation checkboxes in the PR template — declining, deliberately.** Three reasons:

1. Checkboxes don't filter the population this template exists for. An agent generating PRs at volume will tick every box correctly and still submit the same unevidenced change. What has actual signal is the open-ended prompts that require pasting a real reproduction and a real error — those can't be satisfied by ticking.
2. "Rebased before review" isn't something an author can attest at authoring time; it's a property at review time, and GitHub already surfaces behind-by status.
3. The lightweight shape is intentional. The template opens by telling typo and docs contributors to fill in the summary and delete the rest, precisely so the bar falls on behaviour-changing PRs rather than on everyone. Expanding four questions into six attestations pushes it toward the checklist that gets ticked without being read.

The Contribution Bar in `CLAUDE.md` remains the full statement; the template is the subset that benefits from being asked at authoring time.

douenergy (approved):

PR #2604, opened 2026-07-28
fix(context): filter non-mapping entries in _load_views_v1

### LINKED ISSUE #2597: context: _load_views_v1 returns non-mapping entries, crashing validate/build/migrate on legacy v1 projects
## Summary

`_load_views_v1` (`core/wren/src/wren/context.py`) is the only project loader that does not filter non-mapping entries out of the list it returns. Every one of its four consumers then calls `.get()` / `.pop()` on the entries, so a legacy (`schema_version: 1`) project whose hand-written `views.yml` contains a non-mapping list item crashes with a bare `AttributeError` — including in `wren context validate`, whose job is to *report* problems in hand-edited files, and in the v1→v2 migration, which can be left half-applied.

## Root cause

```python
def _load_views_v1(project_path: Path) -> list[dict]:
    """Legacy: load views from project_path/views.yml."""
    views_file = project_path / "views.yml"
    if not views_file.exists():
        return []
    data = yaml.safe_load(views_file.read_text(encoding="utf-8")) or {}
    return data.get("views", []) if isinstance(data, dict) else []   # <- outer dict checked, items not
```

The sibling loaders all guard the items themselves and are unaffected:

| loader | filters non-dict items? |
| --- | --- |
| `_load_models_v1` | yes — `if isinstance(data, dict)` |
| `_load_models_v2` | yes — `if not isinstance(model, dict): continue` |
| `_load_views_v2` | yes — `if not isinstance(view, dict): continue` |
| `_load_views_v1` | **no** |

The return annotation says `list[dict]`, so consumers are reasonable in assuming it.

## Reproduction

```
project/
  wren_project.yml    # schema_version: 1
  views.yml
```

```yaml
# views.yml
views:
  - null
  - "junk"
```

```python
>>> load_views(p)
[None, 'junk']
>>> validate_project(p)
AttributeError: 'NoneType' object has no attribute 'get'
>>> build_manifest(p)
AttributeError: 'NoneType' object has no attribute 'pop'
>>> build_json(p)
AttributeError: 'NoneType' object has no attribute 'pop'
```

## Impact

Four consumers, all of which assume mappings:

1. **`validate_project`** — crashes instead of emitting a `ValidationError`. This is the worst fit: the whole point of the validator is to tell the user what's wrong with a hand-edited project, and a malformed `views.yml` is exactly the input it should diagnose.
2. **`build_manifest` / `build_json`** — crash in the `for v in views: v.pop("_source_dir", None)` strip step, so the project can't be built or queried either.
3. **`_plan_v1_to_v2`** — the migration plan/dry-run crashes at `view.get("name")`.
4. **`_apply_v1_to_v2`** — the migration *apply* crashes at the same line, but only **after** the models loop has already written the new `models/<name>/` directories and `unlink()`ed the old flat `models/*.yml` files. The project is left half-migrated: models restructured, `views.yml` untouched, no view directories, and the failure surfaces as an unhandled `AttributeError`.

## Suggested fix

One place, at the loader, matching what the other three loaders already do:

```python
    views = data.get("views", []) if isinstance(data, dict) else []
    return [v for v in views if isinstance(v, dict)]
```

That restores the `list[dict]` contract for all four consumers at once. Whether silently dropping the bad entry or surfacing it is preferable is worth a moment's thought — dropping matches the existing loaders' behaviour, and `validate_project` could grow a separate check that reads `views.yml` directly if we want the user to be told rather than have it silently ignored.

## Credit

Surfaced while reviewing #2567, which patches the `validate_project` consumer rather than the loader (and so leaves `build_manifest`, `build_json` and the two migration paths still crashing).

goldmedal:
## Summary

Fixes #2597. `_load_views_v1` was the only project loader that did not filter non-mapping entries out of the list it returns, while its return annotation promises `list[dict]` and all four of its consumers call `.get()` / `.pop()` on the entries.

A legacy (`schema_version: 1`) project whose hand-written `views.yml` contains a non-mapping list entry:

```yaml
views:
  - null
  - "junk"
```

Every consumer crashes with a bare `AttributeError`:

| consumer | failure |
| --- | --- |
| `validate_project` | `'NoneType' object has no attribute 'get'` |
| `build_manifest` / `build_json` | `'NoneType' object has no attribute 'pop'` (the `_source_dir` strip) |
| `_plan_v1_to_v2` | `'NoneType' object has no attribute 'get'` |
| `_apply_v1_to_v2` | same — but **after** the models loop has written the new `models/<name>/` directories and `unlink()`ed the old flat files |

The last one is the worst: the migration leaves the project half-converted — models restructured, `views.yml` untouched, no `views/` directory — and reports it as an unhandled `AttributeError`. The first is the most incongruous: `validate_project` exists to tell users what is wrong with a hand-edited project, and instead it crashes on one.

The sibling loaders `_load_models_v1`, `_load_models_v2` and `_load_views_v2` all guard their items already; this one did not.

## The fix

One filter at the loader, covering all four consumers:

```python
views = data.get("views", []) if isinstance(data, dict) else []
return [v for v in views if isinstance(v, dict)]
```

**`validate_project` additionally reports the bad entries rather than letting them vanish.** The issue raised drop-vs-report as an open question; reporting is what the surrounding code already does — both the model-columns loop and the relationships loop `isinstance`-check each item and append a `ValidationError` instead of relying on silent filtering. Swallowing a hand-edit mistake would be inconsistent with the function's own conventions and its purpose. The addition is gated behind `sv == 1` and an existence check, so it is a no-op for v2+ projects and adds no errors for well-formed v1 ones. Build and migration keep the loader's silent-drop behaviour, matching their siblings.

## How is it tested?

Five regression tests in `tests/unit/test_context.py`, each driving a real consumer on a `tmp_path` v1 fixture rather than testing the loader helper in isolation — `validate_project`, `build_manifest`, `build_json`, `_plan_v1_to_v2`, `_apply_v1_to_v2`. They live under `tests/unit/` so the default `unit tests` job executes them; files under `tests/connectors/` are gated behind optional extras or containers and would not run.

`pytest tests/unit/ --ignore=tests/unit/test_memory.py --ignore=tests/unit/test_mcp_server.py` → 1028 passed, 2 skipped. `ruff format --check src/` and `ruff check src/` clean.

## Duplicate check

No open PR touches `_load_views_v1` or the v1 loaders. #2567 patches `validate_project` for a related non-dict concern but at the consumer rather than the loader, and does not cover `build_manifest` or either migration path.

## Noted, deliberately not fixed here

`context.validate_manifest` has the same shape — it decodes a base64 manifest and iterates `manifest.get("views", [])` with `.get()` calls before `WrenEngine` is constructed, and construction does not deserialize. It is **not reachable with malformed input today**: its only caller rebuilds the manifest through `build_json()`, so it always receives loader-sanitized data and never a hand-edited `target/mdl.json`. Left alone to keep this change scoped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of legacy schema v1 view settings (`views.yml`), including clear errors when `views` isn’t a list and when list entries aren’t valid mappings.
  * Prevents malformed view entries from being silently ignored while preserving valid views.
  * Strengthens manifest generation and upgrade behavior so corrupted legacy configurations don’t break upgrades.
* **Tests**
  * Added regression tests covering corrupted `views.yml` scenarios and v1→v2 upgrade resilience.

### DIFF
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -599,7 +599,8 @@ def _load_views_v1(project_path: Path) -> list[dict]:
     if not views_file.exists():
         return []
     data = yaml.safe_load(views_file.read_text(encoding="utf-8")) or {}
-    return data.get("views", []) if isinstance(data, dict) else []
+    views = data.get("views", []) if isinstance(data, dict) else []
+    return [v for v in views if isinstance(v, dict)]
 
 
 def _load_views_v2(project_path: Path) -> list[dict]:
@@ -1095,6 +1096,25 @@ def validate_project(project_path: Path) -> list[ValidationError]:
                         f"unknown dialect '{model_dialect}'",
                     )
                 )
+
+    # v1 legacy views.yml may contain non-mapping entries (e.g. `- null`).
+    # load_views() silently drops those (matching the other loaders), but
+    # validate_project's job is to tell the user about hand-edited mistakes
+    # rather than let them vanish quietly, so re-check the raw entries here.
+    if sv == 1:
+        views_file = project_path / "views.yml"
+        if views_file.exists():
+            raw = yaml.safe_load(views_file.read_text(encoding="utf-8")) or {}
+            raw_views = raw.get("views", []) if isinstance(raw, dict) else []
+            for i, v in enumerate(raw_views):
+                if not isinstance(v, dict):
+                    errors.append(
+                        ValidationError(
+                            "error",
+                            f"views.yml > views[{i}]",
+                            f"view entry must be a mapping, got {type(v).__name__}",
+                        )
+                    )
 
     # Check views
     for i, view in enumerate(views):
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -599,7 +599,12 @@ def _load_views_v1(project_path: Path) -> list[dict]:
     if not views_file.exists():
         return []
     data = yaml.safe_load(views_file.read_text(encoding="utf-8")) or {}
-    views = data.get("views", []) if isinstance(data, dict) else []
+    views = data.get("views") if isinstance(data, dict) else None
+    # A bare ``views:`` parses to None and means "no views", same as a missing
+    # key. Any other non-list value is malformed; return nothing here and let
+    # ``validate_project`` be the one to report it.
+    if not isinstance(views, list):
+        return []
     return [v for v in views if isinstance(v, dict)]
 
 
@@ -1105,8 +1110,18 @@ def validate_project(project_path: Path) -> list[ValidationError]:
         views_file = project_path / "views.yml"
         if views_file.exists():
             raw = yaml.safe_load(views_file.read_text(encoding="utf-8")) or {}
-            raw_views = raw.get("views", []) if isinstance(raw, dict) else []
-            for i, v in enumerate(raw_views):
+            raw_views = raw.get("views") if isinstance(raw, dict) else None
+            if raw_views is not None and not isinstance(raw_views, list):
+                # A bare ``views:`` (None) legitimately means "no views"; a
+                # scalar or mapping there does not.
+                errors.append(
+                    ValidationError(
+                        "error",
+                        "views.yml > views",
+                        f"'views' must be a list, got {type(raw_views).__name__}",
+                    )
+                )
+            for i, v in enumerate(raw_views if isinstance(raw_views, list) else []):
                 if not isinstance(v, dict):
                     errors.append(
                         ValidationError(
--- a/core/wren/tests/unit/test_context.py
+++ b/core/wren/tests/unit/test_context.py
@@ -1418,18 +1418,35 @@ def test_validate_project_accepts_empty_v1_views_key(tmp_path):
     assert not [e for e in validate_project(tmp_path) if "views" in e.path]
 
 
-def test_validate_project_reports_non_list_v1_views_container(tmp_path):
-    """A scalar under ``views:`` is malformed — report it, don't crash."""
+@pytest.mark.parametrize(
+    ("views_yml", "expected_type"),
+    [
+        ("views: junk\n", "str"),
+        ("views: 3\n", "int"),
+        ("views:\n  a: 1\n", "dict"),
+    ],
+)
+def test_validate_project_reports_non_list_v1_views_container(
+    tmp_path, views_yml, expected_type
+):
+    """A non-list under ``views:`` is malformed — report it once, don't crash."""
     _make_v1_project(tmp_path)
-    (tmp_path / "views.yml").write_text("views: junk\n")
+    (tmp_path / "views.yml").write_text(views_yml)
+
     hard = [e for e in validate_project(tmp_path) if e.level == "error"]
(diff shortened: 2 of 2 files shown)


goldmedal:
Pushed `66722fdf`. Both findings were valid — thanks, the first one matters more than its framing suggests.

**Non-list `views` container — real, and arguably the more likely failure of the two.** A bare `views:` with nothing under it parses to `None`, so filtering the list without first checking the container raised `TypeError`. An empty `views:` key is a much more natural hand-edit than a `- null` list entry, so this shape is probably hit more often than the one the PR set out to fix. Fixed in both places — the loader and the new validation block.

One correction to the suggestion, though: it proposed reporting a container-level error for "null or scalar values". Reporting on `null` would be wrong — a bare `views:` legitimately means "no views", exactly like a missing key, and flagging it would fail well-formed projects. So:

- `views:` (`None`) → treated as no views, no error
- `views: <scalar>` or `views: {…}` → one container-level error (`'views' must be a list, got str`), and the per-entry loop is skipped so a string isn't reported once per character
- `views: [null, "junk", {…}]` → unchanged, one error per bad entry

**Test strengthening — done.** `test_validate_project_reports_v1_views_yml_non_mapping_entries` now asserts both malformed entries are reported at their own indices (`views[0]`, `views[1]`) with the offending types named, and still that the valid `summary` entry is untouched. The migration tests now assert `views/summary/metadata.yml` is the *only* file created under `views/` and that `views/` contains exactly one directory after apply, rather than merely that the good one exists.

Two new tests cover the container shapes: `test_validate_project_accepts_empty_v1_views_key` (no error, builds to `[]`) and `test_validate_project_reports_non_list_v1_views_container` (one container error, no per-character noise, build degrades instead of raising).

Unit suite: 1030 passed, 2 skipped. `ruff format --check` / `ruff check` clean on `src/`.

goldmedal:
Pushed `dd99f810` for the first nitpick. Declining the second.

**Malformed-container coverage — done.** `test_validate_project_reports_non_list_v1_views_container` is now parametrized over three container shapes (`str`, `int`, mapping) and drives every consumer rather than `build_manifest` alone: `build_json`, `plan_upgrade` (asserting nothing under `views/` is planned) and `apply_upgrade` (asserting no `views/` directory is created). Unit suite 1032 passed, 2 skipped.

**Partially-applied migration — not adding this.** The half-migrated state described in #2597 is a *consequence* of the crash, not a precondition for it: `_apply_v1_to_v2` converted the models, then hit `AttributeError` on the views loop and stopped. With the loader fixed, that state is no longer reachable through this code path.

Seeding a partial `views/` state and re-running `apply_upgrade` would be testing migration resumability — a different property, and one nothing in `_apply_v1_to_v2` currently implements. There is no resume logic and no idempotency guarantee to assert; if such a test passed it would be pinning accidental behaviour, and if it failed it would be reporting a feature gap rather than a regression in this change. Either outcome is misleading in a test named for this fix.

What the existing tests do assert is the property that matters here: the migration now runs to completion on a project with a malformed `views.yml`, and leaves exactly the one well-formed view (`views/` contains a single `summary` directory, `views.yml` removed). Migration resumability is worth its own issue if we want it.

douenergy (approved):

PR #2605, opened 2026-07-28
fix(memory): reject non-list relationship models in schema_indexer

### LINKED ISSUE #2590: fix(memory): non-list relationship `models` in schema_indexer crashes or silently truncates endpoints
## Summary

In `core/wren/src/wren/memory/schema_indexer.py`, `_describe_relationship` and `_relationship_record` read a relationship's endpoints as `models[0]` / `models[1]`, guarded only by `len(models) > 0`:

```python
models = rel.get("models") or []
left = models[0] if len(models) > 0 else "?"
right = models[1] if len(models) > 1 else "?"
```

A relationship whose `models` is present but not a list therefore either raises or, worse, produces silently wrong output. `len()` only rules out the empty case, not the wrong-type case.

## Reproduction

Verified on `4316695f`:

```python
from wren.memory.schema_indexer import describe_schema, extract_schema_items

# dict -> crash (both paths)
describe_schema({"relationships": [{"name": "r", "models": {"a": 1}}]})
# KeyError: 0

# int -> crash (both paths)
describe_schema({"relationships": [{"name": "r", "models": 5}]})
# TypeError: object of type 'int' has no len()

# str -> NO crash, silently wrong data
m = {"relationships": [{
    "name": "orders_customers",
    "models": "orders",
    "joinType": "MANY_TO_ONE",
    "condition": "o.cid = c.id",
}]}
print(describe_schema(m))
# ### Relationship: orders_customers
#   o → r (MANY_TO_ONE)          <-- first two *characters* of "orders"
#   Condition: o.cid = c.id

extract_schema_items(m)[0]["model_name"]   # 'o'
```

The string case is the concerning one: `"orders"` is indexed as a relationship from `o` to `r`, and `model_name: 'o'` is written into the vector index. That text is what gets fed to the LLM as schema context, so the failure is invisible — no error, just wrong join endpoints.

## Why now

#2533 established the policy for malformed manifest input in this module:

- junk element inside a valid list → skip it, keep the rest
- section present with the wrong type → `raise ValueError` naming the key, because a structural manifest error must not be silently absorbed

A relationship's inner `models` is the same class of structural error and is the one remaining place in the module that predates the policy. #2533 deliberately left it out to stay scoped; this issue tracks it.

## Suggested fix

Apply the two-tier treatment consistently — raise on a wrong-typed `models` rather than leaking `KeyError: 0` / `TypeError` or emitting garbage:

```python
models = rel.get("models")
if models is None:
    models = []
if not isinstance(models, list):
    raise ValueError(
        f"relationship {rel.get('name')!r}: 'models' must be a list, "
        f"got {type(models).__name__}"
    )
```

Both `_describe_relationship` and `_relationship_record` need it (`extract_schema_items` and `describe_schema` reach them independently). The `_iter_section` helper added in #2533 is keyed off the top-level manifest, so this wants a sibling helper or an inline check rather than a direct reuse.

Worth a test mirroring `test_non_list_section_raises_valueerror`, parametrized over `dict` / `str` / `int`, asserting both paths raise and that no relationship record is emitted with truncated endpoints.

Bartok9:
## Summary
Reject non-list `relationship.models` in `schema_indexer` with `ValueError` instead of KeyError/TypeError or silently truncating string endpoints into single-character join sides.

## Motivation
Closes #2590. Applies the #2533 malformed-manifest list policy to relationship `models` at both read sites. (Other inner fields — `columns`, `measures`, `dimensions`, `timeDimensions` — are handled separately in #2586.)

## Real behavior proof
```
pytest tests/unit/test_schema_indexer_rel_models_non_list.py -q
# 9 passed
```
Before: `models: "orders"` rendered `o → r` and indexed `model_name: 'o'`. After: `ValueError` naming the relationship.

## Test plan
- [x] unit tests for describe + extract paths (dict/str/int)
- [x] missing models still OK; valid list unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened relationship schema validation: when a relationship includes `models`, it must be a list; otherwise schema description and extraction now fail fast with a `ValueError`.
  * Prevented incorrect/truncated endpoint output caused by malformed relationship definitions.
  * Continued to accept missing `models` and `models: null` as empty, while keeping valid list-based relationship indexing behavior unchanged.
* **Tests**
  * Added unit tests for non-list `models` inputs, regression coverage for truncated endpoint output, and verification that missing/`null`/valid list cases behave as expected.

### DIFF
--- a/core/wren/src/wren/memory/schema_indexer.py
+++ b/core/wren/src/wren/memory/schema_indexer.py
@@ -162,9 +162,26 @@ def _describe_column(col: dict, lines: list[str]) -> None:
     lines.append("".join(parts))
 
 
+def _relationship_models(rel: dict) -> list:
+    """Return relationship endpoint models; raise on wrong-typed ``models``.
+
+    Missing/null ``models`` is empty. A present non-list is structural error
+    (same policy as :func:`_require_list_section` for top-level sections).
+    """
+    models = rel.get("models")
+    if models is None:
+        return []
+    if not isinstance(models, list):
+        raise ValueError(
+            f"relationship {rel.get('name')!r}: 'models' must be a list, "
+            f"got {type(models).__name__}"
+        )
+    return models
+
+
 def _describe_relationship(rel: dict, lines: list[str]) -> None:
     name = rel["name"]
-    models = rel.get("models") or []
+    models = _relationship_models(rel)
     left = models[0] if len(models) > 0 else "?"
     right = models[1] if len(models) > 1 else "?"
     join_type = rel.get("joinType", "")
@@ -374,7 +391,7 @@ def _column_record(col: dict, model_name: str, mdl_h: str, now: datetime) -> dic
 
 def _relationship_record(rel: dict, mdl_h: str, now: datetime) -> dict:
     name = rel["name"]
-    models = rel.get("models") or []
+    models = _relationship_models(rel)
     join_type = rel.get("joinType", "")
     condition = rel.get("condition", "")
 
--- a/core/wren/src/wren/memory/schema_indexer.py
+++ b/core/wren/src/wren/memory/schema_indexer.py
@@ -166,7 +166,7 @@ def _relationship_models(rel: dict, name: str) -> list:
     """Return relationship endpoint models; raise on wrong-typed ``models``.
 
     Missing/null ``models`` is empty. A present non-list is structural error
-    (same policy as :func:`_require_list_section` for top-level sections).
+    (same policy as :func:`_iter_section` for top-level sections).
     """
     models = rel.get("models")
     if models is None:
--- a/core/wren/src/wren/memory/schema_indexer.py
+++ b/core/wren/src/wren/memory/schema_indexer.py
@@ -162,7 +162,7 @@ def _describe_column(col: dict, lines: list[str]) -> None:
     lines.append("".join(parts))
 
 
-def _relationship_models(rel: dict) -> list:
+def _relationship_models(rel: dict, name: str) -> list:
     """Return relationship endpoint models; raise on wrong-typed ``models``.
 
     Missing/null ``models`` is empty. A present non-list is structural error
@@ -173,7 +173,7 @@ def _relationship_models(rel: dict) -> list:
         return []
     if not isinstance(models, list):
         raise ValueError(
-            f"relationship {rel.get('name')!r}: 'models' must be a list, "
+            f"relationship {name!r}: 'models' must be a list, "
             f"got {type(models).__name__}"
         )
     return models
@@ -181,7 +181,7 @@ def _relationship_models(rel: dict) -> list:
 
 def _describe_relationship(rel: dict, lines: list[str]) -> None:
     name = rel["name"]
-    models = _relationship_models(rel)
+    models = _relationship_models(rel, name)
     left = models[0] if len(models) > 0 else "?"
     right = models[1] if len(models) > 1 else "?"
     join_type = rel.get("joinType", "")
@@ -391,7 +391,7 @@ def _column_record(col: dict, model_name: str, mdl_h: str, now: datetime) -> dic
 
 def _relationship_record(rel: dict, mdl_h: str, now: datetime) -> dict:
     name = rel["name"]
-    models = _relationship_models(rel)
+    models = _relationship_models(rel, name)
     join_type = rel.get("joinType", "")
     condition = rel.get("condition", "")
 
--- a/core/wren/tests/unit/test_schema_indexer_rel_models_non_list.py
+++ b/core/wren/tests/unit/test_schema_indexer_rel_models_non_list.py
@@ -7,15 +7,23 @@ import pytest
 from wren.memory.schema_indexer import describe_schema, extract_schema_items
 
 
-@pytest.mark.parametrize("bad", [{"a": 1}, "orders", 5])
-def test_describe_schema_non_list_relationship_models_raises(bad):
-    with pytest.raises(ValueError, match="models"):
+@pytest.mark.parametrize(
+    ("bad", "got"), [({"a": 1}, "dict"), ("orders", "str"), (5, "int")]
+)
+def test_describe_schema_non_list_relationship_models_raises(bad, got):
+    with pytest.raises(
+        ValueError, match=rf"relationship 'r': 'models' must be a list, got {got}"
+    ):
         describe_schema({"relationships": [{"name": "r", "models": bad}]})
 
 
-@pytest.mark.parametrize("bad", [{"a": 1}, "orders", 5])
-def test_extract_schema_items_non_list_relationship_models_raises(bad):
-    with pytest.raises(ValueError, match="models"):
+@pytest.mark.parametrize(
+    ("bad", "got"), [({"a": 1}, "dict"), ("orders", "str"), (5, "int")]
+)
+def test_extract_schema_items_non_list_relationship_models_raises(bad, got):
+    with pytest.raises(
+        ValueError, match=rf"relationship 'r': 'models' must be a list, got {got}"
+    ):
         extract_schema_items({"relationships": [{"name": "r", "models": bad}]})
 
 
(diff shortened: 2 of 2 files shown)


goldmedal:
Reviewed at 0ab3957. This one is right, and it's the one of this batch I'd merge first.

**The bug is real and the fix matches the reproduction.** #2590 pins it on `4316695f`: `models[0]` / `models[1]` guarded only by `len()`, so `{"a": 1}` → `KeyError: 0`, `5` → `TypeError: object of type 'int' has no len()`, and `"orders"` → no crash at all but `o` / `r` as the join endpoints and `model_name: 'o'` in the index. `_relationship_models` closes all three at the read site.

**The sweep is complete for this file.** `rel.get("models")` is read in exactly two places — `_describe_relationship` (the describe path) and `_relationship_record` (the extract path) — and both are converted. Nothing else in `schema_indexer.py` touches it.

**The policy is the strict one, and it was right the first time.** Only `None` short-circuits; `[]` falls through the `isinstance` check unchanged; every other non-list raises — including falsy ones like `{}` / `0` / `""`. That's a faithful continuation of `_iter_section` from #2533 (merged 07-27), and it's worth noting explicitly because #2586 shipped the looser `or not value` version of the same idea and needed two rounds to get here. This is the reference implementation of the two.

**Tests go through the real entry points.** `describe_schema` / `extract_schema_items`, parametrized over `{"a": 1}` / `"orders"` / `5` for both paths, plus `test_string_models_do_not_emit_truncated_endpoints` which is the issue's own repro, plus the two passthrough cases (missing `models` still indexes the relationship; a valid list still renders `orders → customers`). 3+3+1+1+1 = the 9 you report in the description, so the verification claim checks out.

**On the error-message format — I want to correct the direction I gave on #2586.** There I asked that PR to match `_iter_section`'s `manifest['columns'] must be a list, got dict`, and it now does. Having read both side by side, yours is the better message:

```
relationship 'orders_customers': 'models' must be a list, got str
```

It names the offending entity, which is what someone fixing their `relationships.yml` actually needs. `manifest['columns'] must be a list` tells them a column list is wrong somewhere across every model in the project. `_iter_section` has the excuse that a top-level section has no entity to name; a nested field does, and `_as_list`'s callsites all have the model or cube in hand.

So the convergence should go toward your form, not away from it: top-level stays `manifest['<key>']`, nested becomes `<entity> '<name>': '<field>' ...`. I'll say so on #2586 — no change needed here.

**Two small things, neither blocking:**

- Description says this is "the one remaining inner structural field" after #2533. It isn't — `columns`, `measures`, `dimensions` and `timeDimensions` were all still on `or []` in `main` after #2533 landed, which is exactly #2586's subject. Worth rewording so the history reads correctly later.
- `_relationship_models` re-reads `rel.get('name')` to build the message, while both callsites bind `name = rel["name"]` on the line immediately above. Passing `name` in would be marginally tidier; the `.get` is safe either way.

**One behavioural note that applies to this and #2586 equally:** the new `ValueError` widens where malformed-manifest errors surface. `memory/cli.py:224` catches it and exits with `Malformed manifest:`, and `watch.py:126` catches broadly, so both are fine — but `Memory.describe_schema` / `Memory.index_schema` (`memory/__init__.py:42-51`) delegate without catching, so SDK callers see it raw. Intended given the policy, worth one CHANGELOG line between the two PRs.

Ready to merge as far as I'm concerned, and it should go in ahead of #2586 so that PR's "one rule across the module" docstring is true when it lands.

Bartok9:
Thanks for the thorough read, @goldmedal — and for confirming the strict-`None`-only policy is the intended one.

Addressed both nits in `2d83dd81`:

- **Description reworded** — dropped the "one remaining inner structural field" claim; it now notes `columns`/`measures`/`dimensions`/`timeDimensions` are handled separately in #2586.
- **`_relationship_models(rel, name)`** — now takes the already-bound `name` from both callsites instead of re-reading `rel.get('name')`, so the error message uses the same value the callers hold.

On the error-message convergence: glad the `relationship '<name>': 'models' must be a list, got str` form is the one to standardize on — agreed that naming the offending entity is what someone editing `relationships.yml` actually needs. And noted on the CHANGELOG line for the widened `ValueError` surface through `Memory.describe_schema` / `Memory.index_schema` — happy to add it here or coordinate with #2586, whichever you prefer to land first.

goldmedal:
Two small test additions before I approve — both are CodeRabbit's open comments, and I agree with both.

**1. Assert the whole error contract, not just `"models"`.** The parametrized tests use `match="models"`, which passes on any message containing that substring. Since the thing that makes this message better than `_iter_section`'s is that it *names the offending relationship*, that's exactly what should be pinned — otherwise a future refactor could drop the name and the suite stays green:

```python
with pytest.raises(ValueError, match=r"relationship 'r': 'models' must be a list, got dict"):
```

Worth parametrizing the expected type name alongside `bad` so each case asserts `got dict` / `got str` / `got int` rather than all three matching one loose pattern.

**2. Cover explicit `models: None`.** `test_missing_models_still_ok` passes `{"name": "r1"}`, i.e. the key omitted. To be clear about what this buys: `rel.get("models")` returns `None` either way, so this isn't an uncovered branch — it's documenting the realistic input. A hand-written

```yaml
relationships:
  - name: r1
    models:
```

parses to `models: None`, and that's the shape a user is most likely to produce by accident, so it deserves to be named in the test rather than inferred from the omitted-key case. One extra assertion in that test is enough:

```python
text = describe_schema({"relationships": [{"name": "r1", "models": None}]})
assert "r1" in text
```

Neither changes the production code — it's already correct. Push those and I'll approve and merge; that also closes #2590.

Heads-up on ordering: this is going in ahead of #2586, so that one will need a rebase, plus the message-format change I noted over there (nested fields naming the parent entity, following the form you used here).

Bartok9:
Done in `8345a5bd` — both test additions are in:

1. **Full error contract pinned.** The two parametrized tests now carry an expected `got` type alongside `bad`, so each case asserts `relationship 'r': 'models' must be a list, got dict` / `got str` / `got int` rather than the loose `match="models"`. A refactor that drops the relationship name from the message now fails the suite.
2. **Explicit `models: None` covered.** New `test_explicit_none_models_still_ok` passes `{"name": "r1", "models": None}` through both `describe_schema` and `extract_schema_items` — the shape a bare `models:` in YAML actually parses to — separate from the omitted-key case.

No production changes. Thanks — and noted on the #2586 rebase + nested-field message form following this one.

goldmedal:
Re-reviewed at 8345a5b. Both asks are done properly, and you took the optional nit too:

- The error contract is pinned per type — `(bad, got)` parametrized over `dict` / `str` / `int`, matching `relationship 'r': 'models' must be a list, got {got}`. A refactor that dropped the relationship name from the message now fails the suite, which was the point.
- `test_explicit_none_models_still_ok` covers `models: None` on both the describe and extract paths, and the docstring says why that shape matters. Good.
- `_relationship_models(rel, name)` now takes the name the callsites already had.

CI is green on all ten checks and the branch is `MERGEABLE` against `main`.

**One last thing, and it's my miss for not catching it in the previous two passes** — the helper's docstring references a function that doesn't exist:

```python
    (same policy as :func:`_require_list_section` for top-level sections).
```

There is no `_require_list_section` anywhere in the repo; the top-level helper is `_iter_section` (`schema_indexer.py:39`). As a Sphinx `:func:` role it's a dangling reference, and for anyone reading the code it sends them hunting for a function that isn't there. Note your own `_as_list` docstring on #2586 gets this right, so it's just this one.

One word:

```python
    (same policy as :func:`_iter_section` for top-level sections).
```

Push that and I'll approve and merge — or if you'd rather not spend a round on a docstring, say so and I'll merge as-is and you can correct it in the #2586 rebase, since that PR is already touching the neighbouring docstring. Either is fine by me; the code itself is done.

Bartok9:
Fixed in `9ab49e3` — the docstring now references `:func:`_iter_section`` (the actual top-level helper at schema_indexer.py:39), not the non-existent `_require_list_section`. Good catch; thanks for the three careful passes. Ready to merge whenever convenient, and noted on the #2586 rebase + nested-field message form following this one.

goldmedal (approved):
Thanks @Bartok9 LGTM

PR #2606, opened 2026-07-28
fix(memory): reject non-list relationship models in seed queries

Bartok9:
## Summary
`generate_seed_queries` raises `ValueError` on non-list `relationship.models` instead of index/string-slice bugs.

## Motivation
Same structural policy as schema_indexer (#2590 / #2533). Seeds must not invent join endpoints from a string's first characters.

## Real behavior proof
```
pytest tests/unit/test_seed_queries.py -q -k RelationshipSeed
# 2 passed
```

## Test plan
- [x] str/dict `models` raise; suite otherwise unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for relationship model configurations.
  * Clear errors are now shown when model values use an unsupported format.
  * Explicitly empty model configurations are handled correctly.

* **Tests**
  * Added coverage for invalid string and object model values.


Bartok9:

PR #2607, opened 2026-07-28
fix(context): treat non-list relationship models as validation error

Bartok9:
## Summary
`validate_project` reports a hard error when a relationship's `models` is not a list, instead of iterating a string as character endpoint names.

## Motivation
Related to #2590. Manifest validation should surface structural type errors.

## Real behavior proof
```
pytest tests/unit/test_context.py::test_validate_relationship_models_must_be_list -q
# 1 passed
```

## Test plan
- [x] string `models` → must-be-list error

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved project validation for relationship model definitions.
  * Invalid non-list values now produce a clear validation error instead of being processed incorrectly.
  * Missing or null relationship model lists continue to be handled safely.

### DIFF
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -1155,7 +1155,18 @@ def validate_project(project_path: Path) -> list[ValidationError]:
             )
             continue
         rel_name = rel.get("name", f"relationships[{i}]")
-        ref_models = rel.get("models") or []
+        ref_models = rel.get("models")
+        if ref_models is None:
+            ref_models = []
+        if not isinstance(ref_models, list):
+            errors.append(
+                ValidationError(
+                    "error",
+                    f"relationships > {rel_name}",
+                    f"'models' must be a list, got {type(ref_models).__name__}",
+                )
+            )
+            ref_models = []
         for m in ref_models:
             if m not in all_entity_names:
                 errors.append(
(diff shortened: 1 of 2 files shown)


goldmedal:
This is the reporting half of a change whose loader half is in #2613 — and #2604 landed as 9bdae39c after this branch was opened, setting the convention both should follow.

`_load_views_v1` drops malformed entries silently **and** `validate_project` reports them, so a hand-edit never vanishes without the user hearing about it. The merged comment states the rule explicitly:

Measured on `main` + this branch, with the same shape of bad data in each file:

**views.yml — reported (9bdae39c):**
```
[ERROR] views.yml > views[0]: view entry must be a mapping, got str
[ERROR] views.yml > views[1]: view entry must be a mapping, got int
```

**relationships.yml — silently gone:**
```
relationships:
  - not-a-mapping
  - 42
  - {name: ok, models: [...], join_type: MANY_TO_ONE, condition: ...}

load_relationships() -> [{'name': 'ok', ...}]       # the two junk entries dropped
validate_project()   -> no message mentions them at all
```

Same repo, same file type, same class of bad data, opposite user experience.

### What to do

#2613 and #2607 are the two halves of one change and should be a single PR — the relationships counterpart of 9bdae39c:

- `load_relationships` filters to `list[dict]` (#2613 already does this)
- `validate_project` reports **both** `'relationships' must be a list` and `relationship entry must be a mapping, got <type>` — #2607 currently covers only the `models` field, not the entry itself

Note the ordering trap: with #2613's filter in place, `validate_project` never sees the junk entries, so the reporting must read the **raw** YAML the way 9bdae39c does — not the loader output.

That pair also makes #2606 (`memory/seed_queries.py`) and #2608 (`mcp_server.py`) unnecessary: once `context.py` normalises relationships to `list[dict]`, re-guarding at each consumer is exactly the downstream duplication #2613's own comment argues against ("so every consumer trusts the contract without re-filtering at each callsite"). Please close those two rather than landing three copies of one guard — the contribution bar asks for a mechanical change repeated across files to come as one PR.

### Rebase

Bartok9:

PR #2608, opened 2026-07-28
fix(mcp): skip non-list relationship models in describe_model

Bartok9:
## Summary
`describe_model` only matches relationships when `models` is a real list, so a string value cannot substring-match a model name.

## Motivation
Related to #2590. `"orders" in "orders_customers"` is True in Python — non-list models would attach junk relationships.

## Real behavior proof
```
pytest tests/unit/test_mcp_server.py::test_describe_model_skips_non_list_relationship_models -q
# 1 passed
```

## Test plan
- [x] string models skipped; valid list kept

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model relationship details to ignore malformed relationship data.
  * Prevented incorrect relationship matches caused by string values being interpreted as model lists.

* **Tests**
  * Added coverage to verify that only valid, list-based model relationships are displayed.

### DIFF
--- a/core/wren/src/wren/mcp_server.py
+++ b/core/wren/src/wren/mcp_server.py
@@ -262,11 +262,15 @@ def _register_context_tools(mcp: FastMCP, ctx: ServeContext) -> None:
             for col in model.get("columns", []) or []
         ]
 
-        relationships = [
-            rel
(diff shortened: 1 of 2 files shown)


Bartok9:

PR #2612, opened 2026-07-28
refactor(core): avoid throwaway ModelPlanNode build for aliased model scans

### LINKED ISSUE #2451: refactor(core): avoid throwaway ModelPlanNode build for aliased model scans
## Background

Follow-up to #2449, which fixed CLS (column-level access control) incorrectly denying aliased-scan and `count(*)` queries by pruning inaccessible columns in the wildcard-expansion loop of `ModelSourceNode::new`.

That fix is correct and complete for the reported bug. This issue tracks a **separate structural inefficiency** it surfaced — no behavior change, purely wasteful work in the analyzer.

## Problem

For an aliased model scan such as `SELECT e.c_custkey FROM customer e`, the plan is:

```
Projection → SubqueryAlias(e) → TableScan(customer)
```

Because `ModelAnalyzeRule` uses a **bottom-up** transform, the bare `TableScan(customer)` is visited *before* its wrapping `SubqueryAlias(e)`:

1. In `analyze_table_scan`, the bare scan is looked up with `alias = None`, so `table_ref = customer`. But the scope keys required columns under the alias `e`, so `try_get_required_columns(scope, customer)` misses and falls through to the empty `field = vec![]` branch (`model_anlayze.rs`).
2. Empty required fields trigger the wildcard expansion in `ModelSourceNode::new`, building a full `ModelPlanNode` for the model.
3. That node is then **discarded** — `analyze_subquery_alias_model` rebuilds the node correctly using the alias `e`'s actual required columns.

So every aliased scan does a `build → discard → rebuild`: one wasted wildcard-expanded `ModelPlanNode` per aliased model reference.

## Note

This is distinct from the genuine `count(*)` case, which legitimately has empty required fields and *must* go through the wildcard-prune path — so this cleanup does **not** remove the need for #2449's fix.

## Possible directions

- Avoid materializing the throwaway node when a bare `TableScan` is going to be immediately wrapped and rebuilt by a `SubqueryAlias`.
- Or resolve the alias earlier so the bare-scan lookup finds the correct required columns on the first pass.

Either way this is an **efficiency-only** change: it must not alter query results or access-control behavior, and #2449's regression tests should continue to pass unchanged.

AmirF194:
## Summary

For `FROM customer AS e`, the bottom-up `ModelAnalyzeRule` walk visits the bare
`TableScan` before its `SubqueryAlias` parent, so it looks up required columns
under the alias, finds none, and builds a full wildcard-expanded `ModelPlanNode`
that `analyze_subquery_alias_model` immediately discards and rebuilds correctly
once the alias is known. This skips that first, discarded build for the common
`SubqueryAlias -> TableScan` shape, via a pre-order rewriter pass that builds the
node once with the alias's real required columns.

No behavior change: same plan, same SQL output, one fewer RLAC/CLAC parse per
aliased model scan. `refactor:`, not `fix:`.

N/A, no failure. This is a `refactor:`: it removes a wasted build, it does not
change the query result. PR #2449 already fixed the correctness bug the
throwaway build used to cause (wildcard-expansion access denial on unreferenced
CLS columns); that fix is untouched by this change, it lives in
`ModelSourceNode::new` (plan.rs), which this PR does not modify.

## How is it tested?

- New test `test_aliased_model_scan_builds_model_plan_node_once`
  (`core/src/mdl/mod.rs`) asserts, via a `#[cfg(test)]` call counter on
  `build_model_plan_node`, that an aliased model scan builds it once. Confirmed
  it fails on current `main` (counts 2) and passes with this change.
- Full `wren-semantic-core` suite (148 tests, including `sqllogictest`'s
  `model.slt`/`type.slt`/`view.slt`/`tpch.slt`) passes unchanged.
- `cargo check --all-targets`, `cargo clippy --all-targets --all-features -- -D
  warnings`, and `cargo fmt --check` are all clean.
- Not checked: performance impact was not benchmarked; the claim is one fewer
  build call, not a measured latency number.

## Duplicate check

Searched open PRs (`is:pr is:open`) for `ModelPlanNode`, `analyze_model`, and
this issue number in the body: none touch `model_anlayze.rs` or reference
#2451. `git log -G` on the touched functions shows no prior attempt at this
specific change.

Fixes #2451

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved query analysis for aliased models, including nested subqueries and joins.
  - Correctly resolves required fields for column selections and `count(*)` queries.
  - Preserved validation errors for unknown tables and prevented redundant processing.
  - Ensured transformed queries produce valid, executable SQL.
  - Improved pruning of unreferenced protected columns across supported query patterns.

- **Tests**
  - Added regression coverage for aliased model selections, counts, joins, SQL execution, and transformed query results.

### DIFF
--- a/core/wren-core/core/src/logical_plan/analyze/model_anlayze.rs
+++ b/core/wren-core/core/src/logical_plan/analyze/model_anlayze.rs
@@ -77,39 +77,35 @@ struct ModelRewriter<'a> {
     scope_manager: &'a mut ScopeManager,
     current_scope_id: ScopeId,
     cycle_stack: &'a ModelStack,
-    /// Set by `f_down` when the pre-order shortcut fires, so `f_up` skips the
-    /// bottom-up pass for that node. Correct only because a `Jump` from `f_down`
-    /// still runs `f_up` for the same node (`transform_children` turns `Jump` into `Continue`).
-    shortcut_taken: bool,
 }
 
 impl TreeNodeRewriter for ModelRewriter<'_> {
     type Node = LogicalPlan;
 
     fn f_down(&mut self, plan: LogicalPlan) -> Result<Transformed<LogicalPlan>> {
-        let result = self.rule.shortcut_aliased_table_scan(
+        self.rule.shortcut_aliased_table_scan(
             plan,
             self.scope_manager,
             self.current_scope_id,
             self.cycle_stack,
-        )?;
-        self.shortcut_taken = result.tnr == TreeNodeRecursion::Jump;
-        Ok(result)
+        )
     }
 
     fn f_up(&mut self, plan: LogicalPlan) -> Result<Transformed<LogicalPlan>> {
-        let plan = if std::mem::take(&mut self.shortcut_taken) {
-            plan
-        } else {
-            self.rule
-                .analyze_model_internal(
-                    plan,
-                    self.scope_manager,
-                    self.current_scope_id,
-                    self.cycle_stack,
-                )?
-                .data
-        };
+        // Always run the bottom-up pass, even on a node the shortcut just built
+        // (SubqueryAlias -> Extension(ModelPlanNode)): analyze_model_internal's
+        // SubqueryAlias arm only matches an immediate SubqueryAlias or TableScan
+        // input, so it falls to the `_` arm and reconstructs the same node via
+        // SubqueryAlias::try_new, a no-op.
+        let plan = self
+            .rule
+            .analyze_model_internal(
+                plan,
+                self.scope_manager,
+                self.current_scope_id,
+                self.cycle_stack,
+            )?
+            .data;
         // If the plan contains subquery, we should analyze the subquery recursively
         plan.map_subqueries(|plan| {
             if let LogicalPlan::Subquery(subquery) = &plan {
@@ -435,7 +431,6 @@ impl ModelAnalyzeRule {
             scope_manager,
             current_scope_id,
             cycle_stack,
-            shortcut_taken: false,
         };
         plan.rewrite(&mut rewriter)
     }
--- a/core/wren-core/core/src/logical_plan/analyze/model_anlayze.rs
+++ b/core/wren-core/core/src/logical_plan/analyze/model_anlayze.rs
@@ -38,10 +38,12 @@ thread_local! {
 ///
 /// There are three main steps in this rule:
 /// 1. Analyze the scope of the logical plan and collect the required columns for models and visited tables. (button-up and depth-first)
-/// 2. Analyze the model and generate the ModelPlanNode according to the scope analysis. (button-up and depth-first)
+/// 2. Analyze the model and generate the ModelPlanNode according to the scope analysis. (button-up and depth-first, except the `SubqueryAlias -> TableScan` shortcut in `f_down`, which handles that shape pre-order and skips its subtree)
 /// 3. Remove the catalog and schema prefix of Wren for the column and refresh the schema. (top-down)
 ///
-/// The traverse path of step 1 and step 2 should be same.
+/// Step 2's shortcut may skip a subtree only when no skipped node can carry a subquery
+/// (a bare `TableScan` never does: filter pushdown is an optimizer rule, so `filters`
+/// is always empty at analyzer time). Elsewhere, the traverse path of step 1 and step 2 should be same.
 /// The corresponding scope will be pushed to or popped from the childs of [Scope] sequentially.
 pub struct ModelAnalyzeRule {
     analyzed_wren_mdl: Arc<AnalyzedWrenMDL>,
@@ -701,7 +703,7 @@ impl ModelAnalyzeRule {
 
     /// Pre-order shortcut for `SubqueryAlias -> TableScan` (e.g. `FROM a_model AS alias`).
     /// Builds the `ModelPlanNode` directly with the alias's own required columns,
-    /// mirroring [`Self::analyze_subquery_alias_model`], instead of the bottom-up pass.
+    /// replacing the bottom-up path for this shape entirely (see [`Self::analyze_subquery_alias_model`]).
     fn shortcut_aliased_table_scan(
         &self,
         plan: LogicalPlan,
@@ -786,51 +788,28 @@ impl ModelAnalyzeRule {
         }
     }
 
-    /// Because the bottom-up transformation is used, the table_scan is already transformed
-    /// to the ModelPlanNode before the SubqueryAlias. We should check the patten of Wren-generated model plan like:
-    ///      SubqueryAlias -> SubqueryAlias -> Extension -> ModelPlanNode
-    /// to get the correct required columns
+    /// Rebuilds a `SubqueryAlias -> SubqueryAlias -> Extension -> ModelPlanNode` shape.
+    /// `shortcut_aliased_table_scan` (`f_down`) now intercepts every aliased `TableScan`
+    /// before that nested shape can be produced, so the branch below is defense-in-depth.
     fn analyze_subquery_alias_model(
         &self,
         subquery_alias: SubqueryAlias,
-        scope_manager: &mut ScopeManager,
-        current_scope_id: ScopeId,
+        _scope_manager: &mut ScopeManager,
+        _current_scope_id: ScopeId,
         alias: TableReference,
-        cycle_stack: &ModelStack,
+        _cycle_stack: &ModelStack,
     ) -> Result<Transformed<LogicalPlan>> {
         let SubqueryAlias { input, .. } = subquery_alias;
         if let LogicalPlan::Extension(Extension { node }) =
             Arc::unwrap_or_clone(Arc::clone(&input))
         {
             if let Some(model_node) = node.as_any().downcast_ref::<ModelPlanNode>() {
-                if let Some(model) = self
-                    .analyzed_wren_mdl
-                    .wren_mdl()
-                    .get_model(model_node.plan_name())
-                {
-                    let field = self.resolve_required_fields(
-                        scope_manager,
-                        current_scope_id,
-                        &alias,
-                    )?;
-                    let model_plan_node = self.build_model_plan_node(
-                        Arc::clone(&model),
-                        field,
-                        None,
-                        cycle_stack,
-                    )?;
-                    let model_plan = LogicalPlan::Extension(Extension {
-                        node: Arc::new(model_plan_node),
-                    });
-                    let subquery =
-                        LogicalPlanBuilder::from(model_plan).alias(alias)?.build()?;
-                    Ok(Transformed::yes(subquery))
-                } else {
-                    internal_err!(
-                        "Model {} not found in the WrenMDL",
-                        model_node.plan_name()
-                    )
-                }
+                internal_err!(
+                    "SubqueryAlias wrapping an already-analyzed ModelPlanNode ({}) \
+                     reached the bottom-up rebuild path; shortcut_aliased_table_scan \
+                     should have handled this shape in f_down",
+                    model_node.plan_name()
+                )
             } else {
                 internal_err!("ModelPlanNode not found in the Extension node")
             }
--- a/core/wren-core/core/src/logical_plan/analyze/model_anlayze.rs
+++ b/core/wren-core/core/src/logical_plan/analyze/model_anlayze.rs
@@ -4,7 +4,9 @@ use crate::logical_plan::utils::{belong_to_mdl, expr_to_columns};
 use crate::mdl::context::SessionPropertiesRef;
 use crate::mdl::utils::quoted;
 use crate::mdl::{AnalyzedWrenMDL, Dataset, SessionStateRef};
-use datafusion::common::tree_node::{Transformed, TransformedResult, TreeNode};
+use datafusion::common::tree_node::{
+    Transformed, TransformedResult, TreeNode, TreeNodeRecursion, TreeNodeRewriter,
+};
 use datafusion::common::{internal_err, plan_err, Column, DFSchemaRef, Result, Spans};
 use datafusion::config::ConfigOptions;
 use datafusion::error::DataFusionError;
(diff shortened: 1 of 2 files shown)


goldmedal (commented):
**Sound change — verified no behavior change and the build-count win holds.** I diffed generated SQL across 19 plan shapes against `b590b650` (identical), instrumented build counts (halved for aliased scans, never increased), and ran all CI gates locally including `core-py-ci`. Comments below are cleanups, none blocking.

### Please address in this PR

**1. `test_clac_unreferenced_column_pruned_not_denied` — update the stale comment.** It documents the exact mechanism this PR deletes:

> *"even when the model is referenced with a table alias (which builds a throwaway inner plan with empty required fields, so the model gets wildcard-expanded)"*

After this change the aliased-select half no longer wildcard-expands — only the `count(*)` half does. This is the PR that invalidates the comment, and it's the #2449 guard, so it shouldn't be left describing gone behavior.

**2. Add a snapshot assertion to the new test.** `assert_sql_valid_executable` only proves the SQL parses and runs — it would pass on a plan that silently pruned or added a column, which is the actual risk being claimed against. The file uses `assert_snapshot!` throughout; one here would make the test assert the no-behavior-change claim rather than just the call count.

**3. Cover the two branches the new test misses.** Both are one-liners with the existing helpers:

- `count(*)` on an aliased model — the only path reaching the `visited_dataset` fallback (`field = vec![]`), and the only remaining wildcard-expansion path for aliased scans.
- A join of two aliased models — the only thing exercising `shortcut_taken` across siblings.

### Optional

**4. Drop the needless `SubqueryAlias` rebuild on non-matching paths.** The three early returns do `Transformed::no(SubqueryAlias::try_new(input, alias)?)`, so every non-model `SubqueryAlias` in every plan pays a schema recompute in `f_down` to hand back an unchanged node. Match on `&plan` first:

```rust
let LogicalPlan::SubqueryAlias(SubqueryAlias { input, alias, .. }) = &plan else {
    return Ok(Transformed::no(plan));
};
let LogicalPlan::TableScan(scan) = input.as_ref() else {
    return Ok(Transformed::no(plan));
};
if !belong_to_mdl(&self.analyzed_wren_mdl.wren_mdl(), scan.table_name.clone(),
                  Arc::clone(&self.session_state)) {
    return Ok(Transformed::no(plan));
}
let Some(model) = self.analyzed_wren_mdl.wren_mdl.get_model(scan.table_name.table()) else {
    return Ok(Transformed::no(plan));
};
```

`original_table_scan` is `None`, so `input` never needs consuming — this also removes the `unreachable!()` and the `Arc::unwrap_or_clone`.

**5. Comment the invariant `shortcut_taken` depends on.** It's correct only because `Jump` from `f_down` still runs `f_up` for that node (`transform_children` rewrites `Jump`→`Continue`, then `transform_parent` calls `f_up`). I confirmed this holds in the pinned DataFusion, but if a bump changed it the failure is silent — a parent's `analyze_model_internal` gets skipped. One line on the field would make the coupling explicit.

**6. Extract the block duplicated from `analyze_subquery_alias_model`.** The required-columns lookup + `visited_dataset` fallback + build + alias-wrap is copied verbatim, including the `internal_err!` text. A helper over `(model, alias, scope_manager, current_scope_id)` would stop the two drifting.

**7. Nit.** The counter's doc comment explains per-test threads, but the real dependency is `#[tokio::test]`'s current-thread runtime — under `flavor = "multi_thread"` the increments land elsewhere and the assert reads `0`.

AmirF194:
Thanks for the thorough review, especially catching the stale comment and the missing branches. Pushed `fd2665d2` addressing all three required items plus the optional cleanups.

**1. Stale comment.** Updated `test_clac_unreferenced_column_pruned_not_denied` in `core/wren-core/core/src/mdl/mod.rs` to describe pruning across scan shapes generically instead of naming the throwaway-plan mechanism this PR removes.

**2. Snapshot assertion.** Added `assert_snapshot!(actual, ...)` to the existing aliased-scan case in `test_aliased_model_scan_builds_model_plan_node_once` (`core/wren-core/core/src/mdl/mod.rs`), so the test now asserts the generated SQL shape, not just the build-call count.

**3. Missing branches.** Added the two cases to the same test: `count(*)` on an aliased model (asserts one build via the `visited_dataset` fallback) and a join of two aliased models (asserts two builds, exercising `shortcut_taken` across siblings). Both carry `assert_sql_valid_executable` and `assert_snapshot!`.

**4. SubqueryAlias rebuild.** Applied your replacement in `shortcut_aliased_table_scan` (`model_anlayze.rs`): matches on `&plan` first, returns `Transformed::no(plan)` on every early exit, and the `unreachable!()` / `Arc::unwrap_or_clone` are gone. Since `original_table_scan` stays `None` here, `input` is never consumed, so I clone just `alias` before building the field lookup rather than re-taking ownership of `plan`.

**5. `shortcut_taken` invariant.** Added a comment on the field in `ModelRewriter` naming the `Jump`-still-runs-`f_up` dependency.

**6. Extracted the duplicated block.** Pulled the required-columns/`visited_dataset`-fallback lookup into `resolve_required_fields(scope_manager, current_scope_id, table_ref)`, used by `shortcut_aliased_table_scan`, `analyze_table_scan`, and `analyze_subquery_alias_model`.

**7. Counter doc comment.** Fixed to name `#[tokio::test]`'s current-thread runtime as the real reason, not "one thread per test".

Re-verified after all changes: `cargo test --lib --tests --bins` in the pinned `rust:1-bookworm` container (148/148 passed, same count as before since the new assertions live inside the existing test function), `cargo clippy --all-targets --all-features -- -D warnings` clean, `cargo fmt --check` clean.

goldmedal:
Reviewed locally at `fd2665d2`. The refactor holds up — I A/B'd the shortcut behind a temporary kill-switch across 10 query shapes (plain aliased scan, `count(*)`, join of two aliased models, un-aliased scan, `IN`/`EXISTS`/correlated-scalar subqueries, nested alias, `e.*`) and the generated SQL was byte-identical in every case, with `build_model_plan_node` calls going 2→1 for a single aliased scan and 4→2 for the join and subquery shapes.

One non-blocking suggestion: **`shortcut_taken` looks like dead weight, and dropping it would remove the fragile coupling it documents.**

I tested making `f_up` call `analyze_model_internal` unconditionally: build counts and generated SQL were unchanged across all 10 probe queries, and the full 148-test suite still passed. The reason is that once the shortcut fires, the node is `SubqueryAlias -> Extension(ModelPlanNode)`, which falls into `analyze_model_internal`'s `_ =>` arm and is merely reconstructed via `SubqueryAlias::try_new` — a no-op.

Removing the field would delete the stated dependency on "a `Jump` from `f_down` still runs `f_up`" (true in DataFusion 53 — `transform_children` rewrites `Jump` to `Continue` and `transform_parent` then runs `f_up` — but it is an implementation detail to be tied to), along with the cross-sibling leak hazard that the join case exists to guard against.

If you'd rather keep the flag — it does avoid one redundant `SubqueryAlias::try_new` per aliased scan — then that trade-off is worth stating as the reason, since correctness doesn't appear to require it.

AmirF194:
Thanks for the thorough A/B, that kind of verification is exactly what makes a suggestion like this easy to trust.

Read through `analyze_model_internal`'s match arms and confirmed the same thing: once the shortcut builds the node, it's `SubqueryAlias -> Extension(ModelPlanNode)`, which only matches the `_` catch-all and gets rebuilt via `SubqueryAlias::try_new`, a no-op. Dropped `shortcut_taken` and made `f_up` call `analyze_model_internal` unconditionally. Full suite (149 tests), clippy, and fmt all clean in a fresh container against the new commit.

Also rebased onto current main to clear the conflict, force-pushed to the same branch (33c3eb67). Let me know if anything else needs a look.

core/wren-core/core/src/logical_plan/analyze/model_anlayze.rs
```diff
@@ -698,23 +808,11 @@ impl ModelAnalyzeRule {
...
-                                    "Table {} not found in the visited dataset and required columns map",
-                                    alias);
-                        };
-                        vec![]
-                    };
+                    let field = self.resolve_required_fields(
+                        scope_manager,
+                        current_scope_id,
+                        &alias,
+                    )?;
```
goldmedal: **This branch is now unreachable, so this `resolve_required_fields` call site never runs.**

I instrumented the `if let Some(model_node) = ...downcast_ref::<ModelPlanNode>()` branch with an `eprintln!` and ran the full suite on both sides: **base hits it 16 times** (13× `customer`, 3× `orders`), **this branch hits it 0 times**.

That matches the structure rather than being a test-coverage artifact. Reaching here needs `SubqueryAlias -> SubqueryAlias(model) -> Extension`, and the inner alias is only ever produced by `analyze_table_scan`'s `.alias(quoted(model.name()))` or by the new shortcut. The shortcut's predicates (`belong_to_mdl` + `get_model`) are exactly the ones under which `analyze_table_scan` would rewrite the scan, so any `SubqueryAlias -> TableScan` that could produce that shape is consumed in `f_down` first and never reaches the bottom-up pass.

So the shortcut does not pre-empt the bottom-up path for this shape — it *replaces* it. Two consequences worth resolving:

1. Either remove this branch, or keep it and add a comment naming the shape that is expected to reach it. As it stands a future reader will assume it is live.
2. The doc on `shortcut_aliased_table_scan` ("mirroring [`Self::analyze_subquery_alias_model`], instead of the bottom-up pass") reads as if the old path stays in service. Worth rewording either way.

Separately, and not caused by this PR: `analyze_model_internal`'s `SubqueryAlias -> TableScan` arm gets 0 hits on **both** base and head. That one is pre-existing dead code.

core/wren-core/core/src/logical_plan/analyze/model_anlayze.rs
```diff
@@ -21,6 +23,15 @@ use std::collections::HashSet;
...
+#[cfg(test)]
+thread_local! {
+    /// Counts [`ModelAnalyzeRule::build_model_plan_node`] calls; thread-local because
+    /// `#[tokio::test]`'s default current-thread runtime keeps caller and counter on
+    /// the same thread (a `flavor = "multi_thread"` test would need something else).
+    pub(crate) static BUILD_MODEL_PLAN_NODE_CALLS: std::cell::Cell<usize> =
+        const { std::cell::Cell::new(0) };
+}
+
 /// [ModelAnalyzeRule] responsible for analyzing the model plan node. Turn TableScan from a model to a ModelPlanNode.
```
goldmedal, line 35: **This contract comment is now false** — specifically the two lines just below:

```
/// 2. Analyze the model and generate the ModelPlanNode ... (button-up and depth-first)
/// The traverse path of step 1 and step 2 should be same.
```

Step 2 is now a mixed pre/post-order rewrite that skips a subtree, so the two paths are no longer the same.

That sentence is not decoration — it is the stated invariant behind the sequential pairing of `create_child_scope` (step 1) with `pop_child_scope` (step 2's `f_up`). If those desynchronise, the result is silently mis-resolved required columns, not an error, so the comment is doing real work.

The skip is safe **today** only because the single node it skips is a bare `TableScan`: `map_subqueries` reaches subqueries through `map_expressions`, which for a `TableScan` means `filters`, and those are always empty at analyzer time (filter pushdown is an optimizer rule, and analyzer rules run first).

Please restate the invariant in that narrower form — something to the effect that step 2 may skip a subtree only when no skipped node can carry a subquery. Otherwise a future shortcut over a subquery-carrying node inherits a comment that says the traversal is symmetric when it no longer is.

core/wren-core/core/src/mdl/mod.rs
```diff
@@ -3260,12 +3346,9 @@ mod test {
...
     async fn test_clac_unreferenced_column_pruned_not_denied() -> Result<()> {
-        // A CLS-protected column the query does NOT reference must be pruned,
-        // not denied — even when the model is referenced with a table alias
-        // (which builds a throwaway inner plan with empty required fields, so the
-        // model gets wildcard-expanded), and for `count(*)`. Before the fix, an
-        // aliased scan of `customer` — or `count(*)` — that never selected
-        // `c_name` still denied `customer.c_name`.
+        // A CLS-protected column the query does NOT reference must be pruned, not
+        // denied, in every scan shape (bare, aliased, count(*), wildcard). Before
+        // the fix, an aliased or count(*) scan of `customer` still denied `c_name`.
```
goldmedal, line 3351: Good fix here — but **its production-code counterpart is still stale.** `ModelSourceNode::new` in `logical_plan/analyze/plan.rs` (around lines 1085-1090) still documents the trigger for the wildcard-prune branch as:

```rust
// ... This is an *implicit* all-columns
// expansion (a `count(*)`, or the throwaway inner plan built for
// a model scan referenced with a table alias, whose required
// columns are keyed by the alias so the model is wildcard-
// expanded). ...
```

This PR deletes that second trigger. Afterwards only the empty-required-fields case (`count(*)`-shaped) reaches the branch, so the parenthetical now describes a mechanism that no longer exists — while sitting on top of the CLS prune-vs-deny decision from #2449, which is exactly the code someone will read to understand that behaviour.

`plan.rs` isn't in this diff, so it's a one-line comment edit to pull in.

goldmedal (commented):
Reviewed locally at `33c3eb67`. The refactor is correct and I found nothing blocking.

What I verified independently:

- **SQL parity.** Ran the three snapshot queries against the merge-base (`4483ce8c`); all three outputs are byte-identical to the inline snapshots in this PR.
- **Suite.** `RUST_MIN_STACK=8388608 cargo test --lib --tests --bins` → 149 passed, plus `model.slt` / `tpch.slt` / `type.slt` / `view.slt`. (Without `RUST_MIN_STACK` the run aborts on a stack overflow in `test_composite_key_calculation` — pre-existing, just noting it for anyone reproducing.)
- **The `Jump` claim.** Correct in DataFusion 53: `Transformed::transform_children` normalises `Jump` → `Continue` before `transform_parent` invokes `f_up`, so a `Jump` from `f_down` skips children and still runs `f_up`. Dropping `shortcut_taken` also means correctness no longer depends on that detail, which is a good outcome.

Three follow-ups inline. The first two are worth doing before merge; the third is a one-line comment edit in a file this PR doesn't currently touch.

AmirF194:
Thanks for the thorough review, all three addressed in e830636c.

1. Confirmed your instrumentation independently by reading `analyze_model_internal`'s match arms: `shortcut_aliased_table_scan` now intercepts every `SubqueryAlias -> TableScan` in `f_down` before the intermediate `SubqueryAlias(model) -> Extension(ModelPlanNode)` wrapping (that used to come from a bare, alias-less `analyze_table_scan` call) can ever be produced, so the nested pattern `analyze_subquery_alias_model`'s ModelPlanNode branch rebuilds never arises anymore. Went with defense-in-depth over deletion: turned that branch into an `internal_err!` naming the invariant, so a future change that reopens this path fails loudly instead of silently building a possibly-wrong plan.
2. Reworded the shortcut's doc comment and the rule-level step 1/2 doc to state the narrower invariant: a skipped subtree is safe only because a bare `TableScan` can never carry a subquery (filter pushdown is an optimizer rule, so `filters` is always empty at analyzer time).
3. Left the `plan.rs` comment as-is per your own note that it's optional and out of this PR's file scope; happy to take it in a follow-up if you'd rather not carry it separately.

Re-verified in a fresh container: `cargo test --lib --tests --bins` 149/149, `cargo clippy --all-targets --all-features -- -D warnings` clean, `cargo fmt --check` clean.

goldmedal (approved):
Thanks @AmirF194 👍

PR #2613, opened 2026-07-29
fix(context): harden relationships load + validate like views

Bartok9:
## Summary
Relationships counterpart of 9bdae39c (#2604):
- `load_relationships` filters to `list[dict]` (drop non-mapping entries)
- `validate_project` re-reads **raw** `relationships.yml` and reports:
  - `'relationships' must be a list, got …`
  - `relationship entry must be a mapping, got …`

## Motivation
After #2604, views report hand-edited junk; relationships silently dropped it. Same class of bad data, opposite UX.

## Note on siblings
This supersedes the split approach in #2607 (reporting half only / models-field only). Downstream consumer re-guards #2606/#2608 become unnecessary once context normalizes the contract — closing those as follow-through.

## Verification
```bash
cd core/wren && python -m pytest tests/unit/test_context.py -q -k relationship
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for relationship configuration files.
  * Invalid relationship entries are now reported clearly instead of causing unexpected behavior.
  * Relationship collections with unsupported formats now produce validation errors.
  * Valid relationship definitions continue to load while malformed entries are safely ignored.
  * Validation now preserves the original location of entries, making configuration errors easier to identify and fix.

### DIFF
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -1150,10 +1150,24 @@ def validate_project(project_path: Path) -> list[ValidationError]:
     # load_relationships() silently drops those (matching the other loaders),
     # but validate_project reports hand-edited mistakes rather than letting
     # them vanish quietly — re-check the raw entries here.
+    # Also remember the raw list so the field checks below can keep file indices.
+    raw_relationships_list: list | None = None
     rel_file = project_path / "relationships.yml"
     if rel_file.exists():
         raw = yaml.safe_load(rel_file.read_text(encoding="utf-8")) or {}
+        if raw and not isinstance(raw, dict):
+            # Most likely hand-edit: bare list / scalar root (omitted `relationships:` key).
+            errors.append(
+                ValidationError(
+                    "error",
+                    "relationships.yml",
+                    "relationships.yml must be a mapping with a 'relationships' key, "
+                    f"got {type(raw).__name__}",
+                )
+            )
         raw_rels = raw.get("relationships") if isinstance(raw, dict) else None
+        if isinstance(raw_rels, list):
+            raw_relationships_list = raw_rels
         if raw_rels is not None and not isinstance(raw_rels, list):
             errors.append(
                 ValidationError(
@@ -1218,17 +1232,17 @@ def validate_project(project_path: Path) -> list[ValidationError]:
                     )
                 )
 
-    # Check relationships
+    # Check relationships — walk the raw list when available so indices match
+    # the file (filtered loader positions would renumber past dropped junk).
     all_entity_names = model_names | view_names
-    for i, rel in enumerate(relationships):
+    rel_entries = (
+        list(enumerate(raw_relationships_list))
+        if raw_relationships_list is not None
+        else list(enumerate(relationships))
+    )
+    for i, rel in rel_entries:
         if not isinstance(rel, dict):
-            errors.append(
-                ValidationError(
-                    "error",
-                    f"relationships[{i}]",
-                    "relationship entry must be an object",
-                )
-            )
+            # Non-mappings already reported from the raw pass above.
             continue
         rel_name = rel.get("name", f"relationships[{i}]")
         ref_models = rel.get("models") or []
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -703,12 +703,21 @@ def _cube_migration_target(cube: dict, source_file: str | None) -> tuple[str, st
 
 
 def load_relationships(project_path: Path) -> list[dict]:
-    """Load relationships from project_path/relationships.yml."""
+    """Load relationships from project_path/relationships.yml.
+
+    Non-list ``relationships`` values and non-mapping entries are dropped here
+    (matching ``_load_views_v1``). ``validate_project`` re-reads the raw YAML
+    so hand-edited mistakes are still reported rather than vanishing quietly.
+    """
     rel_file = project_path / "relationships.yml"
     if not rel_file.exists():
         return []
     data = yaml.safe_load(rel_file.read_text(encoding="utf-8")) or {}
-    return data.get("relationships", []) if isinstance(data, dict) else []
+    rels = data.get("relationships") if isinstance(data, dict) else None
+    # A bare ``relationships:`` parses to None and means "no relationships".
+    if not isinstance(rels, list):
+        return []
+    return [r for r in rels if isinstance(r, dict)]
 
 
 def load_instructions(project_path: Path) -> str | None:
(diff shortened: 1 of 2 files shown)


goldmedal:
The loader half here is right, but #2604 landed as 9bdae39c after this branch was opened and set the convention for exactly this situation — and this PR implements only half of it.

`_load_views_v1` drops malformed entries silently **and** `validate_project` reports them, so a hand-edit never vanishes without the user hearing about it. The merged comment states the rule explicitly:

Measured on `main` + this branch, with the same shape of bad data in each file:

**views.yml — reported (9bdae39c):**
```
[ERROR] views.yml > views[0]: view entry must be a mapping, got str
[ERROR] views.yml > views[1]: view entry must be a mapping, got int
```

**relationships.yml — silently gone:**
```
relationships:
  - not-a-mapping
  - 42
  - {name: ok, models: [...], join_type: MANY_TO_ONE, condition: ...}

load_relationships() -> [{'name': 'ok', ...}]       # the two junk entries dropped
validate_project()   -> no message mentions them at all
```

Same repo, same file type, same class of bad data, opposite user experience.

### What to do

#2613 and #2607 are the two halves of one change and should be a single PR — the relationships counterpart of 9bdae39c:

- `load_relationships` filters to `list[dict]` (#2613 already does this)
- `validate_project` reports **both** `'relationships' must be a list` and `relationship entry must be a mapping, got <type>` — #2607 currently covers only the `models` field, not the entry itself

Note the ordering trap: with #2613's filter in place, `validate_project` never sees the junk entries, so the reporting must read the **raw** YAML the way 9bdae39c does — not the loader output.

That pair also makes #2606 (`memory/seed_queries.py`) and #2608 (`mcp_server.py`) unnecessary: once `context.py` normalises relationships to `list[dict]`, re-guarding at each consumer is exactly the downstream duplication #2613's own comment argues against ("so every consumer trusts the contract without re-filtering at each callsite"). Please close those two rather than landing three copies of one guard — the contribution bar asks for a mechanical change repeated across files to come as one PR.

### Rebase

Bartok9:
Thanks @goldmedal — agreed on the #2604 pattern.

Rebased onto current `main` and folded both halves into this PR:
1. `load_relationships` → `list[dict]` (silent drop, same as views loaders)
2. `validate_project` re-checks **raw** YAML for non-list `relationships` and non-mapping entries (so junk is reported, not only filtered)

Also closing #2607 as the incomplete reporting-only half, and #2606/#2608 as redundant consumer re-guards once this lands.

goldmedal (commented):
**Approve with changes.** The core fix is correct and the motivation checks out. Three things should be cleaned up first; one is cosmetic-but-real, two are behavioral.

## What I verified

CI is green across all 8 jobs. `pytest tests/unit/test_context.py` → 132 passed locally.

The fix is genuinely worth having — I reproduced the before/after on `relationships: nope`:

```
MAIN:  load_relationships() -> 'nope'          # str, not list[dict]
       [ERROR] relationships[0]: relationship entry must be an object
       [ERROR] relationships[1]: ... [2] ... [3]     # iterating "n","o","p","e"

PR:    load_relationships() -> []
       [ERROR] relationships.yml > relationships: 'relationships' must be a list, got str
```

Four garbage errors from iterating a string's characters → one accurate one. The loader also now actually honors its `list[dict]` annotation, which is what makes the downstream re-guards unnecessary as claimed — `mcp_server.describe_model` (`rel.get("models")`, line 267) and `context_cli.py:553` both consume it unguarded today.

## 1. Duplicate imports in the test file — should fix

```python
from wren.context import (
    load_relationships,      # ← added
    validate_project,        # ← added
    UpgradeError,
    ...
    load_relationships,      # ← already there (line 25)
    ...
    validate_project,        # ← already there (line 30)
)
```

Both names were already imported. The follow-up commit `a54c844b` moved them inside the parens (addressing the earlier syntax catch) but didn't notice they were redundant. Net effect: `ruff check tests/unit/test_context.py` now fails **I001**.

CI won't catch this — the lint job runs `ruff check src/` only, not `tests/` — so it's not a merge blocker, but the file goes from lint-clean to not. Fix is deleting the two added lines.

## 2. The PR leaves dead code behind — should fix

`context.py:1223-1232`, inside the "Check relationships" loop:

```python
for i, rel in enumerate(relationships):
    if not isinstance(rel, dict):
        errors.append(ValidationError("error", f"relationships[{i}]",
                      "relationship entry must be an object"))
        continue
```

`relationships` now comes from the filtering loader, so this branch is unreachable. It also carries a *different* message for the same condition than the new block ("must be an object" vs "must be a mapping, got int"). The views precedent (9bdae39c) has no equivalent guard in its loop — removing this is what actually makes the two paths symmetric.

## 3. Entry indices no longer match the file — worth a look

Because filtering happens before the loop, indices in the main loop are now positions in the *filtered* list. Repro — file with `- 42` at index 0 and an unnamed relationship at index 1:

```
MAIN:  [ERROR]   relationships[0]: relationship entry must be an object
       [WARNING] relationships > relationships[1]: missing join_type      ← correct

PR:    [ERROR]   relationships.yml > relationships[0]: ... got int
       [WARNING] relationships > relationships[0]: missing join_type      ← now wrong
```

Two errors both labelled `relationships[0]`, pointing at different entries. #2604 has the same wart for views, so it's inherited rather than introduced here — but it bites harder in this file, since `f"relationships[{i}]"` is the fallback display name for real relationships that lack a `name`. Either enumerate raw and `continue` on non-dicts in the main loop, or carry a raw index on the filtered entries.

## 4. Nit — the one hand-edit that still vanishes quietly

A `relationships.yml` whose root is a bare list (i.e. someone omitted the `relationships:` key — plausibly the *most* likely hand-edit mistake) still produces zero diagnostics:

```
root is a bare list: load=[] relationship_errors=[]
root is a scalar:    load=[] relationship_errors=[]
```

The new block guards with `isinstance(raw, dict)` and then does nothing when that's false. Given the PR's stated premise, cheap to close:

```python
if raw and not isinstance(raw, dict):
    errors.append(ValidationError("error", "relationships.yml",
                  f"relationships.yml must be a mapping with a 'relationships' key, got {type(raw).__name__}"))
```

## On the outstanding CodeRabbit nit

It asks the not-a-list test to assert the concrete type. Agreed it's trivial — the other two tests already assert `got str` / `got int`; only `test_validate_project_reports_relationships_not_list` stops at the prefix.

## Not worth changing

`validate_project` now parses `relationships.yml` twice (once via `load_relationships` at line 948, once in the new block). Matches the views precedent, and the alternative is threading raw data through the loader — leave it.

Bartok9:
Thanks @goldmedal — addressed all four points on the latest push:

1. **Duplicate imports** — removed the redundant `load_relationships` / `validate_project` entries (ruff I001 clean).
2. **Dead non-dict branch** — removed the unreachable `"must be an object"` guard in the filtered loop (raw pass owns that report).
3. **File indices** — the field-check loop now enumerates the **raw** list when present, so junk at `[0]` no longer renumbers later unnamed relationships (e.g. missing `join_type` stays `relationships[1]`).
4. **Bare root** — non-mapping `relationships.yml` roots now error: `must be a mapping with a 'relationships' key, got …`.

Also tightened the CodeRabbit not-a-list assertion to `got str`, and added tests for bare-root + index stability. `pytest tests/unit/test_context.py -k relationship` → 8 passed.

Bartok9:
Thanks — tightened the index-stability test so one diagnostic must contain both `relationships[1]` and `join_type` (not split across errors). Pushed.

goldmedal (approved):
Thanks @Bartok9 👍

PR #2614, opened 2026-07-29
fix(context): normalise model columns in load_models

Bartok9:
## Summary
Normalise YAML-sourced model `columns` in `load_models` to `list[dict]`, and have `validate_project` report hand-edited non-list `columns` **and** non-object column entries (mirrors #2604 / 9bdae39c). GenBI inventory formatting no longer owns shape guards; it only stringifies column names safely.

Hand-edited model YAML can set `columns:` to a scalar or mix bare strings into the list. That used to flow through `load_models` and later crash consumers with `AttributeError: 'str' object has no attribute 'get'` — notably `wren serve mcp` → `describe_model` (`mcp_server.py`), which does `col.get(...)` per column. Inventory formatting had the same class of crash.

Measured against a hand-edited project (`models/b.yml` / metadata with `columns: id, customer_id`):

```
load_models -> [{'name': 'orders', 'columns': 'id, customer_id', ...}]
  any non-dict model?      False
  any non-list 'columns'?  True
```

## Motivation
Maintainer review: the reachable untrusted container is `columns`, but the fix belongs in the loader + `validate_project`, not a quiet consumer-side filter. Loader normalises; validate reports so hand-edits never vanish. That also removes the need for a duplicate columns guard in other consumers (e.g. the columns half of #2616).

## Changes
- `_normalize_model_columns` in `load_models` v1/v2: non-list → `[]`; drop non-dict list entries (no bare-string→name inference); omit key when absent (no empty `columns: []` injection)
- `validate_project` re-reads raw model YAML and errors:
  - `models/<…> > <name> > columns: must be a list, got <type>` (including explicit YAML null)
  - `models/<…> > <name> > columns[j]: column entry must be an object`
- Remove unreachable post-loader list/dict column checks (dead after normalisation)
- `_format_model_inventory`: keep `str(name)` so non-string YAML names cannot crash join; drop dead non-list guard
- Direct unit tests for loader/validate/inventory behaviour

## Test plan
- `test_load_models_v2_normalises_non_list_columns`
- `test_load_models_v2_omits_missing_columns_key`
- `test_load_models_v2_drops_non_dict_column_entries`
- `test_validate_project_reports_non_list_model_columns`
- `test_validate_project_reports_null_model_columns`
- `test_validate_project_reports_non_dict_column_entries`
- `test_validate_project_uses_dir_name_when_model_name_missing`
- `test_format_inventory_stringifies_non_string_column_names`
- `test_build_inventory_stringifies_non_string_column_names`

## Duplicate check
Related: closed #2548 (wrong layer), #2604 (views pattern), #2616 (consumer-side columns guard becomes unnecessary once loader owns shape). No open duplicate for loader-side model columns normalisation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of malformed model column definitions.
  - Invalid scalar values and non-mapping entries are safely ignored during model loading.
  - Validation now reports clear errors for improperly formatted column definitions.
  - Build instruction generation supports numeric column names without errors.

- **Tests**
  - Added coverage for invalid, mixed, valid, and numeric column configurations.
  - Added regression coverage for loading and validating malformed model metadata across supported layouts.


goldmedal:
The state guarded here is genuinely reachable, but the guard is in the wrong place — and #2604 landed as 9bdae39c after this branch was opened, setting the pattern for exactly this case.

### The premise checks out

`load_models` filters models but not their `columns`. Measured against a hand-edited project (`models/b.yml` containing `columns: id, customer_id`):

```
load_models -> [{'name': 'orders', 'columns': 'id, customer_id', '_source_dir': 'b'}]
  any non-dict model?      False    # _load_models_v1/v2 already do isinstance(data, dict)
  any non-list 'columns'?  True     # flows through untouched
```

So the comment in the diff is right: models are already dict-filtered, and only the `columns` container is untrusted. Good — that part is accurate and worth stating in the PR description with this output.

### But the fix belongs in the loader

9bdae39c established the convention for malformed project YAML: the loader drops or normalises, and `validate_project` is the single place that reports it to the user, so a hand-edit never vanishes silently. Its merged comment says so directly:

Guarding inside `_format_model_inventory` instead means:

- every other consumer of `model["columns"]` needs its own copy of the same check — #2616 already adds one in `_check_descriptions`
- the user is never told their `columns:` is malformed; the inventory just quietly renders fewer columns

Please move this to `load_models` (normalise `columns` to `list[dict]`) plus a `validate_project` error along the lines of `models/<name> > columns: must be a list, got str`, mirroring 9bdae39c. That also removes the need for the `columns` half of #2616.

One detail to keep either way: coercing a bare string element to a column name (`elif isinstance(c, str)`) is a new inference, not a fix — a YAML author writing `columns: [id, amount]` may well mean that, but it should be a stated decision rather than a side effect of a crash guard.

### Rebase

Bartok9:
Thanks @goldmedal — the loader/`validate_project` placement is the right fix; consumer-side quiet filtering was the wrong layer.

Will:
1. Rebase onto current `main` (post-9bdae39c).
2. Move normalisation into `load_models` (`columns` → `list[dict]` only; no silent bare-string→column-name inference unless we call that out as an explicit product decision — default is drop non-dict entries).
3. Add a `validate_project` error like `models/<name> > columns: must be a list, got <type>` so hand-edits don’t vanish.
4. Drop the formatter-only guard (and the columns half of #2616 becomes unnecessary once the loader owns shape).

Pushing that rewrite on this branch (or a clean replacement if history is cleaner that way) next.

goldmedal:
The fix itself is correct, and I confirmed the narrowing is justified: `_load_models_v1` and `_load_models_v2` both filter on `isinstance(..., dict)`, and `genbi/cli.py:137` is the only caller of `compose_build_instruction` — so the model-level guard from #2548 was indeed unreachable, and `columns` is the only untrusted container here. I also verified the new test fails on `main` (`AttributeError: 'str' object has no attribute 'get'`) and passes on this branch, so it does pin the crash.

**The test doesn't pin the filtering behavior it's meant to protect.** I mutated the fix two ways and all 12 tests stayed green:

1. Deleting the `elif isinstance(c, str) and c: names.append(c)` branch entirely → 12 passed.
2. Replacing it with `else: names.append(str(c))` — i.e. rendering junk instead of skipping it → 12 passed.

The cause is that `assert "id" in text` is satisfied by the static template regardless of the inventory: with `models=[]`, `"id" in text` is already `True`, because the template contains `outside`, `inside`, and `side`. So neither the `"bare"` inclusion nor the `3` exclusion is actually asserted, and the two behaviors this PR adds are both free to regress.

Asserting through the full instruction is what makes this hard. `_format_model_inventory` is the unit under change, so testing it directly allows exact assertions:

```python
from wren.genbi.composer import _format_model_inventory


def test_inventory_normalizes_non_list_columns() -> None:
    assert _format_model_inventory([{"name": "orders", "columns": "id, customer_id"}]) == "- **orders**: "


def test_inventory_keeps_dict_and_bare_string_columns_only() -> None:
    models = [{"name": "customers", "columns": [{"name": "id"}, "bare", 3]}]
    assert _format_model_inventory(models) == "- **customers**: id, bare"
```

Both of these fail under either mutation above. Keeping one `compose_build_instruction`-level test for the no-crash path is still worthwhile.

goldmedal:
@Bartok9, to clean my review queue, I turn it into a draft before addressing the review.

Bartok9:
Thanks @goldmedal — addressed on this branch (still draft while you clear queue).

1. **Rebased** onto current `main` (post-9bdae39c).
2. **Loader owns shape:** `_normalize_model_columns` in `load_models` v1/v2 → `columns` is always `list[dict]`. Non-list → `[]`; non-dict list entries dropped. **No** bare-string→column-name inference.
3. **`validate_project` reports** hand-edits: `models/<…> > <name> > columns: must be a list, got <type>` (raw YAML re-check, same spirit as views in 9bdae39c).
4. **Formatter** only defensively skips non-list / non-dict; no product inference.
5. **Tests:** direct `_format_model_inventory` assertions (empty non-list + dict-only), plus loader/validate coverage; kept one compose no-crash path.

Ready for re-review whenever you undraft / pick it back up.

goldmedal:
@Bartok9  please fix the ci fail

Bartok9:
@goldmedal Fixed the lint failure — `ruff format` on `context.py` (extra blank line before `_normalize_model_columns`) and `composer.py` (comprehension folded to one line). Pushed; CI should go green on the new commit.

Bartok9:
@coderabbitai Thanks — addressed the in-diff note on this branch:

- `validate_project` now reports every **present** non-list `columns` value, including explicit YAML null (`columns:`) → `must be a list, got NoneType`.
- Omitted `columns` key still skipped (unchanged).
- Regression: `test_validate_project_reports_null_model_columns`.

Skipped the out-of-diff `relationships.yml` `[]` root-type note — separate from this PR’s model-columns scope; happy to take that as a follow-up if you want it tracked.

CI was already green after the ruff format fix; this push should stay green.

goldmedal:
Right layer — thanks for moving it here, and for the repro and duplicate check in the description. I ran main (`9bdae39c`) and this head (`ea9c58d5`) side by side; one blocking issue plus some cleanup.

**Blocking: non-dict column entries go from reported to silently dropped.**

Project with `columns: [{name: id, type: INTEGER}, bare_column]`:

- before: `[ERROR] models/orders/metadata.yml > orders > columns[1]: column entry must be an object`
- after:  no error at all, and `bare_column` is silently gone from the manifest

`_normalize_model_columns` drops the entry in the loader, but the new raw re-read in `validate_project` only checks `not isinstance(raw_cols, list)` — it never checks the entries. That makes the pre-existing per-entry check at `context.py:1016` unreachable, so the user gets a manifest that is missing a column and passes validation. The docstring says "`validate_project` re-reads the raw file to report them", but only the container case is reported. Please extend the raw re-read to report per-entry problems too.

**Cleanup: two existing checks are now dead code.**

Both `isinstance(columns, list)` (~999) and `isinstance(col, dict)` (~1016) are unreachable once the loader normalises. This PR rewords the first one's message, but that line can no longer execute — the raw re-read is what fires. Please remove them rather than reword.

Smaller things:

- The `_format_model_inventory` guard is now dead: the only caller is `genbi/cli.py:143`, which passes `load_models(project_path)`. The description says inventory formatting "no longer owns the guard", but the diff keeps the `isinstance` filter and adds two tests that hand-build dicts `load_models` can no longer produce — that's the shape #2548 was closed for. Keep the `str(c.get("name", "?"))` change though: `- name: 3` survives the loader as a dict and raises `TypeError: sequence item 0: expected str instance, int found` on main. That's a real reachable bug and nothing in this PR tests it.
- Models with no `columns` key now get `"columns": []` injected into `target/mdl.json` (`model.get("columns", [])` defaults to a list). Unrelated output change; `if "columns" not in model: return model` avoids it.
- `mname = raw_model.get("name") or Path(src_path).stem` yields `metadata` on the v2 layout — the error reads `models/orders/metadata.yml > metadata > columns: …`. Use the directory name.
- `tests/unit/test_context.py` isn't ruff-formatted at this head: four blank lines before the `# ── Regression` banner, and zero between the new test's last assert and the following `def`. `ruff format --check` reformats it. Both the CI lint job and `just lint` only cover `src/`, so green CI isn't evidence here — the `style(wren): ruff format` commit only touched src.
- `_iter_raw_model_files` re-implements the v1/v2 layout dispatch and re-parses every model YAML. Two places to keep in sync when the layout changes; consider having the loader record the diagnostic on the model instead.

One suggestion for the description: the stronger motivation is `wren serve mcp` → `describe_model` (`mcp_server.py:250`), which does `col.get(...)` per column and raises `AttributeError` on main for a bare-string entry. That makes the loader-side fix more obviously correct than the inventory-formatting example.

Bartok9:
@goldmedal Thanks — addressed on this branch.

**Blocking**
- Raw re-read now reports per-entry problems too: `columns[j]: column entry must be an object` (so bare_column is no longer a silent drop).
- Regression: `test_validate_project_reports_non_dict_column_entries`.

**Cleanup**
- Removed the unreachable post-loader `isinstance(columns, list)` / `isinstance(col, dict)` error paths (shape is owned by loader + raw re-read).
- `_normalize_model_columns`: if `columns` key is absent, leave the model alone (no empty `columns: []` injection).
- v2 display name: `models/<dir>/metadata.yml` uses the directory name when YAML omits `name` (not stem `metadata`).
- `_format_model_inventory`: dropped the dead non-list / non-dict filter; kept `str(c.get("name", "?"))` and added tests for non-string names (`name: 3`).
- `ruff format` on the regression block in `tests/unit/test_context.py`.

Left `_iter_raw_model_files` as-is for this pass (still the simplest mirror of the views pattern); happy to fold diagnostics into the loader in a follow-up if you prefer.

Will nudge the PR description motivation toward `describe_model` / MCP next if useful.

goldmedal:
Re-reviewed at `25b2df0c` on a clean worktree (rebased — 0 commits behind `main`). The blocking item is resolved: a bare-string entry among dict columns now reports `models/orders/metadata.yml > orders > columns[1]: column entry must be an object`, identical to `main`, and it is reported exactly once. Locally `ruff format --check src/` and `ruff check src/` pass and `pytest tests/unit` (minus `test_memory.py`, which fails identically on `main`) is 1134 passed.

Two things left.

### 1. Half of the dead-code cleanup is still there

`context.py:1084-1086`:

```python
columns = model.get("columns", [])
if not isinstance(columns, list):   # unreachable — the loader guarantees a list
    columns = []
```

The `ValidationError` append was removed but the branch itself was kept. Since `load_models` now normalises `columns` to `list[dict]` and the key is either absent or a list, this condition can no longer be true and it no longer produces any output either. Please delete it — `model.get("columns", [])` is enough. The sibling `isinstance(col, dict)` check was removed correctly.

### 2. A v1 project now gets two different spellings of the same file in one run

Validating a v1 project with `models/orders.yml` containing `columns: id, customer_id`:

```text
head:
  [error] models/orders.yml > orders > columns: must be a list, got str        <- new raw re-read, correct path
  [error] models/orders/metadata.yml > orders: model must define either ...    <- existing loop, wrong path for v1
  [warning] models/orders/metadata.yml > orders: model has no columns

main:
  [error] models/orders/metadata.yml > orders: model must define either ...
  [error] models/orders/metadata.yml > orders: columns must be a list
```

The existing loop's `src_path = f"models/{src}/metadata.yml"` has always been wrong for the v1 flat layout — that part is pre-existing. But before this PR the two errors were at least consistently wrong; now the same file is addressed two different ways in the same output, which reads as two separate files. The new label is the correct one, so the better fix is to make the existing loop derive the v1 path from the real filename too (`_source_dir` is already in hand, only the layout branch is missing). Falling back to the old spelling for the new message would also remove the inconsistency, but keeps the wrong path.

Bartok9:
@goldmedal Thanks — both remaining items addressed on this branch.

1. **Dead branch removed:** dropped the unreachable `if not isinstance(columns, list): columns = []` after loader normalisation; `model.get("columns", [])` is enough.
2. **v1 path consistency:** the existing model loop now labels v1 as `models/<stem>.yml` (same as the raw re-read) instead of always `models/<src>/metadata.yml`, so a flat `orders.yml` no longer shows two spellings in one run.

Regression: `test_validate_project_v1_model_paths_use_flat_yml`.

goldmedal:
Re-checked at `f8026d5a`. Both items are fixed: the dead `isinstance(columns, list)` branch is gone, and v1 labels are now consistent — a flat `models/orders.yml` project reports all three diagnostics against `models/orders.yml`, and the new `test_validate_project_v1_model_paths_use_flat_yml` asserts the old spelling is absent. Locally `ruff format --check src/`, `ruff check src/` and `pytest tests/unit` (1135 passed) are clean.

One new problem, plus a rebase note.

### Blocking: column indices are renumbered past dropped entries

Model file with `[0]` a bare string, `[1]` a dict with no `name`, `[2]` a named column:

```yaml
columns:
  - bare_junk
  - {type: INTEGER}
  - {name: amt, type: DOUBLE}
```

```text
main:
  [error] models/orders/metadata.yml > orders > columns[0]: column entry must be an object
  [error] models/orders/metadata.yml > orders > columns[1]: column missing 'name'

this branch:
  [error] models/orders/metadata.yml > orders > columns[0]: column entry must be an object
  [error] models/orders/metadata.yml > orders > columns[0]: column missing 'name'
```

Two contradictory statements about `columns[0]` in one run. The raw re-read walks the file order, but the existing model loop walks the loader's filtered list, so every index after a dropped entry shifts down by one.

This is the same defect that was already fixed for relationships, and the reasoning is still in this function:

> Check relationships — walk the raw list when available so indices match the file (filtered loader positions would renumber past dropped junk).

`test_validate_project_relationship_indices_match_file` guards it there ("Junk at [0] must not renumber a later unnamed relationship's warnings"); nothing guards it for columns, which is why this passed. Please apply the same treatment — derive index-bearing column labels from the raw list (or have the loader preserve the original position) — and add the equivalent regression test.

### Rebase needed, and it is not just textual

#2616 (`b56d4da1`) and #2649 (`6db8bed5`) have both landed in `context.py`, so this branch is 2 commits behind. I trial-rebased: it applies cleanly and 1174 tests pass, but #2616's new column-properties loop then contains

```python
for j, col in enumerate(columns):
    if not isinstance(col, dict):
        continue
```

which becomes unreachable once this PR's loader normalisation is in — the sibling of the branch just deleted here. Please remove it as part of the rebase. (The index problem above exists both before and after the rebase; it is not caused by it.)

### Nit: the new v1 branch carries its own dead half

```python
src_path = f"models/{src}.yml" if not str(src).startswith("models[") else str(src)
```

`_load_models_v1` and `_load_models_v2` both always set `_source_dir`, so the `model.get("_source_dir", f"models[{i}]")` fallback never fires and the `startswith("models[")` half is unreachable. The v2 branch has no equivalent guard either, so the two are asymmetric. `str(src)` is also redundant — `f.stem` / `d.name` are already strings.

Bartok9:
@goldmedal Thanks — addressed on this branch (rebased onto current `main`, head includes #2616/#2649).

**Blocking — column indices**
- Model-loop (and column-properties) diagnostics now walk the **raw** columns list when available, same pattern as relationships, so junk at `[0]` no longer renumbers a later missing-`name` to `columns[0]`.
- Regression: `test_validate_project_column_indices_match_file` (bare junk + nameless dict + named col → object error on `[0]`, missing name on `[1]`).

**Rebase**
- Rebased cleanly onto `main` (pulls in #2616 / #2649).
- Removed the now-unreachable `if not isinstance(col, dict): continue` shape from the post-loader path in favor of raw-list walks (properties loop uses the same index source).

**Nit**
- Dropped the dead `startswith("models[")` half of the v1 `src_path` branch; always `models/{src}.yml` when `_source_dir` is set by the loader.

Happy to take another pass if anything else looks off.

goldmedal:
Re-checked at `002f4b26` (rebased, 0 behind `main`). The index fix works for the ordinary case — a file with `[0]` junk, `[1]` unnamed dict, `[2]` named now produces exactly the same labels as `main`, and `test_validate_project_column_indices_match_file` pins it including the negative assertion. `ruff format --check src/`, `ruff check src/` and `pytest tests/unit` (1175 passed) are clean locally.

Two problems, one of them outside what this PR has been looking at.

### Blocking: the v1→v2 upgrade now silently discards malformed columns

`_apply_v1_to_v2()` loads through `_load_models_v1()` — the loader this PR patches — so the normalisation now applies to migration, not just validation:

```text
source v1 project:
  models/orders.yml -> columns: id, customer_id
  models/items.yml  -> columns: [{name: id, type: INTEGER}, bare_column]

main, after `wren context upgrade`:
  models/orders/metadata.yml -> columns: id, customer_id
  models/items/metadata.yml  -> columns: [{name: id, type: INTEGER}, bare_column]

this branch, after `wren context upgrade`:
  models/orders/metadata.yml -> columns: []
  models/items/metadata.yml  -> columns: [{name: id, type: INTEGER}]
```

`_apply_v1_to_v2()` then `unlink()`s the source `models/*.yml`, so the original values only survive in git. Nothing is printed — no error, no warning — and the upgrade reports success.

Dropping this content in the *validation* path is the point of the PR and I agree with it. Doing it during a migration that deletes the source is a different decision, and right now it is being made implicitly. #2649 has just added a `_plan_v1_to_v2()` preflight at the top of `_apply_v1_to_v2()` precisely so bad input aborts before the first write; extending it to reject a malformed `columns` would fit that shape and keep the upgrade honest. Not normalising on the upgrade path also works. Either way it should be a stated decision with a test.

### `raw_model_columns` is keyed by model name, so duplicate names cross-wire

`raw_model_columns[mname] = raw_cols` is last-write-wins, and duplicate model names are a state `validate_project` itself reports, so this is reachable. Both directions are wrong — v2 layout, two dirs `a/` and `b/` both declaring `name: orders`:

**A real error disappears** (`a/` has `[0]` junk and `[1]` an unnamed dict, `b/` is clean):

```text
main:  [error] models/a/metadata.yml > orders > columns[0]: column entry must be an object
       [error] models/a/metadata.yml > orders > columns[1]: column missing 'name'
       [error] models/b/metadata.yml: duplicate model name 'orders'

head:  [error] models/a/metadata.yml > orders > columns[0]: column entry must be an object
       [error] models/b/metadata.yml: duplicate model name 'orders'
```

**A phantom error appears** (`a/` clean, `b/` holds the junk): head adds `[error] models/a/metadata.yml > orders > columns[1]: column missing 'name'` against a file whose single column is fine.

The v1 flat layout behaves identically. `_source_dir` is already on every loaded model and uniquely identifies the file, so keying the table on it instead of on `name` fixes both layouts at once and also removes the need for the dir-name/stem derivation of `mname`. Worth a regression test with two models sharing a name.

### Nits

- `col_entries` and `prop_entries` are the same five-line expression computed twice; compute it once above both loops.
- `_iter_raw_model_files()` is still the only nested `def` in `validate_project()`, still parses every model YAML a second time, and now also maintains a side table threaded back into the loaded loop by key. Three moving parts where one would do — having the loader record the diagnostic on the model would collapse all of them. Not blocking, but this is the third round spent on the seam between the two lists.

Separately, while reviewing this I hit a pre-existing crash in `validate_project` that is not caused by this PR (`main` behaves identically) — filed as its own issue. It does mean the new `raw_model_columns[mname]` line is a second site for it, worth keeping in mind when that one is fixed.

goldmedal:
@Bartok9 could you address the fix remain blocking items?

Bartok9:
@goldmedal Thanks — both remaining blocking items addressed on this branch (rebased onto current `main`).

**1. v1→v2 upgrade no longer silently discards malformed columns**
- Stated decision: loader normalisation stays for validate/runtime, but migration must not delete source after writing a stripped model.
- `_plan_v1_to_v2` preflight now calls `_reject_malformed_v1_model_columns` and raises `UpgradeError` for non-list `columns` or non-object entries (same shape language as validate).
- Regression: `test_plan_upgrade_v1_to_v2_rejects_malformed_model_columns` (sources unchanged; content still present).

**2. `raw_model_columns` keyed by source, not name**
- Table is now keyed by `_source_dir` / path stem (unique per file), so duplicate `name: orders` cannot drop real errors or invent phantom ones across dirs.
- Regression: `test_validate_project_duplicate_model_names_do_not_crosswire_columns`.

**Nit**
- Single `col_entries` list shared by the name and properties loops.

Happy to take another pass if anything else looks off.

Bartok9:
@goldmedal The two remaining blocking items from your 2026-08-17 re-check are on this head (`6a792454`), pushed just after your ping:

1. **v1→v2 upgrade** — `_plan_v1_to_v2` preflight now calls `_reject_malformed_v1_model_columns` and raises `UpgradeError` for non-list `columns` or non-object entries, so upgrade cannot unlink source after silently emptying columns.
2. **`raw_model_columns` keyed by `_source_dir`** — duplicate model names no longer cross-wire diagnostics.

Happy to take another pass if anything is still off at this SHA.

goldmedal (approved):

PR #2615, opened 2026-07-29
fix(cube): harden list against non-dict cube/measure rows

Bartok9:
`wren cube list` crashed on a malformed manifest. When a cube row was not a dict (e.g. `None`/string slipped into the `cubes` list), or when `measures`/`dimensions`/`timeDimensions` were non-list values, `list_cubes` assumed dict/list shapes and raised while iterating.

Observed error:
```
AttributeError: 'NoneType' object has no attribute 'get'
```
(also `TypeError` when a scalar field was iterated as a list).

## Fix
`list_cubes` now skips cube rows that aren't dicts and only extracts names from list-valued member fields, so a single bad row can't abort the whole listing.

## Verification
```
pytest tests/unit/test_cube_cli_list_guards.py -q
# 1 passed
```

## Duplicate check
Searched open/closed PRs for cube CLI list hardening — no existing/duplicate PR covers this guard.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the cube listing command’s resilience when manifests contain malformed or unexpected entries.
  * Invalid cube records are skipped, while incomplete measure, dimension, and time dimension data no longer causes failures.
  * Valid cube information continues to display correctly.

### DIFF
--- a/core/wren/src/wren/cube_cli.py
+++ b/core/wren/src/wren/cube_cli.py
@@ -174,9 +174,7 @@ def list_cubes(mdl: _MdlOpt = None) -> None:
             raw = cube.get(key, []) or []
             if not isinstance(raw, list):
                 return ""
-            return ", ".join(
-                m.get("name", "") for m in raw if isinstance(m, dict)
-            )
+            return ", ".join(m.get("name", "") for m in raw if isinstance(m, dict))
 
         measures = _names("measures")
         dims = _names("dimensions")
--- a/core/wren/src/wren/cube_cli.py
+++ b/core/wren/src/wren/cube_cli.py
@@ -165,13 +165,22 @@ def list_cubes(mdl: _MdlOpt = None) -> None:
         typer.echo("No cubes defined.")
         return
     for cube in cubes:
+        if not isinstance(cube, dict):
+            continue
         name = cube.get("name", "<unnamed>")
         base = cube.get("baseObject", "?")
-        measures = ", ".join(m.get("name", "") for m in cube.get("measures", []))
-        dims = ", ".join(d.get("name", "") for d in cube.get("dimensions", []))
-        time_dims = ", ".join(
-            td.get("name", "") for td in cube.get("timeDimensions", [])
-        )
+
+        def _names(key: str) -> str:
+            raw = cube.get(key, []) or []
+            if not isinstance(raw, list):
+                return ""
+            return ", ".join(
+                m.get("name", "") for m in raw if isinstance(m, dict)
+            )
+
+        measures = _names("measures")
+        dims = _names("dimensions")
+        time_dims = _names("timeDimensions")
         typer.echo(f"  {name} (base: {base})")
         if measures:
             typer.echo(f"    measures: {measures}")
--- a/core/wren/src/wren/cube_cli.py
+++ b/core/wren/src/wren/cube_cli.py
@@ -174,7 +174,11 @@ def list_cubes(mdl: _MdlOpt = None) -> None:
             raw = cube.get(key, []) or []
             if not isinstance(raw, list):
                 return ""
-            return ", ".join(m.get("name", "") for m in raw if isinstance(m, dict))
+            return ", ".join(
+                m["name"]
+                for m in raw
+                if isinstance(m, dict) and isinstance(m.get("name"), str)
+            )
 
         measures = _names("measures")
         dims = _names("dimensions")
--- a/core/wren/tests/unit/test_cube_cli_list_guards.py
+++ b/core/wren/tests/unit/test_cube_cli_list_guards.py
@@ -27,5 +27,7 @@ def test_list_cubes_skips_bad_rows(capsys):
     assert result.exit_code == 0
     out = result.stdout
     assert "orders" in out
+    assert "bad" not in out
+    assert "measures:" not in out
     assert "dimensions: id" in out
     assert "time dimensions: ts" in out
--- a/core/wren/tests/unit/test_cube_cli_list_guards.py
+++ b/core/wren/tests/unit/test_cube_cli_list_guards.py
@@ -1,0 +1,31 @@
+"""cube list must tolerate non-dict cubes and non-list measure arrays."""
+
+from types import SimpleNamespace
+from unittest.mock import patch
+
+from typer.testing import CliRunner
+
+from wren.cube_cli import cube_app
+
+
+def test_list_cubes_skips_bad_rows(capsys):
+    manifest = {
+        "cubes": [
+            "bad",
+            {
+                "name": "orders",
+                "baseObject": "o",
+                "measures": "nope",
+                "dimensions": [{"name": "id"}],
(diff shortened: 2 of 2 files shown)


Bartok9:
Thanks @coderabbitai — both applied in bb3b7ac: `_names` now skips entries whose `name` isn't a string (guards against `name: None` → TypeError), and the test adds negative assertions (`"bad" not in out`, `"measures:" not in out`) to enforce the skip behavior.

goldmedal:
Reviewing this together with #2617 — the CLI half of one change.

#2615 and #2617 harden the same cube listing behind two entry points — the `wren cube list` CLI and the `list_cubes` MCP tool. They should be one PR, because right now each covers a case the other misses and one of them guards states its data source cannot produce.

### The two data sources differ, and that decides which guards are live

**MCP `list_cubes` reads `context.load_cubes`**, which already normalises. Both `_load_cubes_v1` and `_load_cubes_v2` do `if isinstance(data, dict)` before appending, and both always return a list. Measured against hand-edited cube YAML (a file that is a bare scalar, a file that is a list, and a valid cube with `measures: nope`):

```
load_cubes -> [{'name': 'orders', 'measures': 'nope', 'dimensions': [{'name': 'id'}, 'x'], ...}]
  container is list?        True     # `isinstance(cubes, list)` always holds
  any non-dict cube?        False    # `isinstance(cube, dict)` always holds
  any non-list 'measures'?  True     # member arrays ARE unvalidated
```

So in #2617 the two cube-level guards are dead and only `_member_names` is real. The test reaches the dead branches only by monkeypatching `wren.context.load_cubes` to return `["bad", ...]` — a value the real loader cannot emit.

**The CLI reads `mdl.json` through `json.loads`** (`_load_manifest_dict`), which validates only that the top level is an object. Everything below it is genuinely untrusted, so #2615's cube-level guard is live. But the container is not guarded, and that is a silent-wrong-output bug. Measured on `main` + #2615:

```
cubes: "abc"        -> exit 0, output ''                      # neither listed nor "No cubes defined."
cubes: []           -> exit 0, output 'No cubes defined.'
cubes: ["bad", {…}] -> exit 0, lists orders                   # fixed by this PR
```

A scalar `cubes` prints a blank screen that is indistinguishable from neither-of-the-above. #2617 guards exactly this container case; #2615 does not.

### What to do

Fold the two into one PR that:

- guards the container and the member arrays in **both** entry points, so the CLI and the MCP tool agree
- drops the cube-level guard and its monkeypatched test from the MCP side, where `load_cubes` already guarantees `list[dict]`
- states in the description which side of the `wren-core` deserialization boundary each callsite sits on — the CLI path is upstream (`json.loads`), the MCP path is behind a loader that normalises

Better still, follow 9bdae39c (#2604): if member arrays should be trustworthy, normalise them once in `load_cubes` and report malformed ones from `validate_project`, rather than re-checking at every consumer.

### Rebase

Both branches are 17 commits behind `main` and their CI ran against a base that predates 9bdae39c. They merge cleanly, but a textually clean merge is not a semantically correct one — please rebase and re-run.

Bartok9:
Thanks @goldmedal — same plan as on #2617.

Will fold CLI + MCP cube-list hardening into one PR on current `main`:
- CLI: keep/extend live cube-row guards and add the scalar/`cubes` container guard (blank-output case).
- MCP: keep real member-array hardening only; drop dead cube-level guards that `load_cubes` already makes unreachable.
- Prefer normalising member arrays once in `load_cubes` + `validate_project` reporting (mirroring 9bdae39c) where that fits.

Closing this in favor of that consolidated PR once it’s up; will link both issues/PRs there.

goldmedal:
@Bartok9 Let's close this one and #2617 in favour of the single consolidated cube PR you proposed on Aug 7 — two weeks on, both branches are unchanged and now 27 commits behind, and their CI last ran against a base predating 9bdae39c.

The main reason to restart rather than extend: **#2613 landed on Aug 9 and is the pattern cubes should follow.** You established it there for relationships, matching views in #2604 — loader drops malformed entries, `validate_project` re-reads the raw YAML and reports them so hand-edits don't vanish quietly. Cubes are the last entity without it; `validate_project` on `main` still contains no cube checks at all. Building that once in `load_cubes` + `validate_project` is strictly better than the per-consumer guards on these two branches, and it makes #2617 unnecessary rather than merely incomplete.

Three things from this branch worth carrying into that PR:

**1. Guarding rows without guarding the container trades a crash for silence.** Iterating a `dict` yields its string keys, so they are swallowed by the very guard added here. Measured, `main` vs this branch:

| `manifest["cubes"]` | `main` | this branch |
|---|---|---|
| `"abc"` | exit 1, `AttributeError` | **exit 0, output `''`** |
| `{"name": "orders"}` | exit 1, `AttributeError` | **exit 0, output `''`** |
| `["bad", {"name": "orders", …}]` | exit 1, `AttributeError` | exit 0, lists `orders` ✅ |

Only the last row is the intended fix. For the first two, a blank screen is indistinguishable from `No cubes defined.` never having printed.

**2. `mdl.json` is a build artifact, so the CLI half wants the opposite policy from the project-YAML half.** The default path is `<project>/target/mdl.json` from `wren context build` (`cli.py:157`, and the missing-file hint at `cli.py:31-37`). A malformed cube there means the producer emitted something broken, so skipping the bad rows hides an upstream bug and hands the user a listing that looks complete and isn't. There is no `validate_project` equivalent on that side — the CLI should fail loudly instead:

```
Error: malformed cubes in mdl.json (entry 0 is not an object).
  Hint: re-run `wren context build`.
```

**3. `cube describe` still raises the exact error in this PR's description.** Same `manifest["cubes"]`, same `_load_manifest_dict`, 40 lines below the function fixed here, unguarded at `cube_cli.py:206`. Doing this at `_load_manifest_dict` — the chokepoint every `wren cube` command already routes through — fixes both commands at once, instead of letting `list` and `describe` drift apart as they have here.

Minor, for whenever the tests get rewritten: `test_cube_cli_list_guards.py:3` imports `SimpleNamespace` unused (`ruff` flags it, but the `lint` job runs `ruff check src/` only, so nothing lints `tests/`), the test takes an unused `capsys`, and `_names` is defined inside the loop closing over the loop variable.

Bartok9:
Thanks @goldmedal — closing both as requested.

Will open one consolidated cube PR on current `main` following the #2613 / #2604 pattern:

- **YAML / `load_cubes`:** normalise member arrays in the loader; `validate_project` reports malformed cubes so hand-edits are not silent drops.
- **CLI / `mdl.json`:** treat as build artifact — fail loud at `_load_manifest_dict` (covers `list` + `describe`) with a rebuild hint, not skip-to-blank.
- Carry: container vs row policy split, non-string `name` filter consistency (CLI vs MCP divergence), no dead MCP cube-level guards.

Linking #2615 and #2617 from the replacement when it lands.

Bartok9:
Closing in favor of a single consolidated cube PR (loader + validate_project + loud mdl.json CLI path). See latest comment.

PR #2616, opened 2026-07-29
fix(context): harden description checks for bad MDL rows

Bartok9:
## Summary
Malformed `properties` (non-mapping) on models/views/columns used to surface as a generic AttributeError from description checks (`'str' object has no attribute 'get'`), wrapped by the CLI as `Semantic validation failed`.

This PR reports a clear project-validation error instead: **`properties must be a mapping`**, and keeps `_prop_description` safe if it still sees a non-dict.

```text
# before (main): wren context validate with properties: oops on a model
✗ Semantic validation failed: 'str' object has no attribute 'get'
# exit 1, no pointer that properties shape is wrong

# after:
✗ models/orders/metadata.yml > orders: properties must be a mapping
```

## Duplicate check
- Overlaps conceptually with loader normalisation work (#2604/#2613/#2614) for columns/rows, but this is specifically **properties mapping** at `validate_project` (input boundary), not collection filters in `_check_descriptions`.
- Dropped the earlier dead models/views guards per maintainer feedback.

## Verification
```bash
cd core/wren && uv run pytest tests/unit/test_check_descriptions_guards.py -q
# 2 passed
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved project validation for malformed model, column, and view properties.
  * Validation now reports clear errors when properties are not provided as mappings.
  * Description checks safely handle missing or invalid properties without causing additional failures.

* **Tests**
  * Added coverage for valid, missing, and malformed properties.
  * Added validation tests confirming errors are reported for invalid model, column, and view properties.

### DIFF
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -1644,13 +1644,21 @@ _VALID_LEVELS = frozenset({"error", "warning", "strict"})
 
 
 def _prop_description(item: dict) -> str | None:
-    return (item.get("properties") or {}).get("description")
+    props = item.get("properties") or {}
+    if not isinstance(props, dict):
+        return None
+    return props.get("description")
 
 
 def _check_descriptions(manifest: dict, *, strict: bool = False) -> list[str]:
     warnings: list[str] = []
 
-    for model in manifest.get("models", []):
+    models = manifest.get("models", []) or []
+    if not isinstance(models, list):
+        models = []
+    for model in models:
+        if not isinstance(model, dict):
+            continue
         name = model.get("name", "<unknown>")
         if not _prop_description(model):
             warnings.append(
@@ -1658,14 +1666,24 @@ def _check_descriptions(manifest: dict, *, strict: bool = False) -> list[str]:
                 "add properties.description to improve memory search and agent comprehension"
             )
         if strict:
-            for col in model.get("columns", []):
+            cols = model.get("columns", []) or []
+            if not isinstance(cols, list):
+                cols = []
+            for col in cols:
+                if not isinstance(col, dict):
+                    continue
                 col_name = col.get("name", "<unknown>")
                 if not _prop_description(col):
                     warnings.append(
                         f"Column '{col_name}' in model '{name}' has no description"
                     )
 
-    for view in manifest.get("views", []):
+    views = manifest.get("views", []) or []
+    if not isinstance(views, list):
+        views = []
+    for view in views:
+        if not isinstance(view, dict):
+            continue
         view_name = view.get("name", "<unknown>")
         if not _prop_description(view):
             warnings.append(
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -1117,6 +1117,29 @@ def validate_project(project_path: Path) -> list[ValidationError]:
                     )
                 )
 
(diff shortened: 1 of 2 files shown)


goldmedal:
Thanks for the patch. I reproduced both cases locally against `main` and this branch, and I'd like to see the scope changed before merging.

**Blocking: this makes a malformed `properties` pass validation.**

Project with `properties: oops` on a model, `wren context validate`:

- before: exit 1 — `✗ Semantic validation failed: 'str' object has no attribute 'get'`
- after:  exit 0 — only `⚠ Model 'orders' has no description`

Nothing reports that `properties` is malformed (`validate_project` doesn't check it at all), and the surviving warning is wrong — the user did write `properties`. A validation command should not start accepting input it used to reject. Per `core/wren/.claude/CLAUDE.md`, code at a genuine input boundary should "validate and report rather than crash"; skipping silently is neither.

**Blocking: the `models` / `views` guards are dead code.**

Every loader already filters to dicts — `_load_models_v1`, `_load_models_v2`, `_load_views_v1` (#2604), `_load_views_v2` — and the only production caller is `context_cli.py:490` via `build_json`, so `manifest["models"]` / `["views"]` is always `list[dict]`. `validate_manifest` is not exported in `__all__`. This is the "validate once, at the edge" case the module's CLAUDE.md says is not accepted.

**The description's "crashed" claim doesn't hold.** `context_cli.py:483-494` already wraps this in `try/except Exception`, so the user sees `Semantic validation failed: …` and a non-zero exit, not a traceback. Please paste the actual observed output.

On CodeRabbit's critical comment about the views dry-plan loop: it is right that `view.get("name")` at line 1682 would fail first, but that state is unreachable through the CLI path, so please don't add guards there — that would add more dead code.

Also:

- The test only exercises shapes `build_json` cannot produce. The reachable cases are non-dict `properties` and non-list `columns`; please test those through `validate_manifest` and assert on the warning count so over-reporting can fail the test.
- `... or []` followed by an `isinstance` check is redundant.
- This overlaps #2614, which normalises `columns` in the loader. If that lands, the `columns` guard here is dead. Pick one layer — the loader is consistent with #2604/#2613.
- The branch is 27 commits behind `main` and `context.py` has changed since; please rebase.

Suggested scope: one change — report `properties must be a mapping` from `validate_project` (and let `_prop_description` return `None` safely). Drop the rest.

Bartok9:
Thanks @goldmedal — agreed on the scope.

Rescoped this PR to your suggested shape:

1. **`validate_project`** now errors with `properties must be a mapping` for non-dict `properties` on models, columns, and views (real input boundary).
2. **`_prop_description`** still returns `None` safely on a non-mapping so description checks never AttributeError if they see bad data.
3. **Dropped** the models/views/columns collection guards in `_check_descriptions` (dead after loaders / validate-once-at-edge).
4. **Rebased** onto current `main`.
5. Tests go through `validate_project` (fixture project) and assert three `properties must be a mapping` errors; plus a unit check on `_prop_description`.

On the earlier “crashed” wording: you’re right — the CLI already wraps this as `Semantic validation failed: …` with a non-zero exit, not a raw traceback. PR description updated with the actual observed failure mode.

Happy to trim further if you want properties checks only on models (not columns/views).

goldmedal (approved):
Thanks @Bartok9

PR #2617, opened 2026-07-29
fix(mcp): harden list_cubes member arrays

Bartok9:
MCP `list_cubes` raised `AttributeError: 'str' object has no attribute 'get'` when a cube's `measures`/`dimensions`/`time_dimensions` was non-list or held non-dict members. Repro: call the `list_cubes` tool against a manifest whose cube has `"measures": ["revenue"]` — name extraction called `.get("name")` on the string and the tool failed.

## Summary
`list_cubes` now ignores invalid cube entries and extracts names only from valid dictionary members.

## Verification
```
pytest tests/unit/test_mcp_server.py::test_list_cubes_skips_non_dict_members -q
# 1 passed
```

## Duplicate check
Searched open/closed PRs; #2615 hardens related cube listing but not the member-array normalization in this path. No overlap.

### DIFF
--- a/core/wren/src/wren/mcp_server.py
+++ b/core/wren/src/wren/mcp_server.py
@@ -296,19 +296,28 @@ def _register_context_tools(mcp: FastMCP, ctx: ServeContext) -> None:
         from wren.context import load_cubes  # noqa: PLC0415
 
         cubes = load_cubes(ctx.project)
+        if not isinstance(cubes, list):
+            cubes = []
         result = []
         for cube in cubes:
+            if not isinstance(cube, dict):
+                continue
+
+            def _member_names(key: str) -> list:
+                raw = cube.get(key, []) or []
+                if not isinstance(raw, list):
+                    return []
+                return [
+                    m.get("name") for m in raw if isinstance(m, dict)
+                ]
+
             result.append(
                 {
                     "name": cube.get("name"),
                     "base_object": cube.get("base_object"),
-                    "measures": [m.get("name") for m in cube.get("measures", []) or []],
-                    "dimensions": [
-                        d.get("name") for d in cube.get("dimensions", []) or []
-                    ],
-                    "time_dimensions": [
-                        td.get("name") for td in cube.get("time_dimensions", []) or []
-                    ],
+                    "measures": _member_names("measures"),
+                    "dimensions": _member_names("dimensions"),
+                    "time_dimensions": _member_names("time_dimensions"),
                 }
             )
         return {"cubes": result}
--- a/core/wren/src/wren/mcp_server.py
+++ b/core/wren/src/wren/mcp_server.py
@@ -307,9 +307,7 @@ def _register_context_tools(mcp: FastMCP, ctx: ServeContext) -> None:
                 raw = cube.get(key, []) or []
                 if not isinstance(raw, list):
                     return []
-                return [
-                    m.get("name") for m in raw if isinstance(m, dict)
-                ]
+                return [m.get("name") for m in raw if isinstance(m, dict)]
 
             result.append(
                 {
--- a/core/wren/tests/unit/test_mcp_server.py
+++ b/core/wren/tests/unit/test_mcp_server.py
@@ -405,3 +405,35 @@ def test_query_cube_negative_limit_rejected_consistently():
             limit=-1,
             sql_only=True,
         )
+
+
+def test_list_cubes_skips_non_dict_members(tmp_path, monkeypatch):
+    ctx = _make_ctx(tmp_path)
+    mcp = build_server(ctx)
+    list_cubes = _get_tool(mcp, "list_cubes")
+
+    def fake_load_cubes(project):
+        return [
+            "bad",
+            {
+                "name": "orders",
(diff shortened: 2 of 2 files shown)


goldmedal:
Reviewing this together with #2615 — the MCP half of one change.

#2615 and #2617 harden the same cube listing behind two entry points — the `wren cube list` CLI and the `list_cubes` MCP tool. They should be one PR, because right now each covers a case the other misses and one of them guards states its data source cannot produce.

### The two data sources differ, and that decides which guards are live

**MCP `list_cubes` reads `context.load_cubes`**, which already normalises. Both `_load_cubes_v1` and `_load_cubes_v2` do `if isinstance(data, dict)` before appending, and both always return a list. Measured against hand-edited cube YAML (a file that is a bare scalar, a file that is a list, and a valid cube with `measures: nope`):

```
load_cubes -> [{'name': 'orders', 'measures': 'nope', 'dimensions': [{'name': 'id'}, 'x'], ...}]
  container is list?        True     # `isinstance(cubes, list)` always holds
  any non-dict cube?        False    # `isinstance(cube, dict)` always holds
  any non-list 'measures'?  True     # member arrays ARE unvalidated
```

So in #2617 the two cube-level guards are dead and only `_member_names` is real. The test reaches the dead branches only by monkeypatching `wren.context.load_cubes` to return `["bad", ...]` — a value the real loader cannot emit.

**The CLI reads `mdl.json` through `json.loads`** (`_load_manifest_dict`), which validates only that the top level is an object. Everything below it is genuinely untrusted, so #2615's cube-level guard is live. But the container is not guarded, and that is a silent-wrong-output bug. Measured on `main` + #2615:

```
cubes: "abc"        -> exit 0, output ''                      # neither listed nor "No cubes defined."
cubes: []           -> exit 0, output 'No cubes defined.'
cubes: ["bad", {…}] -> exit 0, lists orders                   # fixed by this PR
```

A scalar `cubes` prints a blank screen that is indistinguishable from neither-of-the-above. #2617 guards exactly this container case; #2615 does not.

### What to do

Fold the two into one PR that:

- guards the container and the member arrays in **both** entry points, so the CLI and the MCP tool agree
- drops the cube-level guard and its monkeypatched test from the MCP side, where `load_cubes` already guarantees `list[dict]`
- states in the description which side of the `wren-core` deserialization boundary each callsite sits on — the CLI path is upstream (`json.loads`), the MCP path is behind a loader that normalises

Better still, follow 9bdae39c (#2604): if member arrays should be trustworthy, normalise them once in `load_cubes` and report malformed ones from `validate_project`, rather than re-checking at every consumer.

### Rebase

Both branches are 17 commits behind `main` and their CI ran against a base that predates 9bdae39c. They merge cleanly, but a textually clean merge is not a semantically correct one — please rebase and re-run.

Bartok9:
Thanks @goldmedal — agreed on the data-source boundary analysis.

Plan:
1. Close this PR and #2615 in favor of one combined follow-up that covers both entry points.
2. Prefer the #2604 / `9bdae39c` pattern: normalise member arrays in `load_cubes` (and guard the CLI `mdl.json` container/`list[dict]` path where `json.loads` is upstream of normalisation), with `validate_project` reporting malformed member arrays rather than re-checking at every consumer.
3. Drop dead cube-level guards + monkeypatched `load_cubes` returns on the MCP side; keep live member-array handling and the CLI container guard.
4. Rebase onto current `main` (post-9bdae39c) and re-run CI before opening the replacement.

Will open the consolidated PR shortly and link both of these from it.

goldmedal:
@Bartok9 Closing request for this one too, alongside #2615 — details of the plan are on that PR.

Short version: with #2613's pattern applied to cubes, **this PR stops being needed at all** rather than needing to be finished. You hardened relationships there on Aug 9 (matching views in #2604) by normalising in the loader and reporting from `validate_project`. Do the same in `load_cubes` and the MCP consumer needs no change — `list_cubes` can go back to reading member arrays directly.

That also settles what the earlier round found here: the two cube-level guards are dead, since `load_cubes` already guarantees `list[dict]` via the `isinstance(data, dict)` gate in both `_load_cubes_v1` and `_load_cubes_v2`. Only `_member_names` does real work, and member arrays are exactly what belongs in the loader.

One divergence worth resolving in the consolidated PR, since it shows the cost of guarding per consumer: this branch's `_member_names` filters non-dict members but not non-string names, while #2615's `_names` filters both. So a member like `{"amount": 1}` is skipped by the CLI and emitted as `[None]` by the MCP tool — same logical fix, same day, two different behaviours.

Both branches are also 27 commits behind now, with CI last run on Jul 29 against a base predating 9bdae39c, so a rebase would be needed either way.

Bartok9:
Thanks @goldmedal — closing alongside #2615.

Agreed: with #2613-style normalisation in `load_cubes` + `validate_project`, the MCP consumer guards here are unnecessary. Replacement PR will unify member-name filtering (dict + string `name`) so CLI/MCP cannot diverge, and keep MCP free of dead cube-level checks.

Bartok9:
Closing in favor of consolidated cube PR with #2615. See latest comment.

PR #2618, opened 2026-07-29
fix(dbt): tolerate non-dict column rows in import helpers

Bartok9:
`_convert_dbt_artifacts` raised `AttributeError: 'str' object has no attribute 'get'` when a dbt `manifest.json` model carried a non-list `columns` value or a list containing non-dict rows (e.g. a bare string). Repro: import a dbt project whose model has `"columns": ["id"]` — the helper called `.get()` on the string and crashed the import.

## Summary
dbt import helpers crashed when `model.columns` was non-list or held non-dict rows. `_find_column`, `_finalize_column_tests`, and `_seed_model_payload` now skip malformed model and column entries.

## Verification
```
pytest tests/unit/test_dbt_column_guards.py -q
# 3 passed
```

## Duplicate check
Searched open/closed PRs and existing dbt guard coverage; #2514 and #2549 harden the seed/query path but not these import helpers (`_find_column`, `_finalize_column_tests`). No overlap with the row-shape guards added here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when processing malformed or unexpected dbt model and column metadata.
  * Invalid entries are now safely skipped instead of interrupting artifact conversion.
  * Seeded column output now includes only valid, named columns.
  * Valid dbt test and primary-key metadata continues to be processed correctly.

* **Tests**
  * Added coverage for malformed model and column data handling.

### DIFF
--- a/core/wren/src/wren/dbt.py
+++ b/core/wren/src/wren/dbt.py
@@ -1059,7 +1059,12 @@ def _infer_attached_node(node: dict[str, Any]) -> str | None:
 
 
 def _find_column(model: dict[str, Any], column_name: str) -> dict[str, Any] | None:
-    for column in model.get("columns", []):
+    columns = model.get("columns", []) or []
+    if not isinstance(columns, list):
+        return None
+    for column in columns:
+        if not isinstance(column, dict):
+            continue
         if column.get("name") == column_name:
             return column
     return None
@@ -1138,8 +1143,15 @@ def _infer_join_type(model: dict[str, Any], target_model: dict[str, Any]) -> str
 
 def _finalize_column_tests(imported_models: list[dict[str, Any]]) -> None:
     for model in imported_models:
+        if not isinstance(model, dict):
+            continue
         primary_key = model.get("primary_key")
-        for column in model.get("columns", []):
+        columns = model.get("columns", []) or []
+        if not isinstance(columns, list):
+            continue
+        for column in columns:
+            if not isinstance(column, dict):
+                continue
             props = column.setdefault("properties", {})
             tests = sorted(set(props.pop("_dbt_tests", [])))
             statuses = props.pop("_dbt_test_statuses", [])
@@ -1279,7 +1291,8 @@ def _seed_model_payload(model: dict[str, Any]) -> dict[str, Any]:
                 "isCalculated": column.get("is_calculated", False),
                 "properties": _camelize_props(column.get("properties", {})),
             }
-            for column in model.get("columns", [])
+            for column in (model.get("columns", []) or [])
+            if isinstance(column, dict) and column.get("name") is not None
         ],
     }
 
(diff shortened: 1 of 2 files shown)


Bartok9:
Thanks @coderabbitai — both points addressed in 0af74ffc: the primary-key loop now skips columns whose `name` is None before any access, and the seed payload comprehension normalizes a non-list `columns` value to an empty list. Existing guard tests still pass.

goldmedal:
All three guards in this PR sit on `imported_models`, which is built by wren's own code — so none of the states they check are reachable. Meanwhile the one place in this file that *does* crash on a malformed dbt manifest is untouched.

### The guarded states cannot occur

`_build_imported_models` appends only what `_build_model_metadata` returns, and that gets its columns from `_extract_columns`, which constructs each column itself:

```python
column = {
    "name": name,
    "type": parse_type(str(data_type), type_dialect),
    "is_calculated": False,
    "not_null": False,
    "properties": _filter_none({"description": description}),
}
columns.append(column)
```

It returns `list[dict]`, and `_build_model_metadata` returns `None` when that list is empty. So for every element of `imported_models`:

- `if not isinstance(model, dict)` — models are dicts built at `_build_model_metadata`
- `if not isinstance(columns, list)` — `_extract_columns` returns a list
- `if not isinstance(column, dict) or column.get("name") is None` — every column is a dict constructed with a `name`

`_find_column`, `_finalize_column_tests` and `_seed_model_payload` are all reached from `imported_models` (lines 780, 827, 1254), so all three added guards are dead code. "An assertion about a condition the code cannot produce is noise" applies to the tests too — they pass literals that the import path cannot produce.

### The reachable bug is one level up

The genuine external boundary is `manifest.json`, read with `json.loads` at line 643 and dereferenced in `_extract_columns`:

```python
manifest_columns = node.get("columns", {}) or {}
merged_names = list(catalog_columns) if catalog_columns else list(manifest_columns)
...
manifest_col = manifest_columns.get(name, {}) or {}
```

A node whose `columns` is not a mapping reaches that `.get`:

```python
>>> _extract_columns({"columns": "id,amount"}, {}, "duckdb")
AttributeError: 'str' object has no attribute 'get'
```

That is a real dbt-manifest input crash, and this PR does not fix it.

### Suggested next step

Please close this and open a replacement that guards `node.get("columns")` in `_extract_columns` (and the matching `catalog_entry.get("columns")`), with the traceback above in the description. That is upstream of everything wren constructs, so one check there removes the need for all three downstream guards.

goldmedal:
Closing this one. All three guards sit on `imported_models`, which wren builds itself — `_extract_columns` constructs every column dict and `_build_model_metadata` drops models with none, so a non-dict model, a non-list `columns`, and a column without a `name` are all unreachable on that path.

The reachable crash is one level up, at the actual `manifest.json` boundary:

```python
>>> _extract_columns({"columns": "id,amount"}, {}, "duckdb")
AttributeError: 'str' object has no attribute 'get'
```

A PR guarding `node.get("columns")` / `catalog_entry.get("columns")` in `_extract_columns`, with that traceback in the description, would be welcome — one check there covers everything downstream. Thanks for looking at this area.

PR #2624, opened 2026-07-30
refactor(connector): centralize LIMIT coercion

Bartok9:
## Summary
Centralize LIMIT handling on a single `coerce_limit()` in `connector/base.py` and wire every interpolating connector through it.

## User-visible change
Invalid `limit` values now raise a consistent `ValueError` instead of a driver-level error after SQL interpolation. This is a **refactor** of the coercion contract across connectors — not an injection fix (call paths that accept `sql` already trust the caller equivalently for `limit`).

## Semantics (strictest of the prior batch)
- `None` → unlimited
- reject `bool`
- reject non-integral values (e.g. `-0.5` must not become `LIMIT 0`)
- reject negatives
- leave `ConnectorABC.query` as `limit: int | None`

## Connectors wired
`postgres`, `trino`, `clickhouse`, `oracle`, `canner`, `bigquery`, `duckdb`, `redshift`, `athena`, `snowflake`, `datafusion`, and `mysql` (private `_coerce_limit` removed).

## Verification
```
cd core/wren && .venv/bin/python -m pytest tests/unit/test_coerce_limit.py -q
8 passed
```

## Notes
Consolidates the closed one-PR-per-connector batch (#2625–#2627, #2634–#2637) per maintainer direction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized query-limit validation across supported data connectors.
  * Rejects negative, fractional, boolean, non-numeric, and oversized values instead of silently converting them.
  * Preserves unlimited queries when no limit is provided.
  * Ensures consistent behavior across Athena, BigQuery, Canner, ClickHouse, DataFusion, DuckDB, MySQL, Oracle, PostgreSQL, Redshift, Snowflake, Trino, and other connectors.

* **Tests**
  * Added coverage for valid limits, absent values, invalid inputs, unsafe strings, and oversized values.

PR #2626, opened 2026-07-30
fix(trino): coerce LIMIT before SQL interpolation

Bartok9:
## Summary
Coerce `limit` to a non-negative int before Trino LIMIT subquery wrapping.

## Motivation
Same injection/negative-limit hole as other connectors that interpolate `{limit}`.

## Verification
```
cd core/wren && .venv/bin/python -m pytest tests/unit/test_trino_coerce_limit.py -q
3 passed
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved query limit handling by accepting valid numeric values and rejecting negative or invalid input.
  - Prevented injection-like content from being included in generated SQL.
  - Added validation to ensure SQL limits are safely formatted before execution.

### DIFF
--- a/core/wren/src/wren/connector/trino.py
+++ b/core/wren/src/wren/connector/trino.py
@@ -456,6 +456,20 @@ def _import_trino():
         ) from e
 
 
+def _coerce_limit(limit: int | None) -> int | None:
+    """Validate and coerce a user-supplied ``limit`` to a non-negative ``int``.
+
+    ``int(limit)`` rejects strings like ``"5 OR 1=1"`` so the value can be
+    safely interpolated into SQL. Negative limits are also rejected.
+    """
+    if limit is None:
+        return None
+    coerced = int(limit)
+    if coerced < 0:
+        raise ValueError(f"limit must be non-negative, got {coerced}")
+    return coerced
+
+
 class TrinoConnector(ConnectorABC):
     """Native trino DB-API connector that bypasses ibis-project."""
 
@@ -482,6 +496,7 @@ class TrinoConnector(ConnectorABC):
         self._closed = False
 
     def query(self, sql: str, limit: int | None = None) -> pa.Table:
+        limit = _coerce_limit(limit)
         trino = _import_trino()
 
         if limit is not None:
--- a/core/wren/tests/unit/test_trino_coerce_limit.py
+++ b/core/wren/tests/unit/test_trino_coerce_limit.py
@@ -1,0 +1,45 @@
+"""Trino query must coerce LIMIT before SQL interpolation."""
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from wren.connector import trino as trino_mod
+
+
+def _connector():
+    c = object.__new__(trino_mod.TrinoConnector)
+    c.connection = MagicMock()
+    c._closed = False
+    return c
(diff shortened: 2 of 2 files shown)


goldmedal:
Closing in favour of a single consolidated change — thanks for the work, the underlying tidy-up is worth doing, just not as one PR per connector.

**Why this is being closed rather than reviewed:**

1. **This is one mechanical change spread across eight PRs.** The contribution bar added in #2602 asks for exactly this to be a single diff: *"A mechanical change repeated across several files or connectors belongs in one PR, not one PR per file. Reviewers need to see the resulting convention in a single diff."*

2. **A shared helper already exists in this same batch, and none of these PRs use it.** #2624 adds `coerce_limit()` to `connector/base.py`. Every other PR in the series re-declares a private `_coerce_limit` in its own module instead of importing it. Merged as-is the repo would carry nine copies of the same function (the eight here plus the existing one at `connector/mysql.py:44`).

3. **The copies have already diverged before merge.** `base.py` (#2624), postgres, trino, oracle, redshift and bigquery use `int(limit)` then a negativity check; #2627 (clickhouse) additionally rejects fractional negatives such as `-0.5` via `numbers.Number`; #2637 (duckdb) additionally rejects `bool` and non-integral `float`. Three different semantics for one contract is the specific outcome a shared helper prevents.

4. **Coverage is inconsistent.** `connector/canner.py:255` interpolates a bare `{limit}` and is not covered by any PR in the series, while `bigquery`, `duckdb` and `redshift` already interpolate `{int(limit)}` today — for those three the only behavioural delta is the negativity check.

**On the `fix:` label and the stated failure.** The reproduction in the description calls `connector.query(sql, limit="1; DROP TABLE users")` directly. Tracing the call paths:

- `run_sql` in `mcp_server.py` declares `limit: int | None`, so a non-numeric string is rejected by tool-argument validation, and `mcp_server.py:84` already rejects negatives and clamps to `MAX_ROW_LIMIT`.
- The CLI declares `--limit/-l` as `Optional[int]`, so a non-integer is rejected at parse time.
- That leaves `Engine.query(sql, limit)` as a Python API. At that boundary the caller already supplies `sql` verbatim — anyone able to pass `limit="1; DROP TABLE t"` can pass that as `sql` instead. `limit` is not a lower-trust channel than `sql` there, so this is not an injection path.

What remains is genuine but smaller: a negative limit currently surfaces as a driver-level error instead of a clear `ValueError`, and the coercion contract is inconsistent across connectors. That is `refactor:`, per *"`fix:` requires a reproducible failure that the change repairs."*

**What we would take instead** — a single PR, `refactor(connector): centralize LIMIT coercion`, that:

- keeps one `coerce_limit()` in `connector/base.py`, with the strictest semantics of the three variants above (reject `bool`, non-integral values, and negatives);
- routes every interpolating connector through it — including `canner.py`, and replacing the private copy in `mysql.py`;
- leaves `ConnectorABC.query`'s signature as `int | None` (see the per-PR note on #2637 below);
- tests the helper once in `tests/unit/test_coerce_limit.py`, with at most a smoke test per connector proving it is wired in, rather than repeating the same six cases eight times.

#2624 is the natural home for that; it is being kept open with a note to that effect.

---

**On this PR specifically** — `trino.py:489` is a bare-`{limit}` site, so it does belong in the consolidated change.

PR #2627, opened 2026-07-30
fix(clickhouse): coerce LIMIT before SQL interpolation

Bartok9:
## Summary
Coerce `limit` to a non-negative int before ClickHouse LIMIT subquery wrapping.

## Failure / reproduction
Before this change, `ClickHouseConnector.query` interpolated the raw `{limit}` directly into the generated SQL:

```python
sql = f"SELECT * FROM ({sql}) LIMIT {limit}"
```

So a client-supplied bad value flowed straight into the query text. E.g. calling `query(sql, limit=-1)` produced:

```
SELECT * FROM (...) LIMIT -1
```

and a non-numeric value like `limit="1; DROP TABLE t"` interpolated verbatim — an injection-shaped input that should never reach the driver.

## Fix
Add `_coerce_limit` and apply it at the start of `query`: reject negatives and non-integer strings, accept numeric strings, so only a validated non-negative int is ever interpolated.

## Verification
```
cd core/wren && .venv/bin/python -m pytest tests/unit/test_clickhouse_coerce_limit.py -q
3 passed
```
Covers: negative rejection, injection-like/non-numeric rejection, numeric-string acceptance.

## Duplicate check
Searched open PRs/issues for existing ClickHouse limit coercion. Related sibling connector PRs (#2624/#2625/#2626) apply the same pattern to other connectors; this PR is the ClickHouse-specific counterpart and does not overlap their files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation of query `LIMIT` inputs by rejecting negative numeric values, including fractional inputs (decimals and fractions), before execution.
  - Prevented negative fractional values from being truncated into non-negative `LIMIT` values.
  - Strengthened limit handling to ensure only valid, non-negative integers are used in generated queries.
- **Tests**
  - Updated and tightened unit coverage for fractional negative `LIMIT` coercion scenarios.

### DIFF
--- a/core/wren/src/wren/connector/clickhouse.py
+++ b/core/wren/src/wren/connector/clickhouse.py
@@ -10,6 +10,7 @@ column-by-column to that schema.
 from __future__ import annotations
 
 import json
+import numbers
 from decimal import Decimal as PyDecimal
 from typing import Any
 from urllib.parse import parse_qsl, unquote, urlparse
@@ -392,8 +393,12 @@ def _coerce_limit(limit: int | None) -> int | None:
     if limit is None:
         return None
     # Reject negatives *before* ``int()`` truncation so a fractional value like
-    # ``-0.5`` cannot slip through as ``LIMIT 0``.
-    if isinstance(limit, (int, float)) and not isinstance(limit, bool) and limit < 0:
+    # ``-0.5`` (float, Decimal, Fraction, ...) cannot slip through as ``LIMIT 0``.
+    if (
+        isinstance(limit, numbers.Number)
+        and not isinstance(limit, bool)
+        and limit < 0
+    ):
         raise ValueError(f"limit must be non-negative, got {limit}")
     coerced = int(limit)
     if coerced < 0:
--- a/core/wren/src/wren/connector/clickhouse.py
+++ b/core/wren/src/wren/connector/clickhouse.py
@@ -391,6 +391,10 @@ def _coerce_limit(limit: int | None) -> int | None:
     """
     if limit is None:
         return None
+    # Reject negatives *before* ``int()`` truncation so a fractional value like
+    # ``-0.5`` cannot slip through as ``LIMIT 0``.
+    if isinstance(limit, (int, float)) and not isinstance(limit, bool) and limit < 0:
+        raise ValueError(f"limit must be non-negative, got {limit}")
     coerced = int(limit)
     if coerced < 0:
         raise ValueError(f"limit must be non-negative, got {coerced}")
--- a/core/wren/src/wren/connector/clickhouse.py
+++ b/core/wren/src/wren/connector/clickhouse.py
@@ -394,11 +394,7 @@ def _coerce_limit(limit: int | None) -> int | None:
         return None
     # Reject negatives *before* ``int()`` truncation so a fractional value like
     # ``-0.5`` (float, Decimal, Fraction, ...) cannot slip through as ``LIMIT 0``.
-    if (
-        isinstance(limit, numbers.Number)
-        and not isinstance(limit, bool)
-        and limit < 0
-    ):
+    if isinstance(limit, numbers.Number) and not isinstance(limit, bool) and limit < 0:
         raise ValueError(f"limit must be non-negative, got {limit}")
     coerced = int(limit)
     if coerced < 0:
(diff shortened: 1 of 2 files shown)


Bartok9:
Good catch, @coderabbitai — `int(-0.5)` truncates to `0` and slipped past the guard. Pushed eebb5ad: `_coerce_limit` now rejects any negative numeric (int/float) *before* `int()` truncation, plus a regression test for `limit=-0.5`.

Bartok9:
Thanks @coderabbitai — good follow-up. Broadened the pre-`int()` negative guard from `(int, float)` to `numbers.Number` in 6c3053a, so `Decimal('-0.5')` and `Fraction(-1, 2)` are now rejected too rather than truncating to `LIMIT 0`. Added regression tests for both.

Bartok9:
Fixed the `lint` failure in a9773cc — hoisted the `Decimal`/`Fraction` imports to module top-level (ruff `PLC0415`); they were function-local in the regression tests. `ruff check` is clean locally now.

goldmedal:
Closing in favour of a single consolidated change — thanks for the work, the underlying tidy-up is worth doing, just not as one PR per connector.

**Why this is being closed rather than reviewed:**

1. **This is one mechanical change spread across eight PRs.** The contribution bar added in #2602 asks for exactly this to be a single diff: *"A mechanical change repeated across several files or connectors belongs in one PR, not one PR per file. Reviewers need to see the resulting convention in a single diff."*

2. **A shared helper already exists in this same batch, and none of these PRs use it.** #2624 adds `coerce_limit()` to `connector/base.py`. Every other PR in the series re-declares a private `_coerce_limit` in its own module instead of importing it. Merged as-is the repo would carry nine copies of the same function (the eight here plus the existing one at `connector/mysql.py:44`).

3. **The copies have already diverged before merge.** `base.py` (#2624), postgres, trino, oracle, redshift and bigquery use `int(limit)` then a negativity check; #2627 (clickhouse) additionally rejects fractional negatives such as `-0.5` via `numbers.Number`; #2637 (duckdb) additionally rejects `bool` and non-integral `float`. Three different semantics for one contract is the specific outcome a shared helper prevents.

4. **Coverage is inconsistent.** `connector/canner.py:255` interpolates a bare `{limit}` and is not covered by any PR in the series, while `bigquery`, `duckdb` and `redshift` already interpolate `{int(limit)}` today — for those three the only behavioural delta is the negativity check.

**On the `fix:` label and the stated failure.** The reproduction in the description calls `connector.query(sql, limit="1; DROP TABLE users")` directly. Tracing the call paths:

- `run_sql` in `mcp_server.py` declares `limit: int | None`, so a non-numeric string is rejected by tool-argument validation, and `mcp_server.py:84` already rejects negatives and clamps to `MAX_ROW_LIMIT`.
- The CLI declares `--limit/-l` as `Optional[int]`, so a non-integer is rejected at parse time.
- That leaves `Engine.query(sql, limit)` as a Python API. At that boundary the caller already supplies `sql` verbatim — anyone able to pass `limit="1; DROP TABLE t"` can pass that as `sql` instead. `limit` is not a lower-trust channel than `sql` there, so this is not an injection path.

What remains is genuine but smaller: a negative limit currently surfaces as a driver-level error instead of a clear `ValueError`, and the coercion contract is inconsistent across connectors. That is `refactor:`, per *"`fix:` requires a reproducible failure that the change repairs."*

**What we would take instead** — a single PR, `refactor(connector): centralize LIMIT coercion`, that:

- keeps one `coerce_limit()` in `connector/base.py`, with the strictest semantics of the three variants above (reject `bool`, non-integral values, and negatives);
- routes every interpolating connector through it — including `canner.py`, and replacing the private copy in `mysql.py`;
- leaves `ConnectorABC.query`'s signature as `int | None` (see the per-PR note on #2637 below);
- tests the helper once in `tests/unit/test_coerce_limit.py`, with at most a smoke test per connector proving it is wired in, rather than repeating the same six cases eight times.

#2624 is the natural home for that; it is being kept open with a note to that effect.

---

**On this PR specifically** — `clickhouse.py:401` is a bare-`{limit}` site. Your `numbers.Number` pre-check is the strictest variant in the series and is the one worth carrying into the shared helper.

Bartok9:
Thanks @goldmedal — agreed on consolidating. Appreciate the detailed rationale (shared helper + one convention, not N private copies).

I'll fold this ClickHouse path into #2624 as the single `refactor(connector): centralize LIMIT coercion` home, carrying the stricter pre-`int()` semantics (`numbers.Number` negatives / fractional rejects) into the shared `coerce_limit()` and wiring the bare-`{limit}` sites (including here and `canner.py`) through it.

Closing as-is is the right call; thanks for the clear bar from #2602.

PR #2628, opened 2026-07-30
fix(wren): bound the session-context cache to 32 LRU entries

ttw225:
## Summary

`get_session_context` used `functools.cache`, and its cache key includes the per-query extracted manifest. As a long-lived process encountered distinct extracted manifests, the cache retained one `SessionContext` for each argument tuple for the lifetime of the process.

This PR replaces it with `lru_cache(maxsize=32)`. The cache now retains at most the 32 most recently used session contexts; an evicted entry is rebuilt when its argument tuple is used again.

The previous cache had no entry-count bound. `WrenEngine._plan()` passes the per-query extracted sub-manifest through:

`extract_by(tables) → effective_manifest → get_session_context(...)`

On the unfixed code, calling `get_session_context` with N distinct `manifest_str` values grows `cache_info().currsize` to N, while `cache_info().maxsize` remains `None`. Each entry retains a `SessionContext` and its analyzed manifest state, so varied workloads can accumulate retained contexts indefinitely.

The value 32 is an explicit initial policy: it establishes a hard bound while keeping recently used planning contexts warm. It is not derived from production memory or cache-hit measurements. Configurability can be added later if operational evidence shows that a different bound is needed.

The runtime-ownership prerequisite shipped in wren-core-py v0.7.2 via #2510: session contexts use a process-wide runtime rather than owning independent runtimes. Evicting a cached context therefore does not tear down a context-specific worker pool.

Cache eviction removes only the cache's reference. An in-flight caller retains its local reference until the operation completes. As with `functools.cache`, `lru_cache` is internally synchronized; concurrent misses may briefly build the same value more than once, but only one entry is retained for that key.

## How is it tested?

`tests/unit/test_session_context_cache.py` uses the decorated production function while replacing the native `SessionContext` constructor with a fake:

- `test_cache_is_bounded_to_32` pins the capacity against the test-side `EXPECTED_MAXSIZE = 32` and verifies that `currsize` stops at the bound.
- `test_cache_evicts_least_recently_used` verifies that access refreshes recency, the surviving key returns the same instance, and the evicted key is rebuilt as a different instance.
- Test keys vary `manifest_str`, matching the production `extract_by → effective_manifest → get_session_context` path.
- The original instance is kept alive during the eviction assertion, so object identity proves removal from the cache rather than object destruction.

The regression tests were also verified against the unfixed `@cache` decorator: the capacity assertion reports `maxsize=None`, and the eviction assertion fails because the supposedly evicted key still returns its original instance.

Verification:

- Targeted cache tests: 2 passed
- Default-CI-equivalent unit suite: 1042 passed, 2 skipped
- Ruff check and format check: passed

## Summary by CodeRabbit

* **Performance**
  * Session context caching is now limited to 32 entries, preventing unbounded memory growth.
  * Least-recently-used entries are automatically removed as new contexts are created.

* **Documentation**
  * Clarified cache eviction behavior and the need to avoid mutating cached session state.

* **Tests**
  * Added coverage for the cache limit and least-recently-used eviction behavior.

### DIFF
--- a/core/wren/.claude/CLAUDE.md
+++ b/core/wren/.claude/CLAUDE.md
@@ -39,7 +39,7 @@ Uses `uv` (not Poetry). `pyproject.toml` uses `hatchling` as build backend.
 - **WrenEngine** is the main entry point. It accepts a base64-encoded MDL JSON string, a `DataSource`, and a connection dict.
 - **Query flow**: `_plan()` → wren-core `SessionContext.transform_sql()` → `_transpile()` via sqlglot → connector `.query()`.
 - **Manifest extraction**: `_plan()` tries to extract a minimal sub-manifest scoped to the query's referenced tables before calling wren-core — this reduces planning overhead. Falls back to the full manifest on error.
-- **`get_session_context` is `@cache`-decorated** — same `(manifest_str, function_path, properties, data_source)` tuple reuses the same SessionContext. Avoid mutating session state.
+- **`get_session_context` is `@lru_cache(maxsize=32)`-decorated** — same `(manifest_str, function_path, properties, data_source)` tuple reuses the same SessionContext; least-recently-used entries are evicted past 32. Never mutate session state: eviction makes the next call for the same tuple rebuild a fresh SessionContext, so any mutation silently disappears at an unpredictable point.
 - **Write dialect mapping**: `canner` → `trino`; file sources (`local_file`, `s3_file`, `minio_file`, `gcs_file`) → `duckdb`. All others use `data_source.name` directly.
 - **WrenEngine is a context manager** (`__enter__` / `__exit__` call `close()`).
 - **Profile-based workflow**: When no explicit `--connection-*` flags are given, the CLI auto-discovers the active profile from `~/.wren/profiles.yml`. Profiles store datasource type + connection fields.
--- a/core/wren/src/wren/mdl/__init__.py
+++ b/core/wren/src/wren/mdl/__init__.py
@@ -1,17 +1,23 @@
 """MDL processing utilities backed by wren-core-py."""
 
-from functools import cache
+from functools import lru_cache
 
 import wren_core
 
 
-@cache
+@lru_cache(maxsize=32)
 def get_session_context(
     manifest_str: str | None,
     function_path: str | None,
     properties: frozenset | None = None,
     data_source: str | None = None,
 ) -> wren_core.SessionContext:
+    """Build (or reuse) a SessionContext for the given manifest tuple.
+
+    Bounded LRU: the cache key includes the per-query extracted manifest
+    (engine.py ``extract_by``), so an unbounded cache would grow one
+    SessionContext per distinct table subset for the life of the process.
+    """
     return wren_core.SessionContext(
         manifest_str, function_path, properties, data_source
     )
--- a/core/wren/tests/unit/test_session_context_cache.py
+++ b/core/wren/tests/unit/test_session_context_cache.py
@@ -20,6 +20,10 @@ class _FakeSessionContext:
 @pytest.fixture()
 def fake_session_context(monkeypatch):
     # Keep this unit test fast and isolated from the native SessionContext.
+    # setattr targets the shared wren_core module object (wren.mdl does
+    # `import wren_core`), so the fake is process-wide until monkeypatch
+    # restores it; clearing the cache on both sides of the yield keeps
+    # fake-backed entries from leaking into other tests.
     monkeypatch.setattr(wren.mdl.wren_core, "SessionContext", _FakeSessionContext)
     get_session_context.cache_clear()
     yield
--- a/core/wren/tests/unit/test_session_context_cache.py
+++ b/core/wren/tests/unit/test_session_context_cache.py
@@ -1,0 +1,59 @@
+"""Tests bounded LRU behavior for get_session_context."""
+
+import pytest
+
+import wren.mdl
+from wren.mdl import get_session_context
+
+pytestmark = pytest.mark.unit
+
+# Asserted literally, never derived from the cache under test: the tests,
+# not the implementation, pin the capacity policy.
+EXPECTED_MAXSIZE = 32
+
+
+class _FakeSessionContext:
+    def __init__(self, *args, **kwargs):
+        self.args = args
+
+
+@pytest.fixture()
+def fake_session_context(monkeypatch):
+    # Keep this unit test fast and isolated from the native SessionContext.
+    monkeypatch.setattr(wren.mdl.wren_core, "SessionContext", _FakeSessionContext)
+    get_session_context.cache_clear()
+    yield
+    get_session_context.cache_clear()
+
+
+def _manifest(i: int) -> str:
+    # Mirrors production keying: engine.py passes the per-query extracted
+    # manifest as manifest_str, so distinct table subsets are distinct keys.
+    return f"manifest-{i}"
+
+
+def test_cache_is_bounded_to_32(fake_session_context):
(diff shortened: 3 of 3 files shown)


goldmedal:
**LGTM — ready to merge.** The fix is correct, minimally scoped, and the tests are genuine regression tests. Two non-blocking suggestions below.

## Verification performed

I didn't just read the diff — I checked out the branch and ran it:

| Check | Result |
| --- | --- |
| Targeted cache tests | 2 passed |
| **Do the tests actually catch the bug?** | Reverted the decorator to `@cache` and re-ran → **both tests fail** (`maxsize=None`; the supposedly evicted key still returns its original instance). The claim in the description is accurate. |
| Full unit suite on the branch | 1058 passed, 2 skipped (`test_memory.py` deselected locally — missing optional `sentence_transformers` dep in my environment, unrelated to this change) |
| `ruff format --check` | Clean on both changed Python files |
| `ruff check` | 40 errors on the branch vs **41 on base `main`** — all pre-existing, and the PR actually removes one (the added module docstring). My local ruff is newer than CI's pin; the CI lint job is green. |

I also confirmed the two load-bearing premises in the description against the source:

1. *The runtime is process-wide, so eviction does not tear down a worker pool* — `core/wren-core-py/src/context.rs:58` holds `static RUNTIME: Mutex<Option<(u32, Arc<Runtime>)>>`, shared across contexts (and rebuilt lazily after a fork). Correct.
2. *The cache key includes the per-query extracted sub-manifest* — `core/wren/src/wren/engine.py:201-220`: `extract_by(tables)` → `to_json_base64` → straight into `get_session_context`. The key space really does grow with the set of table combinations a process sees, so the unbounded retention is real rather than theoretical.

## One risk worth stating explicitly, which I verified is safe

Eviction has exactly one way to break correctness rather than just cost time: `SessionContext` carries mutable state (`register_parquet` / `register_csv`). Under an unbounded cache, registering a table on a cached context is viable because that context never goes away. Once entries can be evicted, the same pattern becomes a silently-failing bug — a later rebuild hands back a clean context and the registration is gone, at an unpredictable point.

I traced every call site: `DataFusionConnector` constructs its own `SessionContext()` directly (`core/wren/src/wren/connector/datafusion.py:27`) and never goes through the cache, so all `register_*` calls land on that non-cached instance. No regression here. The strengthened "never mutate session state" note is a forward-looking guardrail, not a patch over an existing break — worth keeping for exactly that reason.

I also swept the rest of `core/wren` for the same shape: the five `@cache` functions in `connector/mysql.py` are all zero-arg process-wide type maps (no key cardinality at all), and `policy.py:301` already uses `lru_cache(maxsize=128)`. So this was the only unbounded high-cardinality cache in the package.

## Non-blocking suggestions

**1. Consider `maxsize=128` instead of `32`, to match the existing convention in this package.**

This change trades one cost for another rather than being a strict improvement. Under `@cache`, once warm, every distinct table subset was a permanent hit. Under 32, any workload cycling through more than 32 distinct sub-manifests — a multi-panel dashboard, a long agent session touching many table combinations — will rebuild `SessionContext` repeatedly, and rebuilding runs a full MDL analysis via `block_on(apply_wren_on_ctx(...))`. That is not cheap.

Meanwhile each entry holds a *minimised* sub-manifest, so the memory cost of going from 32 to 128 is close to negligible while the thrash risk drops a lot. And `policy.py:301` in the same package already uses `maxsize=128`, so 128 reads as following the existing convention rather than introducing a second magic number. If you'd rather keep 32, making it env-overridable now (e.g. `WREN_SESSION_CACHE_SIZE`) is a one-liner and saves waiting on a release to retune.

**2. (nit) The monkeypatch target in the test reads as more local than it is.**

`wren.mdl` does `import wren_core`, so `monkeypatch.setattr(wren.mdl.wren_core, "SessionContext", ...)` patches the *shared* `wren_core` module object, not a module-local alias — the fake is visible process-wide for the duration of the test. It is safe as written (monkeypatch restores it, and the fixture calls `cache_clear()` on both setup and teardown, which is what stops a fake instance leaking into another test's cache), but a one-line comment noting the real scope would save a future reader from misjudging the isolation.

One detail in the tests I liked: holding the original 32 instances in a list means "key 1 returns a different object" proves *removal from the cache* rather than object destruction. Without that reference the assertion would be much weaker.

goldmedal (approved):

ttw225:
Thanks — your suggestion prompted me to measure the trade-off with the real `wren_core` binding before deciding.

In a local synthetic benchmark on Apple M-series, I generated 128 deterministic 1–4-table extracts from a 24-model manifest using the production key path (`extract_by` → `to_json_base64` → `SessionContext`). Cold construction averaged 2.3 ms (2.2 ms median, 4.9 ms max). Retained RSS averaged approximately 386 KiB per occupied entry, so filling the additional 96 slots from 32 to 128 projected to about 36 MiB of additional RSS in this run.

These are synthetic local measurements rather than production telemetry, but they weaken both assumptions: the high-cardinality extracted manifests were inexpensive to rebuild in absolute terms, while the additional retention was measurable rather than negligible. I also don’t think the `policy.py` value transfers directly: `_canonical_names` caches small frozensets, whereas this cache retains the manifest key and a `PySessionContext` with its analyzed state.

Based on these measurements, I lean toward keeping 32 as the conservative initial bound and deferring configurability until workload evidence shows that retuning is needed. That said, if your experience with production workloads suggests working sets commonly exceed 32 distinct table combinations, I'm happy to adjust to 128 before merge and update the implementation, pinned test, module documentation, and PR description consistently.

I’ve also made the suggested monkeypatch-comment change locally, clarifying that it temporarily affects the shared `wren_core` module object. I’ll include it in the PR update. Thanks again for the thorough verification pass.

goldmedal:
Thanks for measuring rather than deferring to my suggestion — the data settles it, and it settles it against my argument. Keep 32.

I ran the trade-off independently on the branch against the real `wren_core` binding, and got numbers consistent with yours, plus one that I think is the decisive point:

| Build | 24 models | 200 models | 800 models |
| --- | --- | --- | --- |
| `SessionContext(extracted sub-manifest, 3 tables)` | 1.97 ms | 1.74 ms | 1.70 ms |
| `SessionContext(full manifest)` | 2.30 ms | 6.54 ms | 22.80 ms |

The sub-manifest build is flat as the catalog grows. That is the structural reason your choice is right and my thrash argument was wrong: the cache key is the *minimised* extract, so miss cost is bounded by the referenced table subset, not by catalog size. A miss stays ~2 ms in a large deployment as much as a small one — noise next to any database roundtrip. My "rebuilding runs a full MDL analysis, that is not cheap" line does not describe this path.

Your point about `policy.py` is also correct: `_canonical_names` retains small frozensets, so its 128 carries no information about a cache retaining a `PySessionContext`. I withdraw the convention argument along with the perf one. No env override needed either — there is no measured pressure to retune against.

One non-blocking observation for whoever revisits this bound later: not every slot costs the same to refill. The extraction-failure fallback (`engine.py`, `effective_manifest = self.manifest_str`) and the MCP `list_functions` tool both key on the *full* manifest, which is the 22.8 ms column above. In practice those keys stay warm through recency, so it is a corner rather than a concern here.

The monkeypatch-scope comment reads well. Re-verified on `60a70c05`: targeted cache tests 2 passed, CI green. My approval stands — merge at 32.

goldmedal:
Thanks @ttw225 👍

ttw225:
Thanks for running the 200/800-model experiment — scaling the measurement across catalog sizes was the step I missed, and it's what turns "the extracts happened to be cheap" into a structural conclusion.

Honestly, this kind of setting is hard for me to evaluate with confidence. If actual usage ever suggests the bound needs adjusting, I'd be glad to discuss and work on the fix together.

PR #2649, opened 2026-08-06
fix(wren): prevent context upgrade path traversal

### LINKED ISSUE #2630: wren context upgrade writes outside the project directory for traversal-style model names
## Summary

`wren context upgrade` writes files outside the project directory when a legacy
(`schema_version: 1`) project contains a model, view, or cube whose `name` field
resolves upward — e.g. `../../outside/pwned`.

This is the same class of issue as #2580, which added a containment guard to
`write_project_files()`. The upgrade path is a separate code path and did not
receive that guard, so it is still unguarded on `main`.

## Threat model / severity

Local, low severity — consistent with how the sibling issue in #2580 was
classified. It requires the user to run `wren context upgrade` against a
project they did not author. There is no remote trigger. The impact is writing
attacker-chosen filenames (with attacker-chosen YAML content) to attacker-chosen
paths outside the project, which can clobber existing files the user can write.

## Where

`_apply_v1_to_v2()` in `core/wren/src/wren/context.py:1462` builds destination
directories directly from project-supplied names, with no containment check:

- `model_dir = project_path / "models" / name` — `context.py:1475`
- `view_dir  = project_path / "views" / name` — `context.py:1500`
- `cube_dir  = project_path / "cubes" / name` — `context.py:1539`

`name` comes from the project's own YAML via `_load_models_v1()` and friends and
is never validated. After #2580, `write_project_files()` is the only function in
`context.py` that performs a `relative_to()` containment check.

## Reproduction

```python
import tempfile, yaml
from pathlib import Path
from wren.context import plan_upgrade, apply_upgrade

base = Path(tempfile.mkdtemp())
proj = base / "proj"
(proj / "models").mkdir(parents=True)
outside = base / "outside"; outside.mkdir()

(proj / "wren_project.yml").write_text(yaml.dump({"name": "demo", "schema_version": 1}))
(proj / "models" / "orders.yml").write_text(
    yaml.dump({"name": "../../outside/pwned", "columns": [{"name": "id", "type": "int"}]})
)

apply_upgrade(proj, plan_upgrade(proj))
print(sorted(str(p.relative_to(outside)) for p in outside.rglob("*")))
```

Observed on `main`:

```
plan: 1 -> 5
project tree : ['knowledge', ..., 'models', 'wren_project.yml']
OUTSIDE tree : ['pwned', 'pwned/metadata.yml']
ESCAPED: True
```

The file lands outside the project and the command reports success — no error is
raised. The equivalent CLI path is `wren context upgrade` on such a project.

## Suggested fix

Factor the containment check introduced in #2580 into a small helper (resolve
against `project_path.resolve()`, reject anything not strictly inside it, and
reject the project root itself) and apply it to the model/view/cube destinations
in `_apply_v1_to_v2()`. Preflighting all destinations before the first
`mkdir`/`write_text`/`unlink` would also keep a partially-migrated project from
being left on disk, matching the ordering fix in #2580.

Worth deciding separately whether an invalid `name` should abort the upgrade
(consistent with #2580) or be sanitized, since these names also become directory
names in the v2+ layout.

Ray0907:
## Summary

- reject v1 model, view, and cube migration targets that resolve outside their
  collection directories
- validate every migration target during planning and again before apply starts
  mutating the project
- add regression coverage for traversal names in both `plan_upgrade()` and
  `apply_upgrade()`

Fixes #2630.

## Root cause

The v1 to v2 migration joined the YAML `name` directly onto paths such as
`project_path / "models" / name`, then called `mkdir()` and `write_text()`
without resolving the result or checking containment. A name such as
`../../outside/pwned` therefore escaped the project directory.

On `main`, the reproduction wrote outside the project:

```text
planned_path=models/../../outside/pwned/metadata.yml
after_apply_target_exists=True
actual_target_inside_project=False
```

## User impact

Malformed or untrusted legacy projects now fail with `UpgradeError` before any
migration write or delete occurs. Valid v1 projects keep the existing migration
behavior.

After this change, the same reproduction is rejected and preserves the source:

```text
REJECTED: Cannot upgrade: model name '../../outside/pwned' resolves outside models/
NO_OUTSIDE_WRITE: true
SOURCE_PRESERVED: true
```

## Validation

```text
uv run --no-sync pytest tests/unit/test_context.py tests/unit/test_context_cli.py -q
184 passed

uv run --no-sync ruff format --check src/
69 files already formatted

uv run --no-sync ruff check src/
All checks passed!
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved v1-to-v2 upgrade safety by rejecting invalid, non-portable, or traversal-based migration paths.
  * Prevented upgrades from writing outside the project directory or through symlinks.
  * Rejected invalid model SQL and view definitions during migration.
  * Detected duplicate migration targets before making changes.
  * Ensured failed upgrades leave existing files, schemas, and legacy projects unchanged.
  * Upgrades now report clear errors and exit cleanly when migration validation fails.

### DIFF
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -1376,6 +1376,41 @@ class UpgradeError(Exception):
     """Raised when a project upgrade cannot proceed."""
 
 
+def _resolve_upgrade_directory(
+    project_path: Path,
+    collection: str,
+    entity: str,
+    name: Any,
+) -> Path:
+    """Resolve one v1 migration directory without leaving its collection."""
+    if not isinstance(name, str) or not name:
+        raise UpgradeError(
+            f"Cannot upgrade: {entity} name must be a non-empty string, got {name!r}"
+        )
+
+    project_root = project_path.resolve()
+    collection_root = (project_root / collection).resolve()
+    try:
+        collection_root.relative_to(project_root)
+    except ValueError as exc:
+        raise UpgradeError(
+            f"Cannot upgrade: {collection}/ resolves outside the project directory"
+        ) from exc
+
+    target_directory = (collection_root / name).resolve()
+    try:
+        target_directory.relative_to(collection_root)
+    except ValueError as exc:
+        raise UpgradeError(
+            f"Cannot upgrade: {entity} name {name!r} resolves outside {collection}/"
+        ) from exc
+    if target_directory == collection_root:
+        raise UpgradeError(
+            f"Cannot upgrade: {entity} name {name!r} does not identify a directory"
+        )
+    return target_directory
+
+
 def _knowledge_skeleton_targets() -> list[str]:
     """Canonical relative paths of a fresh knowledge/ skeleton.
 
@@ -1462,13 +1497,15 @@ def _plan_v1_to_v2(project_path: Path) -> tuple[list[str], list[str]]:
     """Plan the v1→v2 file restructuring. Returns (files_created, files_deleted)."""
     created: list[str] = []
     deleted: list[str] = []
+    project_root = project_path.resolve()
 
     # Models: flat files → directories
     models = _load_models_v1(project_path)
     for model in models:
         source_dir = model.pop("_source_dir", None)
         name = model.get("name", source_dir or "unknown")
-        dir_path = f"models/{name}"
+        model_dir = _resolve_upgrade_directory(project_path, "models", "model", name)
+        dir_path = model_dir.relative_to(project_root).as_posix()
 
         ref_sql = model.get("ref_sql")
         if ref_sql:
@@ -1485,7 +1522,8 @@ def _plan_v1_to_v2(project_path: Path) -> tuple[list[str], list[str]]:
         name = view.get("name")
         if not name:
             continue
-        dir_path = f"views/{name}"
+        view_dir = _resolve_upgrade_directory(project_path, "views", "view", name)
+        dir_path = view_dir.relative_to(project_root).as_posix()
 
         statement = view.get("statement")
         if statement and "\n" in statement.strip():
@@ -1502,7 +1540,9 @@ def _plan_v1_to_v2(project_path: Path) -> tuple[list[str], list[str]]:
     cubes = _load_cubes_v1(project_path)
     for cube in cubes:
         source_file = cube.pop("_source_file", None)
-        _, target = _cube_migration_target(cube, source_file)
+        name, _ = _cube_migration_target(cube, source_file)
+        cube_dir = _resolve_upgrade_directory(project_path, "cubes", "cube", name)
+        target = (cube_dir / "metadata.yml").relative_to(project_root).as_posix()
         if target in seen_cube_targets:
             raise UpgradeError(
                 f"Cannot upgrade: multiple legacy cube files map to '{target}'"
@@ -1551,12 +1591,15 @@ def apply_upgrade(project_path: Path, result: UpgradeResult) -> None:
 
 def _apply_v1_to_v2(project_path: Path) -> None:
     """Execute the v1→v2 restructuring: write new files, delete old ones."""
+    # Validate every user-derived target before the first filesystem mutation.
+    _plan_v1_to_v2(project_path)
+
     # Write new model directories
     models = _load_models_v1(project_path)
     for model in models:
         source_dir = model.pop("_source_dir", None)
         name = model.get("name", source_dir or "unknown")
-        model_dir = project_path / "models" / name
+        model_dir = _resolve_upgrade_directory(project_path, "models", "model", name)
         model_dir.mkdir(parents=True, exist_ok=True)
 
         ref_sql = model.pop("ref_sql", None)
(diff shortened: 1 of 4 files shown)


goldmedal:
Reviewed at `5feec8cd`. The fix is well-placed and the preflight-before-mutation ordering is the right call — it also picks up the duplicate-cube check before any write, which is a nice side effect. I ran the CI-equivalent gate locally (`pytest tests/unit/ --ignore=tests/unit/test_memory.py --ignore=tests/unit/test_mcp_server.py` → 1082 passed, `ruff check`/`format --check` on `src/` clean) and confirmed the new regression tests fail on `main`.

Two containment gaps still reproduce on this branch.

### 1. A symlinked destination file still escapes the project (same threat model as #2630)

`_resolve_upgrade_directory()` resolves the *directory*, but the writes target `model_dir / "metadata.yml"`, `model_dir / "ref_sql.sql"` and `view_dir / "sql.yml"`. `write_text()` follows a symlink, so a hostile v1 project that ships `models/orders/metadata.yml` as a symlink to a file outside the project overwrites that file — with no traversal in any `name`, and no error raised:

```python
(proj / "models" / "orders").mkdir()
os.symlink(victim, proj / "models" / "orders" / "metadata.yml")  # victim lives outside the project
# models/orders.yml carries a perfectly ordinary  name: orders
apply_upgrade(proj, plan_upgrade(proj))
# -> no error; victim now contains the migrated model YAML
```

This is strictly weaker than the guard added in #2580: `write_project_files()` resolves the **file** target (`(output_dir / file.relative_path).resolve()`), which collapses the symlink before the containment check. Resolving and containing each written file path here too — or rejecting `is_symlink()` targets outright — would close it.

### 2. Path separators inside a name are still accepted, and cause silent data loss

`name: sub/child` stays inside the project, so it passes the check, and the upgrade "succeeds" while destroying the model:

```
planned created:   ['models/sub/child/metadata.yml']
tree after apply:  models/sub/child/metadata.yml, wren_project.yml   # models/orders.yml deleted
_load_models_v2()  -> []                                             # model is gone
```

`_load_models_v2()` only iterates top-level directories containing `metadata.yml`, so the migrated model is unreadable afterwards and the v1 source file has already been unlinked. #2630 explicitly left this open ("these names also become directory names in the v2+ layout"). Rejecting anything that isn't a single path component — e.g. `Path(name).parts != (name,)`, which also catches `.`, `..`, `/` and `\` — would close it and turn the resolve-based check into a second layer rather than the only one.

goldmedal:
Reviewed on a clean worktree of the PR head. The traversal fix itself holds up — I confirmed rejection for `../../outside/pwned`, for an entity directory symlinked out of its collection, for `models/` itself being a symlink out of the project, and for an entity symlink pointing at the collection root. `ruff format --check src/` / `ruff check src/` pass, and `pytest tests/unit/test_context.py tests/unit/test_context_cli.py` is 224 passed locally.

Two things I'd like to see addressed before merge — neither weakens the traversal fix.

### 1. A non-string `ref_sql` still crashes mid-apply and leaves a half-migrated project

This is the same class of bug the third commit fixed for view `statement`, but the model `ref_sql` path was not covered: `_plan_v1_to_v2()` only does `if ref_sql:`, while `_apply_v1_to_v2()` calls `ref_sql.strip()`.

With a project containing a valid `models/aaa.yml` and a `models/zzz.yml` carrying `ref_sql: 42`:

```text
PLAN OK: ['models/aaa/metadata.yml', 'models/zzz/ref_sql.sql', 'models/zzz/metadata.yml']
APPLY RAISED: AttributeError 'int' object has no attribute 'strip'
aaa migrated dir exists: True
aaa.yml source deleted: True      <- source file already removed
zzz.yml source still there: True
schema_version now: 1
```

So on this path the PR's stated invariants — "validate every migration target during planning and again before apply starts mutating the project" and "failed upgrades leave existing files, schemas, and legacy projects unchanged" — do not hold: the first model is migrated and its source deleted, the second is not, and `schema_version` stays at 1.

Suggestion: mirror `_validate_upgrade_view_statement` for `ref_sql` (or generalize both into a single `_validate_upgrade_str_field(entity, name, field, value)`), and call it from both `_plan_v1_to_v2()` and `_apply_v1_to_v2()`.

### 2. The CLI wraps `plan_upgrade` but not `apply_upgrade`

`context_cli.py:925` catches `UpgradeError` from `plan_upgrade()` and prints `Error: ...` with exit 1, but the `apply_upgrade()` call at `context_cli.py:964` is unguarded.

Apply-time re-validation is a central point of this PR — it has dedicated tests for the case where the project changes between plan and apply — yet if that validation ever fires under the CLI, the only shipped consumer, the user gets a Rich traceback instead of the clean error. (That is exactly what the reproduction in point 1 prints today.) Wrapping the apply call in the same `except UpgradeError` block is a one-line fix.

Ray0907:
Thanks for the thorough review. Both points are addressed in `0dcce1c`:

1. v1 model `ref_sql` is now type-validated during planning. Because
   `_apply_v1_to_v2()` reruns the planner before its first filesystem mutation,
   stale-plan apply paths reject non-string values as `UpgradeError` before any
   source file is changed. The regression coverage includes both truthy (`42`)
   and falsy (`0`) non-string values for planning and apply.
2. The CLI now catches `UpgradeError` from `apply_upgrade()`, prints the normal
   `Error: ...` message, and exits with status 1 instead of exposing a traceback.

Validation on the committed head:

- `uv run --no-sync pytest tests/unit/test_context.py tests/unit/test_context_cli.py -q`
  — 229 passed
- `uv run --no-sync ruff format --check src/` — 69 files already formatted
- `uv run --no-sync ruff check src/` — all checks passed

goldmedal (approved):
Thanks @Ray0907 👍

PR #2552, opened 2026-07-20
fix: address CodeRabbit v3 — allow UNION queries, document --api-key

tusharsharma20021114-rgb:
## Summary

Fixes the 2 new CodeRabbit comments from the third review on PR #2551:

1. **`policy.py`** — Change whitelist from `(exp.Select, exp.Explain)` to `(exp.Query, exp.Explain, exp.Subquery)` to allow `UNION`, `INTERSECT`, `EXCEPT`, and parenthesized SELECT queries (BI tools compatibility).

2. **`cli.md`** — Add `--api-key` row to the `wren serve mcp` options table.

### Verification
- `pytest`: 987 passed, 65 failed (pre-existing), 59 skipped
- `cargo check --all-targets --all-features`: clean

## Summary by CodeRabbit

* **New Features**
  * Added optional bearer-token authentication for MCP over HTTP.
  * Added safer, parameterized row-limit handling across database connectors.
  * Added SQL safety checks to reject empty, multi-statement, or non-read-only queries.
* **Bug Fixes**
  * Queries now clearly require an MDL to be loaded before execution.
  * Improved handling of invalid limits and planning errors without unexpected crashes.
* **Changes**
  * Strict mode is now enabled by default, with function names normalized consistently.
* **Documentation**
  * Updated MCP and CLI guidance for HTTP authentication.

PR #2551, opened 2026-07-20
fix: security hardening, config defaults, and test coverage across connectors and Rust core

tusharsharma20021114-rgb:
## Summary

Comprehensive security and stability hardening across the Python connector layer, Rust core (wren-core), config system, and MCP server. All changes add no new dependencies and maintain backward compatibility (deprecation warnings used where defaults changed).

---

## 1. SQL injection prevention — all 15 connectors

Every connector previously interpolated the LIMIT value directly into SQL via f-strings. This is a SQL injection vector when `limit` originates from user input.

**Changes:**
- Added `ConnectorABC._normalize_limit()` — centralized limit sanitization with clamp to `MAX_ROW_LIMIT`, `None` → `MAX_ROW_LIMIT`, negative → `MAX_ROW_LIMIT`
- Added `ConnectorABC._apply_limit_param()` — wraps SQL with parameterized LIMIT (supports `%s`, `?`, `:num` styles)
- Added `ConnectorABC._apply_limit_inline()` — wraps SQL with inline LIMIT (for connectors that don't support parameters in subqueries)
- All 15 connectors use parameterized LIMITs
- DuckDB S3/Minio/GCS secrets: replaced f-string interpolation with parameterized queries, removed `_escape_sql()`

### `strict_mode` default flipped to `True`
- `load_config()` default changed from `False` to `True` — unregistered table references are rejected by default
- Added deprecation warning when `strict_mode` key is absent from config JSON
- Dataclass default (`WrenConfig()`) was already `strict_mode=True` — only `load_config()` was wrong

### `__post_init__` normalization
- `denied_functions` and `allowed_source_functions` lowercased in `__post_init__`
- Removed redundant `.lower()` calls from `load_config()`
- Prevents case-sensitivity bugs

### Denied-function error messages
- Error message now includes the matched denied function name

## 3. MCP server auth — bearer token support

- `--api-key` CLI flag for bearer token auth on Streamable HTTP transport
- `_ApiKeyVerifier` class in `mcp_server.py`
- Connection help text shows `--bearer-token` usage for Claude Code and Codex

## 4. Engine-level SQL safety check

- `basic_safety_check(sql)` added to `dry_plan()`, `query()`, and `dry_run()`

### Configuration
- `unwrapped_used = "deny"` and `expect_used = "deny"` in workspace and standalone crate Cargo.tomls

### Fixes by file
- `mdl/mod.rs` — `// SAFETY:` comment
[...]

PR #2654, opened 2026-08-06
fix(wren): classify bare TimeoutError as DatabaseTimeoutError in WrenEngine

### LINKED ISSUE #2153: `NO_RELEVANT_SQL` Masks Wren Engine Failures
## Summary

When `wren-engine` is unresponsive (e.g., due to CPU deadlock), `wren-ai-service` silently times out and returns `NO_RELEVANT_SQL` — the same error code used for legitimate LLM SQL generation failures. This makes it nearly impossible to distinguish an infrastructure problem from an LLM quality problem without significant manual investigation.

## Impact

Users and operators are misled into debugging LLM configuration (model selection, prompts, parameters) when the actual root cause is a failed or frozen `wren-engine` instance. In our case, this led to multiple unnecessary model switches before the real issue was identified.

## Steps to Reproduce

1. Start the full WrenAI stack via Docker Compose.
2. Allow `wren-engine` to enter a CPU-saturated deadlock state (observed at ~200% CPU, confirmed unresponsive via `curl`).
3. Submit any natural language query from the UI.
4. Observe the result: UI shows "Clarification needed"; `wren-ai-service` logs `NO_RELEVANT_SQL`.

## Debug Timeline

**Symptom**: Every query returns "Clarification needed" in the UI.

**Step 1 — Checked `wren-ai-service` logs**
```
E wren-ai-service] ask pipeline - NO_RELEVANT_SQL: How many orders are there in total?
```
Log only shows the final outcome. No indication of *why* SQL generation failed.

**Step 2 — Mistakenly suspected LLM quality**
Switched models in order: `gpt-4o-mini` → `gpt-4o` → `claude-3-haiku` → back to `gpt-4o-mini`. All produced the same failure. Time spent: significant.

**Step 3 — Checked `wren-ui` logs, found the actual SQL**
```
[DEBUG] WrenEngineAdaptor - Dry run wren engine with body:
"SELECT COUNT(DISTINCT "olist_orders_dataset"."order_id") AS "total_orders" FROM "olist_orders_dataset""
```
The generated SQL was syntactically valid. The dry run was sent to `wren-engine` at `07:50:38` and the task failed at `07:51:09` — exactly 31 seconds later (matching the `engine_timeout: 30` setting).

**Step 4 — Tested `wren-engine` directly**
```bash
docker exec wrenai-wren-ui-1 curl -s --max-time 5 http://wren-engine:8080/v1/mdl
# exit code 28 (timeout) — engine is not responding
```

**Step 5 — Checked container resource usage**
```bash
docker stats wrenai-wren-engine-1 --no-stream
# CPU: 203%, MEM: 674MB — engine is deadlocked
```

**Resolution**: Restarted `wren-engine`. CPU dropped to normal, engine became responsive, queries succeeded.

## Root Cause

`wren-engine` entered a CPU deadlock and stopped responding to HTTP requests. `wren-ai-service` waited until `engine_timeout` (30s) and then classified the result as `NO_RELEVANT_SQL` with no logging of the actual cause.

## Proposed Improvements

1. **Distinguish timeout from SQL failure**: When a dry run times out or returns a connection error, log and surface a distinct error code (e.g., `ENGINE_TIMEOUT` or `ENGINE_UNAVAILABLE`) instead of `NO_RELEVANT_SQL`.

2. **Add explicit timeout logging in `wren-ai-service`**: When the engine call fails due to timeout, log the actual exception/reason, not just the pipeline outcome.

3. **Surface the error to the UI**: Show a different message when the engine is unreachable (e.g., "Unable to reach SQL engine") vs. when no SQL could be generated. This gives operators an actionable signal immediately.

## Environment

- `wren-ai-service` version: 0.29.0
- `wren-engine` version: 0.22.0
- `wren-ui` version: 0.32.2
- LLM: OpenRouter / `gpt-4o-mini`
- Embedder: Gemini / `gemini-embedding-001`

AmirF194:
## Summary

`WrenEngine.query()`/`dry_run()` catch any non-`WrenError` exception from the connector and wrap it as `WrenError(GENERIC_USER_ERROR)`, the same code used for a plain bad SQL query. A connector timeout gets flattened into that same generic bucket, even though four connectors (`postgres`, `athena`, `canner`, `trino`) already special-case `TimeoutError` in their own `except` clauses specifically to keep it from being treated as a generic user error.

`postgres.py`'s `query()`/`dry_run()` both do `except (WrenError, TimeoutError): raise`, so a canceled-statement `TimeoutError` passes through unwrapped. One layer up, `engine.py`'s `query()`/`dry_run()` only special-case `WrenError`; a bare `TimeoutError` isn't one, so it falls into the generic `except Exception` branch and comes out as `GENERIC_USER_ERROR`, indistinguishable from a bad query. `clickhouse.py` doesn't hit this because its own driver-specific `TIMEOUT_EXCEEDED` path raises `DatabaseTimeoutError` (a `WrenError` subclass) directly, so `engine.py`'s `except WrenError: raise` preserves it correctly.

Reproduced with a fake connector that raises a bare `TimeoutError` (mirrors what `postgres.py` re-raises on a canceled statement):

```
wren.model.error.WrenError: [GENERIC_USER_ERROR] canceling statement due to statement timeout phase=SQL_EXECUTION
```

After the fix, the same call raises `DatabaseTimeoutError` with `error_code == ErrorCode.DATABASE_TIMEOUT`.

## How is it tested?

- `tests/unit/test_engine.py::test_query_classifies_bare_timeout_as_database_timeout` and `::test_dry_run_classifies_bare_timeout_as_database_timeout`: both fail against unmodified `engine.py` (assert `DatabaseTimeoutError`, get `WrenError[GENERIC_USER_ERROR]`) and pass on this branch.
- Full `tests/unit/` suite (the `test-unit` CI job's own invocation, `--ignore=tests/unit/test_memory.py --ignore=tests/unit/test_mcp_server.py`): 1078 passed, 2 skipped (pre-existing skips, unrelated to this change).
- `ruff format --check src/` and `ruff check src/` (the `lint` CI job's own commands): clean.
- Not exercised: an actual timed-out query against a live Postgres/Athena/Canner/Trino connection. The four connectors' own `except (WrenError, TimeoutError): raise` clauses are unchanged by this PR; only `engine.py`'s classification of what they re-raise is new.

## Duplicate check

Swept all open PRs touching `core/wren/src/wren/engine.py`: #2258 (adds a YTsaurus path-rewrite method after `dry_run()`), #2551 and #2552 (both add a `basic_safety_check(sql)` call inside `dry_plan()`). Read each diff directly; none touch the `except` blocks in `query()`/`dry_run()` this PR changes.

Fixes #2153

## Summary by CodeRabbit

* **Bug Fixes**
  * Database timeout failures are now consistently reported as database timeout errors during queries and dry runs.
  * Existing Wren errors continue to be preserved, while other unexpected failures retain generic error handling.

### DIFF
--- a/core/wren/src/wren/engine.py
+++ b/core/wren/src/wren/engine.py
@@ -32,7 +32,13 @@ from wren.connector.factory import get_connector
 from wren.mdl import get_manifest_extractor, get_session_context, to_json_base64
 from wren.mdl.cte_rewriter import CTERewriter, get_sqlglot_dialect
 from wren.model.data_source import DataSource
-from wren.model.error import DIALECT_SQL, ErrorCode, ErrorPhase, WrenError
+from wren.model.error import (
+    DIALECT_SQL,
+    DatabaseTimeoutError,
+    ErrorCode,
+    ErrorPhase,
+    WrenError,
+)
 from wren.policy import resolve_model_name, validate_sql_policy
 
 
@@ -117,6 +123,8 @@ class WrenEngine:
             return connector.query(dialect_sql, limit)
         except WrenError:
             raise
+        except TimeoutError as e:
+            raise DatabaseTimeoutError(str(e)) from e
         except Exception as e:
             raise WrenError(
                 ErrorCode.GENERIC_USER_ERROR,
@@ -133,6 +141,8 @@ class WrenEngine:
             connector.dry_run(dialect_sql)
         except WrenError:
             raise
+        except TimeoutError as e:
+            raise DatabaseTimeoutError(str(e)) from e
         except Exception as e:
             raise WrenError(
                 ErrorCode.GENERIC_USER_ERROR,
--- a/core/wren/tests/unit/test_engine.py
+++ b/core/wren/tests/unit/test_engine.py
@@ -14,7 +14,7 @@ import pytest
 from wren import WrenEngine
 from wren.config import WrenConfig
 from wren.model.data_source import DataSource
-from wren.model.error import ErrorCode, WrenError
+from wren.model.error import DatabaseTimeoutError, ErrorCode, WrenError
 
 pytestmark = pytest.mark.unit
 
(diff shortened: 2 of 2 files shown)


goldmedal (approved):
Thanks @AmirF194 👍

PR #2655, opened 2026-08-07
fix(wren): preserve unconstrained PostgreSQL NUMERIC precision

### LINKED ISSUE #2651: wren-cli: PostgreSQL NUMERIC decodes as decimal128(38, 9) and truncates past the 9th decimal
### Environment

`wrenai` 0.8.0 (`uv tool install wrenai==0.8.0 --with trino --with "psycopg[binary]"`),
embedding `wren_core` 0.6.0 · pyarrow 24.0.0 · ibis 12.0.0. `wren --version` →
`wrenai 0.8.0`. Queries run through `wren query -s <sql> -m <mdl.json>
--connection-file <conn.json>`; plans inspected with `wren dry-plan`.
Backend: PostgreSQL 16.14 <or Trino 480>.

> Found by an automated differential sweep I run across semantic query
> services: every query is also re-derived directly against the backend and the
> two results compared, which is where the "raw PostgreSQL" values below come
> from. Happy to re-run any of it against a fix.

### What happens

Every numeric decode logs:

```
wren.connector.postgres:_get_pg_decimal_type:94 - Postgres NUMERIC column has
no scale metadata; defaulting to decimal128(38, 9)
```

For an unconstrained PostgreSQL `NUMERIC` there is no scale metadata to read,
so a default is necessary — but it is applied with no signal to the caller, and
nine decimals is a low ceiling for financial and scientific columns. The values
returned are wrong past the 9th decimal and look complete.

### Repro

```sql
-- <col> is an unconstrained NUMERIC column
SELECT variance(<col>), stddev(<col>) FROM <table>
-- wren query:      103371.211021663      214.335896570
-- raw PostgreSQL:  103371.211021662817   214.335896569810
```

Casting inside the query pushes the cast down and returns full precision on the
same connection:

```sql
SELECT variance(<col>)::text FROM <table>
-- 103371.211021662817   <- full precision
```

So the digits are lost in the client-side decode, not in the backend.

### Why it's worth tracking

The result is silently wrong rather than absent: a caller comparing engine
output against the backend sees agreement to nine decimals and no indication
that anything was capped. This may be the user-visible surface of #2599 rather
than a separate defect.

### Possible directions

1. Read `numeric_scale` from `information_schema` at deploy and use the
   declared scale where one exists (cheapest for the common case; does nothing
   for genuinely unconstrained columns).
2. Widen the fallback, or fall back to string, for unconstrained `NUMERIC`.
3. Surface the existing debug line as a response-level warning, so a caller
   knows precision was capped even when the default is the right choice.

### Related

- #2599 (LocalRuntime schema/type divergence)

Ray0907:
## Summary

- preserve the exact value of PostgreSQL `NUMERIC` columns without a typmod
- keep constrained `NUMERIC(precision, scale)` columns mapped to Arrow decimal types
- apply the same exact-string fallback to unconstrained `NUMERIC[]` elements
- add helper-level and real PostgreSQL regression coverage

Fixes #2651.

## Root cause

`psycopg` reports `scale=None` for an unconstrained PostgreSQL `NUMERIC` column.
The PostgreSQL connector replaced that missing scale with `9`, mapped the column to
`decimal128(38, 9)`, and quantized every value before building the Arrow array.

This produced two observable failures on the current `main` branch:

- `1.123456789012345` was silently returned as `1.123456789`
- `12345678901234567890.123456789012345` raised
  `ArrowInvalid: Rescaling Decimal value would cause data loss`

The connector now returns no Arrow decimal type when the source column has no fixed
scale. Its existing string conversion path then preserves the exact decimal text.
Columns with a declared precision and scale continue to use Arrow decimal values.

## User impact

Queries against unconstrained PostgreSQL `NUMERIC` columns no longer silently round
digits after the ninth decimal or fail while constructing the result table. These
columns are returned as strings because Arrow decimal types require a fixed scale.

## Validation

- confirmed the new regression tests failed against the previous mapping before the
  production change
- `ruff format --check src/`
- `ruff check src/`
- `pytest tests/connectors/test_postgres.py -q` — 24 passed against PostgreSQL 16 in
  Docker
- non-memory unit suite — 1123 passed, 58 skipped, 6 deselected

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PostgreSQL numeric value handling with support for high-precision decimal values.
  * Preserved precision and scale for unconstrained, wide, computed, and array-based numeric results.
  * Added support for decimal128 and decimal256 values based on numeric precision.
  * Added safe string fallback for non-finite or otherwise unrepresentable numeric values.
  * Continued support for constrained numeric columns using appropriately bounded decimal types.
  * Improved consistency when numeric metadata is incomplete, helping preserve accurate query results.

### DIFF
--- a/core/wren/src/wren/connector/postgres.py
+++ b/core/wren/src/wren/connector/postgres.py
@@ -127,6 +127,13 @@ def _infer_pg_decimal_type(values: list) -> pa.DataType | None:
 
 def _get_pg_decimal_type(column, values: list | None = None) -> pa.DataType | None:
     """Use numeric typmod when representable, otherwise infer from values."""
+    if values is not None and any(
+        value is not None
+        and (not isinstance(value, PyDecimal) or not value.is_finite())
+        for value in values
+    ):
+        return None
+
     precision = column.precision
     scale = column.scale
     if precision is not None and scale is not None:
--- a/core/wren/src/wren/connector/postgres.py
+++ b/core/wren/src/wren/connector/postgres.py
@@ -111,6 +111,7 @@ def _infer_pg_decimal_type(values: list) -> pa.DataType | None:
             return None
         saw_decimal = True
         if value.is_zero():
+            max_scale = max(max_scale, max(-value.as_tuple().exponent, 0))
             continue
 
         digits = len(value.as_tuple().digits)
--- a/core/wren/src/wren/connector/postgres.py
+++ b/core/wren/src/wren/connector/postgres.py
@@ -87,31 +87,66 @@ _PG_OID_TO_ARROW: dict[int, pa.DataType] = {
 }
 
 
-def _get_pg_decimal_type(column) -> pa.DataType | None:
-    """Map a constrained numeric column to the narrowest Arrow decimal type.
-
-    Unconstrained NUMERIC has no fixed scale, so return ``None`` rather than
-    rounding values to an arbitrary Arrow decimal scale.
-    """
-    if column.scale is None:
+def _make_pg_decimal_type(precision: int, scale: int) -> pa.DataType | None:
+    """Return the Arrow decimal type that can represent *precision*."""
+    if precision <= 0:
         return None
-    scale = max(0, min(column.scale, 38))
-
-    precision = column.precision if column.precision is not None else 38
-    if precision <= 0 or precision > 38:
-        precision = 38
-    precision = max(precision, scale, 1)
-    precision = min(precision, 38)
-
-    return pa.decimal128(precision, scale)
+    if precision <= 38:
+        return pa.decimal128(precision, scale)
+    if precision <= 76:
+        return pa.decimal256(precision, scale)
+    return None
 
 
-def _get_pg_arrow_type(column) -> pa.DataType:
+def _infer_pg_decimal_type(values: list) -> pa.DataType | None:
+    """Infer one exact Arrow decimal type for returned psycopg values."""
+    max_integer_digits = 0
+    max_scale = 0
+    saw_decimal = False
+
+    for value in values:
+        if value is None:
+            continue
+        if not isinstance(value, PyDecimal) or not value.is_finite():
+            return None
+        saw_decimal = True
+        if value.is_zero():
+            continue
+
+        digits = len(value.as_tuple().digits)
+        exponent = value.as_tuple().exponent
+        scale = max(-exponent, 0)
+        integer_digits = digits + exponent if exponent >= 0 else max(digits - scale, 0)
+        max_integer_digits = max(max_integer_digits, integer_digits)
+        max_scale = max(max_scale, scale)
+
+    if not saw_decimal:
+        return pa.decimal128(1, 0)
+    return _make_pg_decimal_type(max(max_integer_digits + max_scale, 1), max_scale)
+
+
+def _get_pg_decimal_type(column, values: list | None = None) -> pa.DataType | None:
+    """Use numeric typmod when representable, otherwise infer from values."""
+    precision = column.precision
+    scale = column.scale
+    if precision is not None and scale is not None:
+        arrow_type = _make_pg_decimal_type(precision, scale)
+        if arrow_type is not None:
+            return arrow_type
+
+    return _infer_pg_decimal_type(values) if values is not None else None
+
+
+def _get_pg_arrow_type(column, values: list | None = None) -> pa.DataType:
     """Map a psycopg cursor description column to an Arrow type."""
     if column.type_code == 1700:
-        return _get_pg_decimal_type(column) or pa.string()
+        decimal_type = _get_pg_decimal_type(column, values)
+        return decimal_type if decimal_type is not None else pa.string()
     if column.type_code == 1231:
-        inner_type = _get_pg_decimal_type(column)
+        items = None
+        if values is not None:
+            items = [item for value in values if value is not None for item in value]
+        inner_type = _get_pg_decimal_type(column, items)
         return pa.list_(inner_type) if inner_type is not None else pa.list_(pa.string())
     return _PG_OID_TO_ARROW.get(column.type_code, pa.string())
 
@@ -122,9 +157,16 @@ def _build_pg_arrow_table(cursor) -> pa.Table:
         return pa.table({})
 
     rows = cursor.fetchall()
+    column_values = [
+        [row[index] for row in rows] for index in range(len(cursor.description))
+    ]
     fields = [
-        pa.field(column.name, _get_pg_arrow_type(column), nullable=True)
-        for column in cursor.description
+        pa.field(
+            column.name,
+            _get_pg_arrow_type(column, column_values[index]),
+            nullable=True,
+        )
+        for index, column in enumerate(cursor.description)
     ]
     schema = pa.schema(fields)
 
@@ -133,7 +175,7 @@ def _build_pg_arrow_table(cursor) -> pa.Table:
     else:
         arrays = [
             _build_pg_column(
-                [row[index] for row in rows],
+                column_values[index],
                 field.type,
                 cursor.description[index].type_code,
             )
--- a/core/wren/src/wren/connector/postgres.py
+++ b/core/wren/src/wren/connector/postgres.py
@@ -121,6 +121,7 @@ def _infer_pg_decimal_type(values: list) -> pa.DataType | None:
         max_scale = max(max_scale, scale)
 
     if not saw_decimal:
+        # No returned value exposes precision or scale; use the smallest valid type.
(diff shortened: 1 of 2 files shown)


goldmedal (changes requested):
Good diagnosis of #2651, and the constrained path is correctly preserved. One blocking issue plus a few small things.

---

## Blocking — the string fallback fires on far more than unconstrained columns, and nothing pins the new contract

PostgreSQL reports no typmod for *any* computed numeric expression, so `column.scale is None` is not limited to columns declared as bare `NUMERIC`. Verified on PG 16 with `amount NUMERIC(10,2)`:

| expression | numeric_precision | numeric_scale |
|---|---|---|
| `amount` | 10 | 2 |
| `sum(amount)` | NULL | NULL |
| `avg(amount)` | NULL | NULL |
| `amount * 2` | NULL | NULL |
| `variance(raw)` | NULL | NULL |

So after this change every aggregate and arithmetic result over `numeric` comes back as an Arrow **string**, even when the source column is fully constrained. `SELECT SUM(revenue) FROM orders` is now a string column. (`mysql.py` still returns `decimal128` for the same logical query.)

Two ways forward:

**A — derive the scale from the returned values (preferred).** `_build_pg_arrow_table` already calls `cursor.fetchall()` *before* it builds the schema fields, so the values are in hand at type-selection time. For a `scale is None` column, scan the non-null `Decimal`s, take the max required scale and precision, pick `decimal128` (or `decimal256` when it doesn't fit), and fall back to string only when it genuinely can't be represented. That gets exactness — the point of #2651 — while keeping the column numeric for essentially every real query. The issue lists "fall back to string" as one of three options, not the recommended one.

**B — keep the string fallback,** but then it needs:

- the PR description rewritten to state that *all* numeric aggregate and arithmetic results become strings, not just unconstrained columns;
- a test that pins the new contract, e.g. asserting `SELECT SUM(c_numeric) FROM type_zoo` has type `pa.string()`. Right now the tests only prove values are exact — nothing fails if someone reverts this to a decimal type;
- a retained log signal. The PR deletes the one `logger.debug` line, replacing a silent truncation with a silent type change.

---

## Non-blocking, actionable

1. **The array fallback has no value-level test.** "apply the same exact-string fallback to unconstrained `NUMERIC[]` elements" is covered only by the type assertion `_get_pg_arrow_type(_column(1231)) == pa.list_(pa.string())`. Nothing exercises the list-of-string branch of `_build_pg_column` with `Decimal` items, and the live test can't cover it — `type_zoo.c_numeric_arr` is `NUMERIC(38,9)[]`, i.e. constrained. Add:

   ```python
   _build_pg_column([[Decimal("1.123456789012345")]], pa.list_(pa.string()), 1231)
   ```

2. **`NUMERIC(p,s)` with `p > 38` still overflows.** Precision is clamped to 38, so `ArrowInvalid: Rescaling Decimal value would cause data loss` is still reachable for constrained columns. Pre-existing, but it's the same failure mode this PR is removing — either handle it (`decimal256`) or say explicitly it's out of scope.

3. `_get_pg_decimal_type(column) or pa.string()` relies on `pa.DataType` truthiness while the next line uses explicit `is not None`. (I checked — no `pa.DataType` is falsy, so not a bug.) Make both `is not None`.

4. `test_unconstrained_numeric_preserves_exact_value` exists twice, module-level and as a class method. Legal, but confusing in `-v` output — rename one.

5. `_column` can be `types.SimpleNamespace(type_code=..., precision=..., scale=...)` instead of a locally-defined empty class.

**No action needed on these** — flagging so they don't get "fixed": the new module-level tests *are* selected in CI (the `postgres` matrix job runs `-m postgres`, and module-level `pytestmark` applies to them), and `logger` is still used at `postgres.py:303`, so removing the debug line leaves no dead import.

Ray0907:
Thanks for the detailed review. I’ve addressed the blocking issue using option A:

- infer a common precision and scale from returned `Decimal` values;
- use `decimal128` or `decimal256` when representable;
- fall back to exact strings only when values exceed Arrow decimal capacity or are non-finite;
- add live coverage for `SUM`, `AVG`, arithmetic expressions, and unconstrained `NUMERIC[]`;
- handle constrained `NUMERIC(p, s)` where `p > 38`;
- add scalar and array regressions for `NaN` and infinity;
- address the remaining test and style suggestions.

Decimal inference for representable unconstrained `NUMERIC[]` values is intentional and follows option A. Mixed-scale inference and exact-string fallback beyond Arrow’s decimal capacity are both covered by tests.

The branch is rebased onto the latest `main`, and all current CI checks pass.

@goldmedal Would you mind taking another look? Thanks!

goldmedal (changes requested):
Thanks — this is a much better fix than the original approach. Everything from the last round is addressed:

- Aggregates and arithmetic stay numeric now (`test_computed_numeric_expressions_remain_decimal` pins `SUM` / `AVG` / `* 2`), which was the blocking concern.
- `decimal256` for `p > 38`, with fall-through to value inference above 76 — that also closes the pre-existing `NUMERIC(50,10)` overflow I flagged.
- Array path has real value-level coverage now, both inferred and string-fallback.
- Non-finite handling is a genuine improvement over `main`, which crashed on `'NaN'::numeric`.
- Explicit `is not None`, `SimpleNamespace`, duplicate test name — all fixed.

One blocking issue in the new code.

---

## Blocking — the typmod path lost its clamping, and now crashes on typmods `main` handles

The old `_get_pg_decimal_type` clamped before building the type:

```python
scale = max(0, min(scale, 38))
precision = max(precision, scale, 1)
```

`_make_pg_decimal_type(precision, scale)` passes both straight through. Since PostgreSQL 15, `scale > precision` and negative scale are legal typmods, and psycopg surfaces them verbatim.

Repro:

```sql
CREATE TABLE odd (a numeric(3,5));
INSERT INTO odd VALUES (0.00123);
SELECT a FROM odd;
```

`pg_attribute.atttypmod` is `196617` there, so psycopg reports `precision=3, scale=5`. Running this branch's `_get_pg_arrow_type` + `_build_pg_column` on that column:

| column | `main` | this branch |
|---|---|---|
| `numeric(3,5)` | `decimal128(5, 5)` → `0.00123` ✅ | **`ArrowInvalid: Decimal type with precision 5 does not fit into precision inferred from first array element: 3`** |
| `numeric(3,-2)` | `ArrowInvalid` | `ArrowInvalid` (unchanged) |

Note `pa.decimal128(3, 5)` *constructs* fine — it only blows up at array-build time, which is why nothing catches it earlier.

The negative-scale row is pre-existing, not a regression. But `scale > precision` is a straight regression: it works on `main` and fails here, and it's the exact failure mode this PR set out to remove.

The fix is cheap and fits the new design — treat an unrepresentable typmod the same way you already treat `precision > 76`, i.e. don't trust it, fall through to `_infer_pg_decimal_type(values)`, which computes the right type from the data anyway:

```python
if precision is not None and scale is not None and 0 <= scale <= precision:
    arrow_type = _make_pg_decimal_type(precision, scale)
    if arrow_type is not None:
        return arrow_type
```

That fixes both rows of the table at once. Please add a live case for it — `numeric(3,5)` and `numeric(3,-2)` columns in `type_zoo` would do it, and the container is already `postgres:16` so both typmods are accepted.

---

## Minor

1. **Empty and all-NULL unconstrained numeric columns get `decimal128(1, 0)`.** `_infer_pg_decimal_type` returns that when `saw_decimal` is False. No data loss, but it means a query returning zero rows reports a different schema than the same query returning rows — worth a short comment, or reusing the typmod/string default for the no-values case.

2. **The schema is now value-dependent by design** — the same query over different data can return different precision/scale, and a constrained column silently degrades to `string` when a `NaN` shows up. That's the right trade-off for exactness, but it's surprising enough to belong in the `_get_pg_arrow_type` docstring rather than only in the PR description.

3. `Decimal("0.00")` alone in a column infers `decimal128(1, 0)` and renders as `0` — zeros are skipped by the `value.is_zero()` branch, so declared scale is dropped. Numerically exact, cosmetic only; flagging in case the display scale matters to you.

**Verification note:** I exercised this branch's `_get_pg_arrow_type` / `_build_pg_column` directly on the values above, and read the typmod encoding out of `pg_attribute` and decoded it with psycopg3's formula. I did not run psycopg itself in this environment, so please confirm the crash with a live test before fixing.

Ray0907:
Follow-up pushed in `9018b7eb`: unconstrained zero Decimal values now contribute their scale during inference, so `Decimal("0.00")` stays `decimal128(2, 2)` instead of collapsing to `decimal128(1, 0)`. I added a regression test; the PostgreSQL suite passes with 35 tests, and Ruff checks pass. @goldmedal, would you mind taking another look? Thanks!

goldmedal (approved):
Thanks @Ray0907, look nice 👍

PR #2657, opened 2026-08-10
fix(wren): preserve wide MySQL and Doris decimals

Ray0907:
## Summary

- Map MySQL and Doris DECIMAL metadata to `decimal128` through precision 38 and
  `decimal256` through precision 76.
- Inspect fetched Decimal values and widen the Arrow type when expression metadata
  understates the actual precision.
- Return exact strings when a concrete value exceeds Arrow Decimal256's 76-digit limit,
  including a safe schema fallback for empty or all-null over-wide results.
- Run the native MySQL connector regression suite in Wren CI.

Wide MySQL DECIMAL values were always assigned `decimal128`, so a value such as
`DECIMAL(65,30)` failed while constructing the Arrow table.

MySQL expression metadata can also disagree with the returned value. On MySQL 8.0.36:

```sql
SELECT CAST('<65-digit integer>' AS DECIMAL(65,0))
     * CAST('<12-digit integer>' AS DECIMAL(12,0)) AS x;
```

The driver reported precision 66 (which derived an Arrow precision of 65), but returned
a 77-digit `Decimal`. The connector then raised:

```text
ArrowInvalid: Decimal type with precision 77 does not fit into precision inferred from first array element: 65
```

The same failure is reproducible with `DECIMAL(40,0) * DECIMAL(40,0)`, which returns an
80-digit value. This change derives the final column type after fetching values: values
that fit within 76 digits remain exact Arrow decimals, while wider values are emitted as
exact strings instead of raising or rounding.

## How is it tested?

- `tests/unit/test_mysql_helpers.py` covers decimal128/decimal256 boundaries, unsigned
  metadata, value-aware integer and scale widening, over-wide string fallback, and
  empty/all-null schemas.
- `tests/connectors/test_mysql_connector.py` runs against MySQL 8.0.36 and covers
  `DECIMAL(65,30)`, 66-digit addition, 77/80-digit multiplication, and wide `SUM`
  metadata.
- The Wren CI workflow now collects both MySQL connector suites.

Local verification:

- 62 targeted value-aware tests passed.
- 1,096 non-memory unit tests passed, with 1 skipped.
- 37 tests across both MySQL connector suites passed.
- Ruff format/lint and `git diff --check` passed.

## Duplicate check

Searched open PRs in `Canner/WrenAI` for `MySQL DECIMAL`, `Decimal256`, and `decimal`
in titles; no duplicates were found.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MySQL handling of high-precision decimal values, including wider numeric ranges.
  * Preserved numeric results when values fit supported precision and scale limits.
  * Returned oversized, invalid, or non-finite decimal results as exact strings without losing precision.
  * Improved precision handling for decimal calculations, including addition and multiplication.

* **Tests**
  * Added coverage for wide decimals, boundary conditions, null values, empty results, and exact string fallbacks.

### DIFF
--- a/core/wren/src/wren/connector/mysql.py
+++ b/core/wren/src/wren/connector/mysql.py
@@ -369,7 +369,10 @@ def _mysql_decimal_type_for_values(
     target_integer_digits = max(precision - scale, integer_digits)
     target_precision = target_integer_digits + target_scale
     if target_precision > _ARROW_DECIMAL256_MAX_PRECISION:
-        return pa.string()
+        target_scale = _ARROW_DECIMAL256_MAX_PRECISION - target_integer_digits
+        if target_scale < value_scale:
+            return pa.string()
+        return pa.decimal256(_ARROW_DECIMAL256_MAX_PRECISION, target_scale)
     return _arrow_decimal_type(target_precision, target_scale)
 
 
--- a/core/wren/src/wren/connector/mysql.py
+++ b/core/wren/src/wren/connector/mysql.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 import json
 from contextlib import closing
 from decimal import Decimal as PyDecimal
+from decimal import InvalidOperation
 from functools import cache
 
 import pyarrow as pa
@@ -321,7 +322,10 @@ def _arrow_decimal_type(precision: int, scale: int) -> pa.DataType:
 
 def _mysql_decimal_value_shape(value) -> tuple[int, int] | None:
     """Return integer digits and scale needed to preserve one decimal value."""
-    value = value if isinstance(value, PyDecimal) else PyDecimal(str(value))
+    try:
+        value = value if isinstance(value, PyDecimal) else PyDecimal(str(value))
+    except (InvalidOperation, TypeError, ValueError):
+        return None
     if not value.is_finite():
         return None
     _, digits, exponent = value.as_tuple()
@@ -345,8 +349,8 @@ def _mysql_decimal_type_for_values(
         _mysql_decimal_value_shape(value) for value in values if value is not None
     ]
     if not shapes:
-        if precision > _ARROW_DECIMAL256_MAX_PRECISION:
-            return pa.string()
+        precision = min(precision, _ARROW_DECIMAL256_MAX_PRECISION)
+        scale = min(scale, precision)
         return _arrow_decimal_type(precision, scale)
     if any(shape is None for shape in shapes):
         return pa.string()
@@ -382,6 +386,7 @@ def _mysql_field_arrow_type(
     flags: int = 0,
     precision: int | None = None,
     scale: int | None = None,
+    values: list | None = None,
 ) -> pa.DataType:
     from MySQLdb.constants import FLAG  # noqa: PLC0415
 
@@ -401,6 +406,13 @@ def _mysql_field_arrow_type(
     # — the previous hard-coded ``decimal128(38, 9)`` would lose digits when
     # ``D > 9``.
     if type_code in decimal_codes:
+        if values is not None:
+            return _mysql_decimal_type_for_values(
+                precision,
+                scale,
+                is_unsigned=bool(flags & FLAG.UNSIGNED),
+                values=values,
+            )
         return _arrow_decimal_from_mysql_field(
             precision, scale, is_unsigned=bool(flags & FLAG.UNSIGNED)
         )
@@ -444,19 +456,13 @@ def _build_mysql_arrow_table(cursor) -> pa.Table:
         precision = col[4] if len(col) > 4 else None
         scale = col[5] if len(col) > 5 else None
         flags = flag_list[i] or 0
-        if col[1] in _mysql_decimal_codes():
-            from MySQLdb.constants import FLAG  # noqa: PLC0415
-
-            arrow_type = _mysql_decimal_type_for_values(
-                precision,
-                scale,
-                is_unsigned=bool(flags & FLAG.UNSIGNED),
-                values=[row[i] for row in rows],
-            )
-        else:
-            arrow_type = _mysql_field_arrow_type(
-                col[1], flags, precision=precision, scale=scale
-            )
+        arrow_type = _mysql_field_arrow_type(
+            col[1],
+            flags,
+            precision=precision,
+            scale=scale,
+            values=[row[i] for row in rows],
+        )
         fields.append(pa.field(col[0], arrow_type, nullable=True))
     schema = pa.schema(fields)
 
--- a/core/wren/src/wren/connector/mysql.py
+++ b/core/wren/src/wren/connector/mysql.py
@@ -275,7 +275,7 @@ def _arrow_decimal_from_mysql_field(
     when the column is signed. The declared ``DECIMAL(M, D)`` precision ``M``
     is recovered as::
 
-        M = length - (1 if unsigned else 0) - (1 if D > 0 else 0)
+        M = length - (0 if unsigned else 1) - (1 if D > 0 else 0)
 
     MySQL allows precision up to 65 and scale up to 30. Doris uses the same
     protocol and supports precision and scale up to 76 when Decimal256 is
@@ -283,6 +283,20 @@ def _arrow_decimal_from_mysql_field(
     ``decimal256`` covers both data sources' wider decimals. The previous
     hard-coded ``decimal128(38, 9)`` would silently lose digits when ``D > 9``.
     """
+    precision, scale = _mysql_decimal_precision_scale(
+        display_length, scale, is_unsigned
+    )
+    precision = min(precision, _ARROW_DECIMAL256_MAX_PRECISION)
+    scale = min(scale, precision)
+    return _arrow_decimal_type(precision, scale)
+
+
+def _mysql_decimal_precision_scale(
+    display_length: int | None,
+    scale: int | None,
+    is_unsigned: bool,
+) -> tuple[int, int]:
+    """Return raw MySQL decimal precision and scale without Arrow clamping."""
     if display_length is None or display_length <= 0:
         precision = _MYSQL_DECIMAL_FALLBACK_PRECISION
     else:
@@ -294,11 +308,69 @@ def _arrow_decimal_from_mysql_field(
             precision = _MYSQL_DECIMAL_FALLBACK_PRECISION
     if scale is None or scale < 0:
         scale = _MYSQL_DECIMAL_FALLBACK_SCALE
-    precision = min(int(precision), _ARROW_DECIMAL256_MAX_PRECISION)
+    precision = int(precision)
     scale = min(int(scale), precision)
+    return precision, scale
+
+
+def _arrow_decimal_type(precision: int, scale: int) -> pa.DataType:
     if precision <= _ARROW_DECIMAL128_MAX_PRECISION:
         return pa.decimal128(precision, scale)
     return pa.decimal256(precision, scale)
+
+
+def _mysql_decimal_value_shape(value) -> tuple[int, int] | None:
+    """Return integer digits and scale needed to preserve one decimal value."""
+    value = value if isinstance(value, PyDecimal) else PyDecimal(str(value))
+    if not value.is_finite():
+        return None
+    _, digits, exponent = value.as_tuple()
+    if exponent >= 0:
+        return len(digits) + exponent, 0
+    value_scale = -exponent
+    return max(len(digits) - value_scale, 0), value_scale
+
+
+def _mysql_decimal_type_for_values(
+    display_length: int | None,
+    scale: int | None,
+    is_unsigned: bool,
+    values: list,
+) -> pa.DataType:
+    """Choose an exact Arrow type from decimal metadata and concrete values."""
+    precision, scale = _mysql_decimal_precision_scale(
+        display_length, scale, is_unsigned
+    )
+    shapes = [
+        _mysql_decimal_value_shape(value) for value in values if value is not None
+    ]
+    if not shapes:
+        if precision > _ARROW_DECIMAL256_MAX_PRECISION:
+            return pa.string()
+        return _arrow_decimal_type(precision, scale)
+    if any(shape is None for shape in shapes):
+        return pa.string()
+
+    integer_digits = max(shape[0] for shape in shapes if shape is not None)
+    value_scale = max(shape[1] for shape in shapes if shape is not None)
+    if integer_digits + value_scale > _ARROW_DECIMAL256_MAX_PRECISION:
+        return pa.string()
+
+    if precision > _ARROW_DECIMAL256_MAX_PRECISION:
+        # Preserve as much metadata scale as possible while reserving enough
+        # integer capacity for every fetched value.
+        target_scale = max(
+            value_scale,
+            min(scale, _ARROW_DECIMAL256_MAX_PRECISION - integer_digits),
+        )
+        return pa.decimal256(_ARROW_DECIMAL256_MAX_PRECISION, target_scale)
+
+    target_scale = max(scale, value_scale)
+    target_integer_digits = max(precision - scale, integer_digits)
+    target_precision = target_integer_digits + target_scale
+    if target_precision > _ARROW_DECIMAL256_MAX_PRECISION:
+        return pa.string()
+    return _arrow_decimal_type(target_precision, target_scale)
 
 
 def _mysql_field_arrow_type(
@@ -367,9 +439,20 @@ def _build_mysql_arrow_table(cursor) -> pa.Table:
         # a hard-coded ``decimal128(38, 9)``.
         precision = col[4] if len(col) > 4 else None
         scale = col[5] if len(col) > 5 else None
-        arrow_type = _mysql_field_arrow_type(
-            col[1], flag_list[i] or 0, precision=precision, scale=scale
-        )
+        flags = flag_list[i] or 0
+        if col[1] in _mysql_decimal_codes():
+            from MySQLdb.constants import FLAG  # noqa: PLC0415
+
+            arrow_type = _mysql_decimal_type_for_values(
+                precision,
+                scale,
+                is_unsigned=bool(flags & FLAG.UNSIGNED),
+                values=[row[i] for row in rows],
+            )
+        else:
+            arrow_type = _mysql_field_arrow_type(
+                col[1], flags, precision=precision, scale=scale
+            )
         fields.append(pa.field(col[0], arrow_type, nullable=True))
     schema = pa.schema(fields)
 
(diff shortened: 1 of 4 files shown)


core/wren/src/wren/connector/mysql.py
```diff
@@ -295,9 +308,69 @@ def _arrow_decimal_from_mysql_field(
...
+            min(scale, _ARROW_DECIMAL256_MAX_PRECISION - integer_digits),
+        )
+        return pa.decimal256(_ARROW_DECIMAL256_MAX_PRECISION, target_scale)
+
+    target_scale = max(scale, value_scale)
+    target_integer_digits = max(precision - scale, integer_digits)
+    target_precision = target_integer_digits + target_scale
+    if target_precision > _ARROW_DECIMAL256_MAX_PRECISION:
+        return pa.string()
+    return _arrow_decimal_type(target_precision, target_scale)
```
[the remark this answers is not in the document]
Ray0907: Fixed in `bc255dde795271890289679bb03aa4d4e2828bed`. `_mysql_decimal_type_for_values` now rebalances metadata scale within the observed-safe interval `[value_scale, 76 - integer_digits]`. Added unit regressions for p76s75/value12 -> `decimal256(76,74)` and p76s1/value0.12 -> `decimal256(76,2)`; true overflow remains an exact string. Verified 18 targeted decimal tests, 1,098 non-memory unit tests (1 skipped), and 37 live MySQL tests.

core/wren/tests/connectors/test_mysql_connector.py
```diff
@@ -118,6 +118,72 @@ def test_decimal_large_scale(connector: MySqlConnector) -> None:
...
+    expected = str(int(left) * int(right))
+
+    tbl = connector.query(
+        f"SELECT CAST('{left}' AS DECIMAL({left_digits}, 0)) "
+        f"* CAST('{right}' AS DECIMAL({right_digits}, 0)) AS product"
+    )
+
+    assert len(expected) in {77, 80}
+    assert tbl.schema.field("product").type == pa.string()
+    assert tbl.column("product").to_pylist() == [expected]
```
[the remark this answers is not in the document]
Ray0907: Keeping this integration test after verification. On MySQL 8.0.36, both `DECIMAL(65,0) * DECIMAL(12,0)` and `DECIMAL(40,0) * DECIMAL(40,0)` return the exact 77/80-digit values through MySQLdb in this connector path. The test computes the expected result with Python integer multiplication and compares the returned exact string; both cases passed locally and in GitHub MySQL CI. Because the reported behavior is directly observed and reachable, removing this test would reduce regression coverage.

goldmedal:
Reviewed the branch locally. The core fix looks correct — I fuzzed `_mysql_decimal_type_for_values` + `_build_mysql_column` over 30,000 random combinations of metadata (`M`/`D`/unsigned) and 0–3 concrete `Decimal` values and saw no exceptions and no precision loss on either the decimal path or the exact-string path. `tests/unit/test_mysql_helpers.py` passes against this branch (35 passed, 1 skipped). Mirroring the value-aware approach already used in `postgres.py` is the right call.

Three things worth addressing:

### 1. The empty / all-NULL fallback to `pa.string()` is unnecessary and makes the schema data-dependent (design decision)

`core/wren/src/wren/connector/mysql.py:347`

```python
if not shapes:
    if precision > _ARROW_DECIMAL256_MAX_PRECISION:
        return pa.string()
```

Observed on this branch:

```
empty rows, metadata precision 88 -> string
all NULL,   metadata precision 88 -> string
one fitting value                 -> decimal256(76, 0)
```

When there is no non-NULL value there is nothing to convert, so a numeric type cannot overflow — the string fallback buys no safety here, but it does make the same query return a different Arrow type depending on how many rows come back. Concrete case: with `amount DECIMAL(65, 0)`, `SELECT SUM(amount) AS total FROM orders WHERE <predicate>` yields a string column when the predicate matches nothing and `decimal256(76, 0)` when it matches one row. Consumers that key off the Arrow schema (formatting, downstream aggregation, type checks) see the column flip type.

Suggestion: return `_arrow_decimal_type(_ARROW_DECIMAL256_MAX_PRECISION, min(scale, _ARROW_DECIMAL256_MAX_PRECISION))` for the no-values case, matching the "values present and representable" branch.

Note this behaviour is currently pinned by `test_decimal_metadata_above_arrow_limit_without_values_uses_string`, so it reads as deliberate — if it is, it would help to say why in the comment, since the safety argument in the current comment ("no concrete value proves that a narrower numeric schema would be safe") does not apply when there are no values to represent.

### 2. `_mysql_decimal_value_shape` raises on a non-`Decimal` value instead of falling back

`core/wren/src/wren/connector/mysql.py:322`

```python
value = value if isinstance(value, PyDecimal) else PyDecimal(str(value))
```

A `bytes` value raises `decimal.InvalidOperation` (verified: `b"1.5"` → `str()` gives `"b'1.5'"` → `ConversionSyntax`). `_build_mysql_arrow_table` explicitly documents that it may run against "older / non-MySQLdb cursors", and the postgres sibling `_infer_pg_decimal_type` handles this by returning `None` (which degrades to a string column) rather than raising.

Suggestion: treat an unparseable or non-`Decimal` value the same way as a non-finite one — return `None`, so the column degrades to the exact-string path. `_build_mysql_column`'s string branch already decodes `bytes` correctly, so this turns a crash into a correct result.

(The same `PyDecimal(str(v))` pattern pre-exists in `_build_mysql_column`, so this is not a regression from this PR — just the cheapest place to fix it.)

### 3. The metadata-only decimal path is now unreachable in production

`_build_mysql_arrow_table` intercepts decimal type codes before calling `_mysql_field_arrow_type`, so the `if type_code in decimal_codes:` branch at `core/wren/src/wren/connector/mysql.py:403` — and therefore `_arrow_decimal_from_mysql_field` at line 265 — is dead for the only caller in `src/`. That leaves two decimal-typing implementations that can drift, with the unit tests covering mostly the one that no longer runs, so a future regression in the live path would not be caught there.

Suggestion: pick one entry point — either pass `values` through `_mysql_field_arrow_type` so all field typing still flows through it, or drop the decimal branch from `_mysql_field_arrow_type` and keep `_arrow_decimal_from_mysql_field` only if something else needs metadata-only typing.

core/wren/src/wren/connector/mysql.py
```diff
@@ -366,8 +455,13 @@ def _build_mysql_arrow_table(cursor) -> pa.Table:
...
         precision = col[4] if len(col) > 4 else None
         scale = col[5] if len(col) > 5 else None
+        flags = flag_list[i] or 0
         arrow_type = _mysql_field_arrow_type(
-            col[1], flag_list[i] or 0, precision=precision, scale=scale
+            col[1],
+            flags,
+            precision=precision,
+            scale=scale,
+            values=[row[i] for row in rows],
```
goldmedal: **Hot-path regression: `values` is materialized for every column (Major)**

This builds `[row[i] for row in rows]` for **every** column, and then `_build_mysql_column` builds the same per-column list a second time below. Only DECIMAL columns consume `values`, so every non-decimal column pays for a wasted full-result copy.

Measured on 20 columns x 200k rows, best of 3 runs, stub cursor so only this function is timed:

| column type | before | after |
| --- | --- | --- |
| INT | 71 ms | 108 ms (+52%) |
| DECIMAL(12,4) | 483 ms | 1738 ms (+260%) |
| VARCHAR | 448 ms | 481 ms |

The decimal cost is `_mysql_decimal_value_shape` calling `Decimal.as_tuple()` on 4M values. For a semantic layer, decimal-heavy result sets are the common case, not the exception.

Suggestions, cheapest first:

- Materialize the columns once (`columns = list(zip(*rows))`) and reuse them for both the type derivation and `_build_mysql_column`.
- Pass `values` only when `type_code in _mysql_decimal_codes()`, so non-decimal columns are untouched.
- Consider optimistic-then-fallback: try `pa.array(values, type=<metadata type>)` first and run the value-shape scan only when it raises `ArrowInvalid`. That keeps the common path at the previous cost while staying exact for the pathological cases this PR targets.

core/wren/src/wren/connector/mysql.py
```diff
@@ -295,16 +309,84 @@ def _arrow_decimal_from_mysql_field(
...
+        # Preserve as much metadata scale as possible while reserving enough
+        # integer capacity for every fetched value.
+        target_scale = max(
+            value_scale,
+            min(scale, _ARROW_DECIMAL256_MAX_PRECISION - integer_digits),
+        )
+        return pa.decimal256(_ARROW_DECIMAL256_MAX_PRECISION, target_scale)
+
+    target_scale = max(scale, value_scale)
+    target_integer_digits = max(precision - scale, integer_digits)
```
goldmedal: **Unnecessary escalation past the decimal128 boundary (Major)**

Reserving the *declared* integer capacity even when no fetched value needs it lets a widened scale push the column into `decimal256` gratuitously:

```python
_mysql_decimal_type_for_values(40, 4, False, [Decimal("1.234567")])
# -> decimal256(40, 6)
```

That is a declared `DECIMAL(38, 4)` whose observed value needs 1 integer digit and scale 6. `decimal128(38, 6)` holds it comfortably, and decimal128 is considerably better supported downstream than decimal256 (polars and several engines handle decimal256 poorly or not at all).

Please trim unused integer capacity to stay within precision 38 when the observed values fit, and escalate to `decimal256` only when they genuinely do not.

core/wren/src/wren/connector/mysql.py
```diff
@@ -295,16 +309,84 @@ def _arrow_decimal_from_mysql_field(
...
+        precision = min(precision, _ARROW_DECIMAL256_MAX_PRECISION)
+        scale = min(scale, precision)
+        return _arrow_decimal_type(precision, scale)
+    if any(shape is None for shape in shapes):
+        return pa.string()
+
+    integer_digits = max(shape[0] for shape in shapes if shape is not None)
+    value_scale = max(shape[1] for shape in shapes if shape is not None)
+    if integer_digits + value_scale > _ARROW_DECIMAL256_MAX_PRECISION:
+        return pa.string()
```
goldmedal, line 366: **Result-dependent column type deserves an explicit contract (Major)**

The derived type depends on which rows come back, and `MySqlConnector.query()` applies a LIMIT via `_apply_limit`, with the MCP path probing at `limit + 1`:

```python
_mysql_decimal_type_for_values(67, 0, False, [D("9"*70)])             # decimal256(70, 0)
_mysql_decimal_type_for_values(67, 0, False, [D("9"*70), D("9"*77)])  # pa.string()
```

So the same query at `LIMIT 1` and `LIMIT 2` can return a numeric column and a string column, and one outlier row silently degrades the whole column to text. That may well be the right trade-off versus raising, but today it is both silent and undocumented. Please:

- log a warning when a decimal column falls back to `pa.string()`, naming the column, and
- state the contract in the `_build_mysql_arrow_table` / `_mysql_decimal_type_for_values` docstring, so consumers know the decimal type is derived per result set rather than per column.

core/wren/src/wren/connector/mysql.py
```diff
@@ -324,6 +406,13 @@ def _mysql_field_arrow_type(
...
     # ``D > 9``.
     if type_code in decimal_codes:
+        if values is not None:
+            return _mysql_decimal_type_for_values(
+                precision,
+                scale,
+                is_unsigned=bool(flags & FLAG.UNSIGNED),
+                values=values,
+            )
         return _arrow_decimal_from_mysql_field(
```
goldmedal, line 424: **Minor**

- This branch is now unreachable in production: the only caller of `_mysql_field_arrow_type` always passes a list for `values`, so `values is not None` is always true and `_arrow_decimal_from_mysql_field` survives as a test-only helper with 7 unit tests behind it. Fixing the first comment by passing `values` only for decimal type codes makes this branch live again; otherwise please drop it so the tests describe real behaviour.
- `tests/connectors/test_mysql_connector.py:166` — `assert len(expected) in {77, 80}` is loose for a parametrized test; carry the expected digit count in the parameters so each case asserts its own value.
- No coverage for negative over-wide values. I checked and the behaviour is correct (`-9...9` at 77 digits becomes a string, at 76 digits stays `decimal256(76, 0)`), so it is worth pinning down next to `tests/unit/test_mysql_helpers.py:295`.

goldmedal (commented):
Verified locally rather than by reading: fetched the branch, ran `tests/unit/test_mysql_helpers.py` (37 passed), and exercised `_build_mysql_arrow_table` / `_mysql_decimal_type_for_values` directly with a stub cursor. All 9 CI checks are green.

The value-aware approach fixes more than the reported case. On `main`, metadata that understates scale raises `ArrowInvalid: Rescaling Decimal value would cause data loss`; this branch handles it correctly. Nice catch on the `M = length - (0 if unsigned else 1) - ...` docstring too, the previous formula contradicted the code.

Four things I'd like addressed before merge, left inline: a measured hot-path regression, gratuitous escalation to `decimal256`, an undocumented result-dependent column type, and a path that is now unreachable in production.

goldmedal:
Hi @Ray0907, to clear my review queue, I converted this PR to a draft. After addressing the comment, feel free to request my review.

Ray0907:
@goldmedal Ready for re-review.

Addressed all remaining comments:
- use metadata-first decimal conversion and scan concrete values only after Arrow conversion fails
- materialize result columns once and keep Decimal128 when observed values fit
- document result-dependent decimal types and warn with the column name on exact-string fallback
- restore the metadata-only production path
- tighten digit-count assertions and cover negative 76/77-digit boundaries

Rebased onto the latest `main`. Local validation: 1,227 non-memory unit tests passed (2 skipped), 37 MySQL connector tests passed, and Ruff format/lint passed.

goldmedal:
LGTM. Recording what I ran on `60fe1777`, so the evidence is on the record rather than in my terminal.

Verified against three real MySQL-protocol servers in containers: MySQL 8.0.36, MariaDB 11.8.9, and Apache Doris 2.1 (all-in-one).

### A/B against `main`

Same server, same query, only `connector/mysql.py` swapped.

MySQL 8.0.36:

| case | `main` | this PR |
| --- | --- | --- |
| `DECIMAL(65,30)` column | `ArrowInvalid` | `decimal256(65, 30)`, exact |
| `DECIMAL(39,0)` column | `ArrowInvalid` | `decimal256(39, 0)`, exact |
| `DECIMAL(38,4)` column | `decimal128(38, 4)`, exact | unchanged |
| 66-digit addition | `ArrowInvalid` | `decimal256(66, 0)`, exact |
| 77- / 80-digit multiplication | `ArrowInvalid` | exact string |

Apache Doris 2.1 with `enable_decimal256=true`:

| case | `main` | this PR |
| --- | --- | --- |
| `DECIMAL(50,0)` column | `ArrowInvalid` | `decimal256(50, 0)`, exact |
| `DECIMAL(76,0)` column | `ArrowInvalid` | `decimal256(76, 0)`, exact |
| `DECIMAL(76,30)` column | `ArrowInvalid` | `decimal256(76, 30)`, exact |

Two things this surfaced. On `main` a plain `DECIMAL(39,0)` column is already unreadable, so the fix covers more than the expression-metadata case in the description. And since the connector CI matrix is MySQL and Postgres only, the Doris half of the title had not been executed anywhere — it does work, at `DECIMAL(50,0)`, `DECIMAL(76,0)` and `DECIMAL(76,30)`.

### How often the value-aware fallback actually fires

Swept each server, deriving precision/scale exactly as the connector does and comparing against the fetched values:

| server | decimal columns exercised | metadata understated the value | wrong results |
| --- | --- | --- | --- |
| MySQL 8.0.36 | 2,613 | 194 | 0 |
| MariaDB 11.8.9 | 2,255 | 122 | 0 |
| Doris 2.1 (decimal256 on) | 384 | 0 | 0 |

All 316 understatements are MySQL's 65-digit expression clamp (metadata says 65, the value carries 66-70 digits). Every one took the fallback and came back exact, so this path is not an edge case — ordinary wide arithmetic hits it. Doris never understates: with `enable_decimal256=false` it raises `Arithmetic overflow` (E-124) rather than returning a value its metadata cannot describe.

### Other checks

| | result |
| --- | --- |
| `tests/connectors/test_mysql_connector.py` + `test_mysql.py` against real MySQL 8.0.36 | 37 passed |
| unit suite | 1,227 passed, 1 skipped |
| `ruff format --check src/` / `ruff check src/` (what CI lints) | clean |
| fuzz of `_mysql_decimal_type_for_values` + `_build_mysql_column`, 40,000 metadata/value combinations | 0 exceptions, 0 value loss |

Performance, 20 `DECIMAL(12,4)` columns x 100k rows, best of 3, stub cursor: `main` 602 ms, the earlier value-scan-first revision of this PR 1538 ms, this revision **490 ms**. The optimistic restructure came out faster than `main`, and `DECIMAL(38,4)` + `1.234567` correctly stays `decimal128(38, 6)` rather than escalating to `decimal256`.

### One thing to know about later, not a merge blocker

The optimistic path treats an exception as the signal to re-derive from values, which is sound only if PyArrow always raises when the type cannot hold the value. It does not: for `decimal128` targets, values with 39-40 significant digits can wrap int128 and return a different number, sign included, with no exception (PyArrow 23.0.1).

```python
>>> v = Decimal("654920502397576889438654552236818180.152")
>>> pa.array([v], type=pa.decimal128(38, 2), from_pandas=False).to_pylist()[0]
Decimal('-25644231444300037488094662626718242.76')            # silently wrong
>>> pa.array([v], type=pa.decimal256(38, 2), from_pandas=False)
ArrowInvalid: Rescaling Decimal value would cause data loss   # decimal256 raises
```

4,000 random values per row into `decimal128(38, s)`: 39 digits -> 68 silently wrong / 3932 raised; 40 digits -> 4 / 3996; 38 digits and 45+ digits -> 0 wrong.

I could not reach this on any real server, which is why it is a footnote. It needs a metadata-derived `decimal128` to receive a 39-40 digit value, and across the 5,252 decimal columns above that never happened — MySQL and MariaDB only understate via the 65-digit clamp, which lands in `decimal256` where PyArrow raises correctly, and Doris refuses the query instead.

If it ever needs hardening, converting at `decimal256` and casting down closes it, since `decimal256` raises correctly:

```python
if pa.types.is_decimal128(arrow_type):
    wide = pa.decimal256(arrow_type.precision, arrow_type.scale)
    return _build_mysql_column(values, wide).cast(arrow_type), arrow_type
return _build_mysql_column(values, arrow_type), arrow_type
```

Measured on this branch: 545 ms (+11%) instead of 490 ms, silent corruption in 20k randomised runs drops from 3 to 0, and `tests/unit/test_mysql_helpers.py` passes unchanged (45 passed). Worth remembering if the Doris side grows its own decimal semantics, since nothing in CI would catch it there.

goldmedal (approved):
Thanks @Ray0907 👍

PR #2665, opened 2026-08-12
fix(memory): skip unnamed cubes/measures/dimensions in schema_indexer

### LINKED ISSUE #2591: refactor(memory): align cube/measure/dimension guards with the unnamed-skip policy in schema_indexer
## Summary

After #2533, `core/wren/src/wren/memory/schema_indexer.py` skips **unnamed** models, relationships, and views — an entry that isn't a dict, or has no `name`, is dropped rather than indexed. Cubes and their children were left on the older, weaker guard: `isinstance(x, dict)` only, with no name check.

The result is two different policies in one module, and nameless records reaching the vector index.

## Reproduction

Verified on `4316695f`:

```python
from wren.memory.schema_indexer import extract_schema_items

items = extract_schema_items({
    "cubes": [{
        "baseObject": "orders",
        "measures": [{"expression": "sum(x)"}],
        "dimensions": [{"expression": "y"}],
    }]
})
for i in items:
    print(i["item_type"], repr(i["item_name"]))

# cube            ''
# measure         ''
# cube_dimension  ''
```

Compare with a nameless model, which #2533 now drops:

```python
extract_schema_items({"models": [{"columns": []}]})   # []
```

The record builders use `cube.get("name", "")` throughout (`_cube_record`, `_measure_record`, `_cube_dimension_record`, `_time_dimension_record`), so nothing raises — the records are simply created with an empty `item_name` and embedded. They then occupy rows in `schema_items` and can be returned by `_search_schema`, contributing nameless entries to the schema context handed to the LLM.

## Why this wasn't folded into #2533

Raised during review of #2533 and deliberately deferred: unlike the model/relationship/view case, this is not fixing a crash. Cubes with no name index successfully today, so switching them to the skip policy changes behavior for manifests that currently work. That deserves its own change rather than riding along in a resilience fix.

## Scope

Decide the intended policy, then apply it uniformly:

1. **Skip unnamed** — align cubes/measures/dimensions/timeDimensions with models/relationships/views, adding `and x.get("name")` at the four iteration sites in `extract_schema_items` and the corresponding ones in `describe_schema` / `_describe_cube`.
2. **Or keep them** — if a nameless cube child is considered legitimate (e.g. a measure identified only by expression), leave the behavior and document why the two policies differ, so the asymmetry reads as intentional rather than an oversight.

(1) is the consistent reading and matches the direction of #2533, but it is a real behavior change for anyone with a partially-authored cube, so it should be a deliberate call. Whichever way it goes, a test asserting the chosen policy would stop the two halves of this module from drifting again.

AmirF194:
## Summary

`schema_indexer.py`'s cube iteration used `isinstance(x, dict)` only, while the model/relationship/view iteration added by #2533 also checks `x.get("name")`. A cube, measure, dimension, or timeDimension with no name was indexed anyway, with `item_name=""`, unlike a nameless model (dropped by #2533).

Adds `and x.get("name")` at the four iteration sites, in both `extract_schema_items` and the `describe_schema` / `_describe_cube` path, per the issue's own stated preference for option 1 (skip unnamed) over option 2 (keep and document).

## A documented choice this respects rather than reverses

`test_extract_unnamed_cube_falls_back_to_unnamed_form` (and the docstring on `_as_list`) deliberately document that a cube is not required to have a name, and pin the `"cube (unnamed): 'measures' must be a list, got int"` message `_as_list` produces for a malformed nested field on an unnamed cube. That test still passes: the child collections are resolved through `_as_list` (so a non-list `measures`/`dimensions`/`timeDimensions` still raises) before the new name check, and only the append-to-output step is skipped when the cube itself has no name.

## Verification

- New regression test (`test_schema_indexer_unnamed_cubes.py`) reproduces the issue's own repro (unnamed cube, unnamed measure/dimension/timeDimension): fails on unmodified `main` (asserts caught 2 rendered `### Cube:` headers instead of 1, and an empty `item_name` in the extracted items) and passes on this branch. Ran both in Docker (`python:3.11-slim`, `uv sync`, `wren.memory.schema_indexer.__file__` provenance-printed).
- Full `tests/unit/` suite, matching this repo's `test-unit` CI job (excludes `test_memory.py`/`test_mcp_server.py`): 1106 passed, 2 skipped.
- `ruff format --check` / `ruff check` on `src/` clean.
- Did not check: line-coverage of the diff via `coverage run`: this repo's `wren_core` PyO3 extension panics under `coverage run` with an unrelated `env_logger` double-init error in this environment. I could not get a coverage report. I reasoned through branch coverage manually against the test scenarios instead of measuring it.

Fixes #2591

## Summary by CodeRabbit

* **Bug Fixes**
  * Unnamed cubes are no longer included in schema descriptions or indexing results.
  * Unnamed measures, dimensions, and time dimensions are excluded from extracted schema data.
  * Named schema elements continue to be included as expected.

* **Tests**
  * Added coverage for handling unnamed cubes and child schema elements.

### DIFF
--- a/core/wren/src/wren/memory/schema_indexer.py
+++ b/core/wren/src/wren/memory/schema_indexer.py
@@ -230,12 +230,25 @@ def _describe_relationship(rel: dict, lines: list[str]) -> None:
 def _describe_cube(cube: dict, lines: list[str]) -> None:
     name = cube.get("name", "")
     base = cube.get("baseObject", "?")
-    lines.append(f"### Cube: {name} (base: {base})")
+    # Type-validate children via _as_list before the unnamed-cube skip below.
     measures = [
         m
         for m in (_as_list(cube.get("measures"), "cube", name, "measures"))
-        if isinstance(m, dict)
+        if isinstance(m, dict) and m.get("name")
     ]
+    dims = [
+        d
+        for d in (_as_list(cube.get("dimensions"), "cube", name, "dimensions"))
+        if isinstance(d, dict) and d.get("name")
+    ]
+    tdims = [
+        td
+        for td in (_as_list(cube.get("timeDimensions"), "cube", name, "timeDimensions"))
+        if isinstance(td, dict) and td.get("name")
+    ]
+    if not name:
+        return
+    lines.append(f"### Cube: {name} (base: {base})")
     if measures:
         lines.append("  Measures:")
         for m in measures:
@@ -248,11 +261,6 @@ def _describe_cube(cube: dict, lines: list[str]) -> None:
             if expr:
                 line += f": {expr}"
             lines.append(line)
-    dims = [
-        d
-        for d in (_as_list(cube.get("dimensions"), "cube", name, "dimensions"))
-        if isinstance(d, dict)
-    ]
     if dims:
         lines.append("  Dimensions:")
         for d in dims:
@@ -265,11 +273,6 @@ def _describe_cube(cube: dict, lines: list[str]) -> None:
             if expr and expr != dname:
                 line += f": {expr}"
             lines.append(line)
-    tdims = [
-        td
-        for td in (_as_list(cube.get("timeDimensions"), "cube", name, "timeDimensions"))
-        if isinstance(td, dict)
-    ]
     if tdims:
         lines.append("  Time dimensions:")
         for td in tdims:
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
Thanks @AmirF194

AmirF194:
Thanks for the review and merge!

PR #2666, opened 2026-08-13
perf(core-py): remove per-context call lock for same-context concurrency

### LINKED ISSUE #2504: Follow-up (wren-core-py): document + test physical-table visibility through exec_ctx after per-call catalog snapshot

ttw225:
Removes the per-context `call_lock` from `wren-core-py`, letting engine calls on one `SessionContext` run concurrently. Single-thread throughput is unchanged and transform throughput scales 3.19x at 8 threads; the safety argument, the documented contract, and a dedicated concurrency regression are below.

## Why

`PySessionContext` serializes every engine call with a per-context `call_lock`. Its premise is recorded on the field itself:

> "Serializes engine calls on this context. `transform_sql` re-applies the MDL catalog onto the context's shared catalog list on every call (`apply_wren_on_ctx` -> `register_table_with_mdl`), while `query`, `dry_run`, and `list_tables` read that same shared list, so two concurrent calls on the same context can observe each other's half-registered catalogs. The GIL used to provide this serialization implicitly; now that the GIL is released around blocking work, keep the same per-context ordering with a lock. Calls on different contexts run in parallel."

That premise has been removed by this PR's prerequisites. #2495 gives each transform a private top-level catalog-list snapshot (with a pointer-identity guard pinning the intended provider sharing), and #2619 makes the RLAC cycle-detection stack per-invocation with a `RefCell` whose `!Sync` property keeps it off the shared rule. Distinct-name registrations update the live-shared provider, and `list_tables` is explicitly best-effort. The lock no longer protects the state described by the comment; it only serializes otherwise independent calls on one context.

The old mode-2 test comment recorded the same warning ("This is the mode that catches the same-context catalog race if the per-context call lock is ever removed while the GIL is released"). It was correct before the prerequisites; that mode now pins the supported post-prerequisite behavior, and a dedicated stronger regression (below) covers the paths it lacked.

## What

- Remove `call_lock`: the field and its doc, three constructor initializations, all ten guard acquisitions, and the `lock_calls()` helper. No replacement synchronization is added.
- Document the concurrency contract: the README Concurrency section is the normative copy; the struct doc describes the mechanism and points to it. Boundaries stated explicitly: state-mutating statements accepted by `query()` execute but are outside the contract, and `ANALYZE`-prefixed `dry_run` input becomes `EXPLAIN ANALYZE` and executes (verified: `dry_run("ANALYZE INSERT ...")` mutates the target table). Same-name concurrent registration is unsupported; `list_tables` is best-effort; overlapping `load_mdl` raises `RuntimeError` via PyO3's exclusive borrow.
- Add `test_same_context_concurrency.py`: 8 barrier-started threads on one shared context — an RLAC subquery contention phase (nested analysis on the long-lived `ModelAnalyzeRule`), then a named schedule mixing transforms, semantic/physical queries (content-asserted, RLAC-filtered rows), `dry_run`, function lookup, enumeration with a physical-only sentinel, and distinct-name Parquet/CSV registration. The hammer runs in a subprocess with a deadline, `faulthandler`, and signal-decoded exit reporting, and works under both default and `--import-mode=importlib` pytest modes.

## Benchmark (same-context, one shared SessionContext)

Release builds of base and candidate from the same HEAD, three alternating baseline/candidate round pairs (before-1 → after-1 → before-2 → after-2 → before-3 → after-3), every rep a ≥3s measured window, medians recomputed from raw counts:

| Workload | T=1 | T=2 | T=4 | T=8 |
|---|---:|---:|---:|---:|
| transform | 1.029x | 1.291x | 2.041x | 3.190x |
| query_small | 0.995x | 1.521x | 2.251x | 2.488x |
| query_heavy | 1.013x | 1.586x | 2.252x | 3.149x |
| mixed | 0.997x | 1.708x | 2.599x | 3.539x |

Transform T=1→T=8 scaling is 1.006x before (the lock-serialization plateau) and 3.121x after. One after-round contained a transient environmental slowdown (preserved in the raw data), so the supported claim is preservation of aggregate T=1 throughput across three rounds, not that every individual round avoided slowdown. These are local same-context microbenchmark numbers, not a production workload or released-performance claim.

## Residual coverage

No committed test overlaps `load_mdl` with another call: the deterministic embedded-interpreter test cannot link because `wren-core-base`'s `python-binding` feature enables pyo3's `extension-module`, which suppresses libpython linking. The documented rejection relies on PyO3's generated exclusive borrow for `&mut self` (overlap raises `RuntimeError`, verified manually in both directions). A cross-crate feature split would make the test buildable and could be a follow-up if you want it.

## Question for the maintainer

`query()` accepts state-mutating SQL, and `ANALYZE`-prefixed `dry_run` input executes. The removed lock previously serialized those paths; this PR documents them as outside the same-context concurrency contract instead of changing what is accepted. Should a follow-up enforce read-only inputs with DataFusion `SQLOptions`, or should the binding intentionally keep them accepted-but-unsupported? Restricting them here would be a user-visible behavior change beyond #2504.

## Test plan

- `cargo test` (default features) and `cargo test --no-default-features`: 37 passed each
- `uv run --no-sync pytest`: 41 passed; the new hammer passes 10 consecutive runs and under `--import-mode=importlib`
- Downstream `core/wren` DataFusion connector tests against the locally built wheel: 9 passed
- Benchmark protocol and raw artifacts (480 reps, independently recomputed) available on request

Closes #2504

## Summary by CodeRabbit

* **New Features**
  * Session contexts now support concurrent queries, SQL transformations, dry runs, function lookups, table enumeration, access-control analysis, and file registrations.
  * Concurrent operations maintain isolated per-call state and provide consistent results.

* **Bug Fixes**
  * Prevented unnecessary serialization of session operations, improving responsiveness during parallel workloads.
  * Improved reliability when listing tables and registering data sources concurrently.

* **Documentation**
  * Added guidance covering concurrency behavior, registration constraints, table listing, and model loading.

### DIFF
--- a/core/wren-core-py/README.md
+++ b/core/wren-core-py/README.md
@@ -63,11 +63,31 @@ Visibility contract:
 - Brand-new *top-level* catalogs are the exception — they must exist before
   MDL construction, `load_mdl`, or a transform, each of which snapshots the
   top-level catalog list.
-- `load_mdl` must not overlap other calls on the same context; overlapping
-  calls raise `RuntimeError`.
+- For `load_mdl`'s overlap rule, see the Concurrency section below.
 
 For complete runnable examples (fixture files, matching manifests, decoding
 the returned bytes), see `tests/test_physical_tables.py`.
+
+### Concurrency
+
+Calls on one `SessionContext` run in parallel. Each `transform_sql` works
+on a private top-level catalog snapshot and analyzer state is
+per-invocation, so supported concurrent calls never observe each other's
+intermediate state. The contract:
+
+- Concurrent execution is supported for the read-only inputs accepted by
+  `transform_sql` and `query`, and for the registration APIs. `dry_run` is
+  concurrency-safe for statements that `EXPLAIN` only plans. An
+  `ANALYZE`-prefixed input becomes `EXPLAIN ANALYZE` and executes; like a
+  state-mutating statement accepted by `query()`, it is outside the
+  concurrency contract. Function lookup methods are read-only and
+  concurrency-safe.
+- `register_parquet` / `register_csv` are safe under distinct table names;
+  registering the same name concurrently is unsupported.
+- `list_tables` is a best-effort enumeration: registrations that land
+  mid-call may or may not appear, but the result is always well-formed.
+- `load_mdl` must not overlap other calls on the same context; overlapping
+  calls raise `RuntimeError`.
 
 ## Developer Guide
 
--- a/core/wren-core-py/src/context.rs
+++ b/core/wren-core-py/src/context.rs
@@ -72,6 +72,14 @@ fn shared_runtime() -> PyResult<Arc<Runtime>> {
 }
 
 /// The Python wrapper for the Wren Core session context.
+///
+/// Calls on one context can run concurrently. Each `transform_sql` applies
+/// the MDL onto a private top-level catalog snapshot (see
+/// `clone_catalog_list` in wren-core's `mdl::context`), and analyzer rules
+/// keep their mutable state per invocation. See the Concurrency section in
+/// `README.md` for the supported operation contract. `load_mdl` takes
+/// `&mut self`, so PyO3's exclusive borrow rejects overlapping calls on the
+/// same context with a `RuntimeError`.
 #[pyclass(name = "SessionContext")]
 pub struct PySessionContext {
     /// Base context — physical tables are registered here.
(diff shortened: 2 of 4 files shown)


goldmedal (approved):
Thanks @ttw225, this PR look good to me.  I'll check the mutating SQL issue 👍

PR #2668, opened 2026-08-14
fix(cube): load/validate like views; fail loud on mdl.json

Bartok9:
## Failure
`load_cubes` dropped bad YAML quietly with no `validate_project` cube-member checks. CLI `wren cube list`/`describe` crashed or went blank on malformed `target/mdl.json` (build artifact).

## Fix
- Loader + `validate_project` follow #2613 / #2604 (drop + report raw YAML).
- CLI fails loud at `_load_manifest_dict` with a `wren context build` hint.
- MCP listing only reads string names from dict members (loader already `list[dict]`).

Replaces #2615 and #2617 (goldmedal 2026-08-14).

## Verification
`pytest tests/unit/test_context.py -k cube tests/unit/test_cube_cli.py` — 38 passed.

## Duplicate check
No open PR with this loader+validate+loud-CLI split.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cube loading and validation for malformed YAML, metadata, and member definitions.
  - Explicitly empty member lists are handled correctly, while missing lists remain valid and malformed collections produce clear errors.
  - Added clearer repair guidance for invalid entries and members without names.
  - Cube listings now include only valid named measures, dimensions, and time dimensions, excluding malformed entries.
  - Improved validation messages for missing cube names by using a clear fallback label.

### DIFF
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -660,6 +660,53 @@ def load_cubes(project_path: Path) -> list[dict]:
     return _load_cubes_v2(project_path)
 
 
+def _report_malformed_cube_members(
+    errors: list[ValidationError], src_path: str, raw: dict
+) -> None:
+    """Report non-list containers and non-mapping member entries."""
+    cube_label = raw.get("name") if isinstance(raw.get("name"), str) else src_path
+    for key in ("measures", "dimensions", "time_dimensions"):
+        val = raw.get(key)
+        if val is None:
+            continue
+        if not isinstance(val, list):
+            errors.append(
+                ValidationError(
+                    "error",
+                    f"{src_path} > {cube_label} > {key}",
+                    f"'{key}' must be a list, got {type(val).__name__}",
+                )
+            )
+            continue
+        for i, item in enumerate(val):
+            if not isinstance(item, dict):
+                errors.append(
+                    ValidationError(
+                        "error",
+                        f"{src_path} > {cube_label} > {key}[{i}]",
+                        f"{key[:-1]} entry must be a mapping, got {type(item).__name__}",
+                    )
+                )
+
+
+def _normalise_cube_member_lists(cube: dict) -> dict:
+    """Drop non-mapping measure/dimension/time_dimension entries.
+
+    ``validate_project`` re-reads the raw YAML so hand-edited mistakes are
+    still reported rather than vanishing quietly (same pattern as views /
+    relationships in #2604 / #2613).
+    """
+    for key in ("measures", "dimensions", "time_dimensions"):
+        raw = cube.get(key)
+        if raw is None:
+            continue
+        if not isinstance(raw, list):
+            cube[key] = []
+            continue
+        cube[key] = [item for item in raw if isinstance(item, dict)]
+    return cube
+
+
 def _load_cubes_v1(project_path: Path) -> list[dict]:
     """Legacy: load cube YAML files from project_path/cubes/*.yml."""
     cubes_dir = project_path / "cubes"
@@ -670,7 +717,7 @@ def _load_cubes_v1(project_path: Path) -> list[dict]:
         data = yaml.safe_load(f.read_text(encoding="utf-8"))
         if isinstance(data, dict):
             data["_source_file"] = f.name
-            cubes.append(data)
+            cubes.append(_normalise_cube_member_lists(data))
     return cubes
 
 
@@ -692,7 +739,7 @@ def _load_cubes_v2(project_path: Path) -> list[dict]:
         data = yaml.safe_load(meta_file.read_text())
         if isinstance(data, dict):
             data["_source_file"] = str(meta_file.relative_to(cubes_dir))
-            cubes.append(data)
+            cubes.append(_normalise_cube_member_lists(data))
     return cubes
 
 
@@ -1267,6 +1314,45 @@ def validate_project(project_path: Path) -> list[ValidationError]:
                     "warning", f"relationships > {rel_name}", "missing join_type"
                 )
             )
+
+    # Cube YAML may contain non-mapping files / member-array junk.
+    # load_cubes() drops those (matching views / relationships). Re-read the
+    # raw files here so hand-edits are reported rather than vanishing quietly.
+    cubes_dir = project_path / "cubes"
+    if cubes_dir.is_dir():
+        if sv == 1:
+            for f in sorted(cubes_dir.glob("*.yml")):
+                raw = yaml.safe_load(f.read_text(encoding="utf-8"))
+                src_path = f"cubes/{f.name}"
+                if not isinstance(raw, dict):
+                    errors.append(
+                        ValidationError(
+                            "error",
+                            src_path,
+                            f"cube file must be a mapping, got {type(raw).__name__}",
+                        )
+                    )
+                    continue
+                _report_malformed_cube_members(errors, src_path, raw)
+        else:
+            for d in sorted(cubes_dir.iterdir()):
+                if not d.is_dir():
+                    continue
+                meta_file = d / "metadata.yml"
+                if not meta_file.exists():
+                    continue
+                raw = yaml.safe_load(meta_file.read_text(encoding="utf-8"))
+                src_path = f"cubes/{d.name}/metadata.yml"
+                if not isinstance(raw, dict):
+                    errors.append(
+                        ValidationError(
+                            "error",
+                            src_path,
+                            f"cube metadata must be a mapping, got {type(raw).__name__}",
+                        )
+                    )
+                    continue
+                _report_malformed_cube_members(errors, src_path, raw)
 
     # Check cubes — only structural / reference checks here. Deep validation
     # (measure cycles, hierarchy levels) runs Rust-side in
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -664,9 +664,15 @@ def _report_malformed_cube_members(
     errors: list[ValidationError], src_path: str, raw: dict
 ) -> None:
     """Report non-list containers and non-mapping member entries."""
-    cube_label = raw.get("name") if isinstance(raw.get("name"), str) else src_path
+    # Prefer cube name; bare path twice reads poorly when name is missing.
+    if isinstance(raw.get("name"), str) and raw["name"]:
+        cube_label = raw["name"]
+    else:
+        cube_label = "cube"
     for key in ("measures", "dimensions", "time_dimensions"):
         val = raw.get(key)
+        # Bare `dimensions:` parses to None and means "no members" (loader
+        # normalises to []). Only non-list non-null values are malformed.
         if val is None:
             continue
         if not isinstance(val, list):
@@ -687,20 +693,32 @@ def _report_malformed_cube_members(
                         f"{key[:-1]} entry must be a mapping, got {type(item).__name__}",
                     )
                 )
+                continue
+            if not isinstance(item.get("name"), str):
+                errors.append(
+                    ValidationError(
+                        "error",
+                        f"{src_path} > {cube_label} > {key}[{i}]",
+                        f"{key[:-1]} entry must have a string 'name'",
+                    )
+                )
 
 
 def _normalise_cube_member_lists(cube: dict) -> dict:
     """Drop non-mapping measure/dimension/time_dimension entries.
+
+    Explicit null keys (YAML ``dimensions:``) become empty lists so the
+    build artifact never carries null member containers into the CLI.
 
     ``validate_project`` re-reads the raw YAML so hand-edited mistakes are
     still reported rather than vanishing quietly (same pattern as views /
     relationships in #2604 / #2613).
     """
     for key in ("measures", "dimensions", "time_dimensions"):
-        raw = cube.get(key)
-        if raw is None:
+        if key not in cube:
             continue
-        if not isinstance(raw, list):
+        raw = cube[key]
+        if raw is None or not isinstance(raw, list):
             cube[key] = []
             continue
         cube[key] = [item for item in raw if isinstance(item, dict)]
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -714,7 +714,10 @@ def _load_cubes_v1(project_path: Path) -> list[dict]:
         return []
     cubes = []
     for f in sorted(cubes_dir.glob("*.yml")):
-        data = yaml.safe_load(f.read_text(encoding="utf-8"))
+        try:
+            data = yaml.safe_load(f.read_text(encoding="utf-8"))
+        except yaml.YAMLError:
+            continue
         if isinstance(data, dict):
             data["_source_file"] = f.name
             cubes.append(_normalise_cube_member_lists(data))
@@ -736,7 +739,10 @@ def _load_cubes_v2(project_path: Path) -> list[dict]:
         meta_file = d / "metadata.yml"
         if not meta_file.exists():
             continue
-        data = yaml.safe_load(meta_file.read_text())
+        try:
+            data = yaml.safe_load(meta_file.read_text())
+        except yaml.YAMLError:
+            continue
         if isinstance(data, dict):
             data["_source_file"] = str(meta_file.relative_to(cubes_dir))
             cubes.append(_normalise_cube_member_lists(data))
@@ -1322,8 +1328,18 @@ def validate_project(project_path: Path) -> list[ValidationError]:
     if cubes_dir.is_dir():
         if sv == 1:
             for f in sorted(cubes_dir.glob("*.yml")):
-                raw = yaml.safe_load(f.read_text(encoding="utf-8"))
                 src_path = f"cubes/{f.name}"
+                try:
+                    raw = yaml.safe_load(f.read_text(encoding="utf-8"))
+                except yaml.YAMLError as e:
+                    errors.append(
+                        ValidationError(
+                            "error",
+                            src_path,
+                            f"invalid YAML: {e}",
+                        )
+                    )
+                    continue
                 if not isinstance(raw, dict):
                     errors.append(
                         ValidationError(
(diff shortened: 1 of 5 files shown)


Bartok9:
@goldmedal Consolidated as you asked on #2615 / #2617.

- YAML: `load_cubes` drops non-mapping members; `validate_project` re-reads raw cube files (same pattern as #2613 / #2604).
- CLI `mdl.json`: fail loud at `_load_manifest_dict` (covers `list` + `describe`) with a `wren context build` hint — no skip-to-blank.
- No dead MCP cube-level guards; MCP only takes string names from dict members.

#2615 and #2617 stay closed.

Bartok9:
Thanks CodeRabbit — both inline notes are real, I'll take them on this branch rather than defer.

1. **YAMLError escaping `validate_project`:** catch in both cube loaders (skip that file) and again in the raw-file pass so invalid YAML becomes a `ValidationError` with `src_path` instead of aborting before the report loop.
2. **Non-string member `name`:** reject in `list_cubes` (measures / dimensions / timeDimensions) with the existing malformed-MDL + rebuild hint, plus a regression for numeric names.

Will push those two plus tests; description Failure section / docstring nits are separate if they still fail the template check after.

core/wren/src/wren/context.py
```diff
@@ -660,17 +660,67 @@ def load_cubes(project_path: Path) -> list[dict]:
...
+    """Drop non-mapping measure/dimension/time_dimension entries.
+
+    ``validate_project`` re-reads the raw YAML so hand-edited mistakes are
+    still reported rather than vanishing quietly (same pattern as views /
+    relationships in #2604 / #2613).
+    """
+    for key in ("measures", "dimensions", "time_dimensions"):
+        raw = cube.get(key)
+        if raw is None:
+            continue
```
goldmedal: **Blocking — a bare `measures:` / `dimensions:` key still produces `null`, and that breaks both the CLI and wren-core.**

`None` skips normalisation here, so an ordinary empty YAML key survives into `mdl.json` as `null`. Reproduced on this branch with nothing hand-edited:

```yaml
# cubes/om/metadata.yml
measures:
  - name: total
    expression: SUM(o_totalprice)
    type: double
dimensions:
time_dimensions:
```

```console
$ wren context validate   → 1 warning(s), 0 errors.
$ wren context build      → Built: 1 models, 0 views      # mdl.json: "dimensions": null
$ wren cube list
TypeError: 'NoneType' object is not iterable               # cube_cli.py:229, empty output, no hint
$ wren cube query --cube order_metrics --measures total --sql-only
Error: Invalid manifest JSON: invalid type: null, expected a sequence at line 35 column 24
```

So this is not only the listing crash the PR targets — the manifest is rejected by wren-core, i.e. `wren cube query` / MCP `query_cube` on such a project cannot work at all. `[]` is accepted (verified directly against `wren_core.SessionContext`), and normalising here fixes producer and every consumer at once:

```suggestion
        raw = cube.get(key)
        if raw is None:
            cube[key] = []
            continue
```

Worth a regression test on both sides — the current tests only cover `dimensions: not-a-list` (a `str`), never the empty-key `None`:

```python
def test_load_cubes_normalises_empty_member_keys(tmp_path):
    _make_v2_cube_project(tmp_path)
    _write_cube(
        tmp_path,
        "om",
        "name: order_metrics\nbase_object: orders\nmeasures:\ndimensions:\ntime_dimensions:\n",
    )
    cubes = load_cubes(tmp_path)
    assert cubes[0]["measures"] == []
    assert cubes[0]["dimensions"] == []
    assert cubes[0]["time_dimensions"] == []
```

core/wren/src/wren/cube_cli.py
```diff
@@ -150,17 +150,75 @@ def _load_manifest_dict(mdl: str | None) -> dict:
...
+            typer.echo(
+                f"Error: malformed cubes in mdl.json (entry {i} is not an object).\n"
+                "  Hint: re-run `wren context build`.",
+                err=True,
+            )
+            raise typer.Exit(1)
+        for key in ("measures", "dimensions", "timeDimensions"):
+            members = cube.get(key)
+            if members is None:
+                continue
```
goldmedal: **Blocking (reader half of the same bug) — the loud-fail guard skips `null`, so the crash walks straight past it.**

`list_cubes` reads members with `cube.get(key, [])` (no `or []`, unlike the identical code in `mcp_server.py`), so a `null` that this `continue` waves through raises a raw `TypeError` in `", ".join(...)` two frames later — the exact failure mode this helper exists to replace.

With the `_normalise_cube_member_lists` fix in place a rebuild yields `[]`, so failing loud here is truthful and consistent with the rest of the helper:

```suggestion
            members = cube.get(key)
            if members is None:
                name = cube.get("name", f"entry {i}")
                typer.echo(
                    f"Error: malformed cubes in mdl.json ({name!r} {key} is null, expected a list).\n"
                    "  Hint: re-run `wren context build`.",
                    err=True,
                )
                raise typer.Exit(1)
```

If you would rather not hard-fail on artifacts built by an older CLI, the tolerant alternative is `cube.get(key) or []` at the three join sites (lines 228-232) — same as `mcp_server.py` already does. Either way the producer fix is the one that must land; this one only decides how old `mdl.json` files behave.

core/wren/src/wren/context.py
```diff
@@ -660,17 +660,67 @@ def load_cubes(project_path: Path) -> list[dict]:
...
+        for i, item in enumerate(val):
+            if not isinstance(item, dict):
+                errors.append(
+                    ValidationError(
+                        "error",
+                        f"{src_path} > {cube_label} > {key}[{i}]",
+                        f"{key[:-1]} entry must be a mapping, got {type(item).__name__}",
+                    )
+                )
```
goldmedal, line 695: **A member without a `name` passes validation, then hard-fails at `cube list` with a hint that provably cannot help.**

```console
$ cat cubes/om/metadata.yml
name: order_metrics
base_object: orders
measures:
  - expression: SUM(o_totalprice)   # no name
    type: double

$ wren context validate  → 1 warning(s), 0 errors.
$ wren context build     → Built: 1 models, 0 views
$ wren cube list         → Error: ... ('order_metrics' measures[0].name is not a string).
                             Hint: re-run `wren context build`.
$ wren context build && wren cube list   → byte-identical error (the loop never terminates)
```

The fault is in the project YAML, so validation is where it should surface — the new `mdl.json` check then only has to catch genuine producer bugs:

```suggestion
        for i, item in enumerate(val):
            if not isinstance(item, dict):
                errors.append(
                    ValidationError(
                        "error",
                        f"{src_path} > {cube_label} > {key}[{i}]",
                        f"{key[:-1]} entry must be a mapping, got {type(item).__name__}",
                    )
                )
                continue
            if not isinstance(item.get("name"), str):
                errors.append(
                    ValidationError(
                        "error",
                        f"{src_path} > {cube_label} > {key}[{i}]",
                        f"{key[:-1]} entry must have a string 'name'",
                    )
                )
```

Unrelated nit on line 667: when the cube has no `name`, `cube_label` falls back to `src_path` and the messages read `cubes/x.yml > cubes/x.yml > measures`.

core/wren/src/wren/cube_cli.py
```diff
@@ -150,17 +150,75 @@ def _load_manifest_dict(mdl: str | None) -> dict:
...
+                        f"Error: malformed cubes in mdl.json ({name!r} {key}[{j}] is not an object).\n"
+                        "  Hint: re-run `wren context build`.",
+                        err=True,
+                    )
+                    raise typer.Exit(1)
+                if not isinstance(item.get("name"), str):
+                    name = cube.get("name", f"entry {i}")
+                    typer.echo(
+                        f"Error: malformed cubes in mdl.json ({name!r} {key}[{j}].name is not a string).\n"
+                        "  Hint: re-run `wren context build`.",
```
goldmedal: The hint asserts the build artifact is stale, but as the previous comment shows the usual cause is the cube YAML — and re-running the build reproduces the error verbatim. Pointing at the source keeps the message actionable in both cases:

```suggestion
                        f"Error: malformed cubes in mdl.json ({name!r} {key}[{j}].name is not a string).\n"
                        "  Hint: fix the cube definition in cubes/*/metadata.yml, then re-run `wren context build`.",
```

(Same wording applies to the three sibling messages in this helper.)

core/wren/src/wren/context.py
```diff
@@ -660,17 +660,67 @@ def load_cubes(project_path: Path) -> list[dict]:
...
     cubes_dir = project_path / "cubes"
     if not cubes_dir.is_dir():
         return []
     cubes = []
     for f in sorted(cubes_dir.glob("*.yml")):
-        data = yaml.safe_load(f.read_text(encoding="utf-8"))
+        try:
+            data = yaml.safe_load(f.read_text(encoding="utf-8"))
+        except yaml.YAMLError:
+            continue
```
goldmedal, line 738: This swallow is worth a second look — the PR description calls it "the same pattern as views / relationships in #2604 / #2613", but those loaders only filter non-mapping *entries*; neither catches `YAMLError`, so a parse error still propagates. Measured difference on this branch:

| | this PR | `main` |
| --- | --- | --- |
| `load_cubes()` on a cube with a YAML syntax error | `[]`, silent | raises `YAMLError` |
| `wren context validate` | `[ERROR] cubes/om/metadata.yml: invalid YAML: …` ✅ | raises |
| `wren context build --no-validate` | succeeds, cube silently missing | fails |
| MCP `list_cubes` / `describe_cube` | cube silently missing | tool error |

Validation reporting it is the clear win, and `build` validates by default, so the main path is covered. The gap is the paths that never call `validate_project`: MCP's two cube tools call `load_cubes` directly, so an agent sees the cube vanish with no signal — which is the same "silent drop hides a producer bug" argument `_require_cubes_list` makes for `mdl.json`. Also note the build summary only prints models and views, so a dropped cube leaves no trace there either.

Not blocking, but could the loaders record the skipped file (a `logging` warning, or returning the failures for callers to surface) rather than dropping it entirely?

core/wren/src/wren/cube_cli.py
```diff
@@ -150,17 +150,75 @@ def _load_manifest_dict(mdl: str | None) -> dict:
     if not isinstance(manifest, dict):
         typer.echo("Error: MDL JSON must be an object.", err=True)
         raise typer.Exit(1)
+    _require_cubes_list(manifest)
     return manifest
 
 
+def _require_cubes_list(manifest: dict) -> list:
```
goldmedal, line 157: Nit: the return value is discarded by the only caller, and both `return` statements are dead. Either annotate `-> None` and drop them, or have `_load_manifest_dict` use it (`manifest["cubes"] = _require_cubes_list(manifest)`) — the latter would also let this helper normalise instead of only checking.

goldmedal (changes requested):
Reviewed on `b50481e0` in a local worktree, with the CLI actually exercised end to end rather than read.

**The direction is right** — re-reading the raw YAML inside `validate_project` is the correct place for this, both layouts (v1 flat files, v2 per-cube dirs) are covered, and both CodeRabbit points are addressed.

**But the crash the PR set out to remove is still reachable, on the most ordinary path there is:** a cube YAML with an empty `dimensions:` key. `validate` reports 0 errors, `build` succeeds, and then `wren cube list` dies with a bare `TypeError` while `wren cube query` fails inside wren-core (`invalid type: null, expected a sequence`). `null` slips through both new guards, on the producer side and the reader side. Details and a one-line fix are inline. This is pre-existing on `main`, not a regression from this PR — but it is the same bug class this PR exists to close, so it seems worth closing properly.

Second, a nameless measure passes `validate`, builds, and then hard-fails with `Hint: re-run wren context build`, which reproduces the error identically — an inescapable loop. Suggested fix: catch it in validation, where the actual fault lives.

Verification I ran on the branch:

- `uvx ruff format --check src/` and `uvx ruff check src/` — both clean
- `pytest tests/unit/ --ignore=tests/unit/test_memory.py --ignore=tests/unit/test_mcp_server.py` — 1114 passed, 2 skipped
- `pytest tests/unit/test_mcp_server.py` — 16 passed
- `git merge-tree` against current `main` — clean (branch is 3 `context.py` commits behind, no conflicts)
- `ruff format --check` on the two touched test files wants 2 reformats; CI only gates `src/`, so this is cosmetic

Bartok9:
@goldmedal Thanks — both blocking paths were real; pushed a fix on this branch.

1. **Producer:** `_normalise_cube_member_lists` now treats present-but-null member keys (`dimensions:`) as `[]`, so `build` no longer writes null containers into `mdl.json`.
2. **Reader:** `_require_cubes_list` no longer `continue`s on null — fails loud with the same malformed-MDL path (covers `list` / `describe`).
3. **Validate:** nameless members get `must have a string 'name'` in `_report_malformed_cube_members` so the rebuild loop cannot form.
4. **Hints:** member-level CLI errors point at `cubes/*/metadata.yml`, then rebuild.
5. **cube_label:** falls back to `"cube"` instead of repeating `src_path`.

Non-blocking YAMLError-in-loader note: left as-is for this pass (validate still reports; default build validates). Happy to add a `logging.warning` on skip in a follow-up if you want signal on MCP/`--no-validate` paths without changing return type.

Tests: null-normalise, nameless-measure validate, null-dimensions `cube list`.

Bartok9:
Thanks @goldmedal — this is a clean review of the producer/consumer split, and I agree with the blocking pair.

**Producer (`_normalise_cube_member_lists`):** empty YAML keys (`measures:` / `dimensions:` / `time_dimensions:`) becoming `null` in `mdl.json` is the real bug. `[]` is what wren-core accepts; waving `None` through is how `wren cube list` TypeErrors and `wren cube query` / MCP `query_cube` reject the manifest. Normalising `None` → `[]` at load time is the right fix, and a regression that goes through `load_cubes` / `build_manifest` (not only the helper in isolation) is the right test.

**Reader (`list_cubes`):** with that producer fix, failing loud on leftover `null` is truthful. The current `cube.get(key, [])` + skip-`None` path still lets a stale artifact crash in `", ".join(...)`. Either hard-fail with a hint, or tolerate with `or []` like `mcp_server.py` — but the producer change is the one that must land; the reader only decides how old `mdl.json` behaves.

**Name / hint:** a member without a string `name` should fail in `validate_project` (source YAML), not only in `list_cubes` with a “re-run build” hint that reproduces the same error. Pointing the mdl.json messages at `cubes/*/metadata.yml` then `wren context build` is more actionable.

**YAMLError swallow:** not blocking. Views/relationships don’t catch parse errors either; validate-by-default covers the main CLI path. The gap is MCP `list_cubes` / `describe_cube` calling `load_cubes` directly, so a syntax error can look like a missing cube. A warning (or returned failures) instead of a silent drop would close that without changing the happy path.

**Nit on `_require_cubes_list`:** unused return / dead `return`s — annotate `-> None` or actually assign into the manifest. Non-blocking.

I won’t touch the branch. Happy to re-review once those producer + reader bits are in.

Bartok9:
@goldmedal Quick correction on my earlier note at 12:17 UTC — that read was against your review on `b50481e0` and I failed to re-check HEAD.

**Blocking items from your review are already on this branch** (commits `b56f081` → `09250ad`):

1. **Producer:** `_normalise_cube_member_lists` maps present-but-null `measures` / `dimensions` / `time_dimensions` → `[]`
2. **Reader:** `_require_cubes_list` fails loud on null member lists (no silent `continue`)
3. **Validate:** nameless / non-string member `name` → `ValidationError` (`must have a string 'name'`)
4. **Hints:** point at `cubes/*/metadata.yml`, then `wren context build`
5. **Tests:** load_cubes null keys + validate nameless measure + null-dimensions `cube list` + v2 invalid YAML

CI is green on `09250ad` (lint, unit, mcp, memory, mysql, postgres, ui).

Non-blocking left as discussed: YAMLError swallow in loader (validate still reports; default build validates). Happy to add a `logging.warning` on skip in a follow-up if you want MCP/`--no-validate` signal.

Could you re-review on current HEAD when you have a moment? Thanks again for the end-to-end CLI exercise — that null empty-key path was the real gap.

Bartok9:
@goldmedal Re-requesting review — the blocking producer/reader/validate/hint items from your `b50481e0` review are on HEAD (`09250ad`). CI is green. Remaining note on YAMLError-in-loader left non-blocking as discussed (validate still reports; default build validates).

core/wren/src/wren/cube_cli.py
```diff
@@ -150,17 +150,87 @@ def _load_manifest_dict(mdl: str | None) -> dict:
...
+    """mdl.json is a build artifact: fail loud on malformed cubes.
+
+    Project YAML is normalised in load_cubes + reported by validate_project.
+    The CLI reads target/mdl.json from `wren context build`. Skipping bad
+    rows here would hide a producer bug and look like an empty listing.
+    """
+    if "cubes" not in manifest or manifest["cubes"] is None:
```
goldmedal: **[P2] Treating `cubes: null` as an empty list hides a malformed manifest**

This condition handles a missing `cubes` field and an explicitly null field identically, but the latter is an invalid manifest. Reproduced with `{"catalog":"c","schema":"s","cubes":null}`:

- `wren cube list` exits 0 and prints `No cubes defined.`
- `wren cube describe orders` reports `Cube 'orders' not found.`
- `wren cube query ...` fails with `invalid type: null, expected a sequence`

The serde default for Rust's `Vec<Cube>` applies only when the field is absent; an explicit `null` is still rejected. Please split these cases so only a missing field returns an empty list, while `null` follows the existing malformed-cubes loud-failure path, and add a regression test.

Bartok9:
@goldmedal Thanks — P2 on `cubes: null` was real and distinct from the missing-key case.

Pushed on HEAD:
- **`_require_cubes_list`:** only a *missing* `cubes` key returns `[]`. Explicit `null` fails loud with the same malformed-cubes + rebuild hint (aligned with Rust `Vec<Cube>` rejecting null while defaulting absent → empty).
- **Tests:** `test_cube_list_null_cubes_fails_loud` + `test_cube_list_missing_cubes_key_is_empty`.

Local: `pytest tests/unit/test_cube_cli.py -k 'null_cubes or missing_cubes or malformed_cubes_scalar'` — 3 passed.

Bartok9:
@goldmedal Thanks — agreed on the missing-vs-null split.

Addressed in b63f0dcc: `_require_cubes_list` now treats only a missing `cubes` key as `[]` (serde default), and fails loud when `cubes` is explicitly `null` (same malformed path as a non-list). Regression coverage for both cases is in `test_cube_cli.py`.

CI is green on the earlier checks; unit/DB jobs still rolling on that commit.

goldmedal (approved):

goldmedal:
Thanks @Bartok9 👍

PR #2676, opened 2026-08-17
fix(skills): wrap generate-mdl SKILL.md's relationships.yml example under the required key

AmirF194:
## Root cause

`generate-mdl`'s SKILL.md, Step 3 — Write relationships, showed:

```yaml
# relationships.yml
- name: orders_customers
  models:
    - orders
    - customers
  join_type: many_to_one
  condition: "orders.customer_id = customers.customer_id"
```

a bare top-level YAML list. `load_relationships()` (`core/wren/src/wren/context.py`)
only reads a top-level `relationships:` mapping key:

```python
data = yaml.safe_load(rel_file.read_text(encoding="utf-8")) or {}
rels = data.get("relationships") if isinstance(data, dict) else None
if not isinstance(rels, list):
    return []
```

so an agent following the doc verbatim ends up with a `relationships.yml`
that loads as zero relationships. `validate_project()` does report a clear
`relationships.yml must be a mapping with a 'relationships' key, got list`
error for this exact shape (and by default `wren build` runs validation
first and aborts on it), so the failure today is loud rather than silent —
but the doc's own example has never actually matched what the loader
expects.

## Fix

Wrap the example under the `relationships:` key, matching what
`load_relationships`/`validate_project` require:

```yaml
# relationships.yml
relationships:
  - name: orders_customers
    models:
      - orders
      - customers
    join_type: many_to_one
    condition: "orders.customer_id = customers.customer_id"
```

## Verification

- New regression test (`test_generate_mdl_skill_step3_example_round_trips`
  in `tests/unit/test_context.py`) reads the skill's own served content via
  `wren.skills_delivery.get_skill("generate-mdl")`, extracts the Step 2/Step 3
  YAML blocks, and feeds them through `load_relationships`/`validate_project`.
  Fails on unmodified `main` (`load_relationships` returns 0 relationships
  instead of 1); passes with this fix.
- Full `tests/unit/` (excl. `test_memory.py`/`test_mcp_server.py`, matching
  this repo's `test-unit` CI job): 1146 passed, 2 skipped, no regressions.
- `ruff format --check src/` / `ruff check src/` clean.
- Not verified: an end-to-end `wren context init && wren build` run against a
  live database: this is a pure-doc/test change and doesn't touch runtime code
  paths beyond what the new test already exercises directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the MDL generation workflow’s relationships YAML structure to ensure relationship definitions are recognized properly.

* **Tests**
  * Added regression coverage to verify relationship loading and project validation for generated v2 projects.

### DIFF
--- a/core/wren/src/wren/skills_content/generate-mdl/SKILL.md
+++ b/core/wren/src/wren/skills_content/generate-mdl/SKILL.md
@@ -213,12 +213,13 @@ From foreign key constraints discovered in Phase 2:
 
 ```yaml
 # relationships.yml
-- name: orders_customers
-  models:
-    - orders
-    - customers
-  join_type: many_to_one
-  condition: "orders.customer_id = customers.customer_id"
+relationships:
+  - name: orders_customers
+    models:
+      - orders
+      - customers
+    join_type: many_to_one
+    condition: "orders.customer_id = customers.customer_id"
 ```
 
 Join type mapping:
--- a/core/wren/tests/unit/test_context.py
+++ b/core/wren/tests/unit/test_context.py
@@ -2011,6 +2011,47 @@ def test_validate_project_reports_relationships_bare_root(tmp_path: Path) -> Non
     )
 
 
+def _extract_fenced_yaml(markdown: str, heading: str) -> str:
+    """Pull the first ```yaml fenced block under a markdown heading."""
+    after_heading = markdown[markdown.index(heading) :]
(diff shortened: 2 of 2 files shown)


AmirF194:
Checking in, no pressure. Let me know if anything here needs rework or more context.

goldmedal (approved):
Thanks @AmirF194, nice catch 👍

AmirF194:
Thanks for the quick merge!

PR #2677, opened 2026-08-17
feat(core): support ordered cube queries

byjoh:
## What changed

- add strict CubeQuery orderBy parsing and public types
- validate ordered members are selected and unique
- generate native ORDER BY using selected-output ordinals
- preserve legacy time ordering when orderBy is absent or empty
- add Rust and Python binding regression coverage

## Why

Consumers need deterministic native ranking without SQL wrapper rewrites. Reporting-specific limit policy remains outside general Wren Core.

## Validation

- Rust focused ordering/public API tests: 8 passed
- Python cube binding tests: 13 passed
- cargo fmt --check
- cargo clippy --all-targets --all-features -- -D warnings
- Ruff format/check
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Added cube query sorting by selected measures and dimensions in ascending or descending order.
  * Generated SQL now uses selected column positions for explicit ordering.
  * Preserved default ordering by the first time dimension when no custom sorting is provided.

* **Bug Fixes**
  * Added validation to reject unknown, duplicate, or improperly formatted sort criteria.
  * Queries without selectable members can no longer specify ordering.

### DIFF
--- a/core/wren-core-py/tests/test_cube.py
+++ b/core/wren-core-py/tests/test_cube.py
@@ -3,7 +3,6 @@
 import json
 
 import pytest
-
 from wren_core import cube_query_to_sql
 
 MANIFEST = json.dumps(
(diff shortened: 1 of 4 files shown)


goldmedal:
Thanks @byjoh, overall looks good but some comment.

I suggest adding `Default` derive to `CubeQuery` to avoid a breaking change for downstream:

```diff
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct CubeQuery {
```

Every field already satisfies `Default`, so the derive is free. Callers — including this PR's new `core/wren-core/core/tests/cube_public_api.rs` — then become:

```rust
let query = CubeQuery {
    cube: "OrdersCube".to_string(),
    measures: vec!["revenue".to_string()],
    dimensions: vec!["status".to_string()],
    order_by: vec![CubeOrderBy {
        member: "revenue".to_string(),
        direction: SortDirection::Desc,
    }],
    ..Default::default()
};
```

This PR focuses on the core change but we should also deliver this change to each SDK. I filed the follow-up issue to track them:


- #2700 — the WASM TypeScript SDK's `CubeQueryInput` has no `orderBy`, so TS callers need a cast (it works at runtime; the binding is JSON-string based and passes it through).
- #2701 — the `wren` CLI and the MCP `query_cube` tool cannot reach ordering at all, which is why agent-driven "top N" is currently unordered. This PR is the enabling half of that fix.

goldmedal (approved):
Thanks @byjoh 👍.  I'll merge this PR first so the follow-up issue can keep going.

PR #2679, opened 2026-08-18
fix(wren): accept only read-only SELECT statements in the query path

goldmedal:
## Problem

The CLI executed DDL and DML against the connected database. Reproduced on MySQL 8 with a one-model manifest, reading the result from a separate `mysql` client rather than through the engine's own connection:

```
$ wren query -m mdl.json --connection-file conn.json -s "DROP TABLE t"
# no error

mysql> SHOW TABLES LIKE 't';
Empty set
```

`TRUNCATE TABLE t` behaves the same way (rows 2 → 0).

Nothing in the path distinguished a read from a write:

- `validate_sql_policy` only checked *which tables* may be named and *which functions* are denied. Both checks are gated on `strict_mode`, which is **off by default**, and `_plan` only called the policy layer at all when `strict_mode or denied_functions` was set.
- `_plan` then handed the planned SQL to the connector, which executes it verbatim. The MySQL, Redshift and Canner connectors run with `autocommit=True`, so the write is durable.
- On PostgreSQL the statement still executes but is rolled back, because the connector never commits. That is a psycopg default, not a guard — no code prevents the write, and reading back on the engine's own connection shows the mutation before it disappears.

## Change

A read-only statement check, **always on and independent of strict mode** — strict mode governs *which tables* may be named, not *what may be done to them*.

Three design points, each forced by something measured rather than chosen for style:

**Allow-list, not deny-list.** sqlglot folds every construct it cannot model into a single `exp.Command` node, so `EXPLAIN`, `SHOW` and `DO $$ ... $$` are indistinguishable there. Anything not positively recognised as a read-only query has to be rejected.

**Root check plus whole-tree scan.** The root type alone is not sufficient — both of these root at `exp.Select` while mutating:

```sql
WITH x AS (DELETE FROM t RETURNING id) SELECT * FROM x   -- PostgreSQL data-modifying CTE
SELECT * INTO t2 FROM t1                                 -- T-SQL, creates t2
```

sqlglot's `exp.DDL` / `exp.DML` markers are a second net rather than the mechanism: `Drop`, `TruncateTable`, `Alter` and `Grant` carry neither.

**Planning is checked too, not just input.** `CTERewriter` inlines MDL view statements and model `ref_sql`, so `SELECT * FROM v` — which the input check passes — becomes `WITH v AS (DROP TABLE public.t) SELECT * FROM v` when the view statement is DDL, and that output is what reaches the connector. `validate_planned_sql` covers it at the end of `_plan`, which is enough for `query`, `dry_run` and `dry_plan` since all three route through there.

Unlike the input check, the output check **fails open on parse errors**: the caller does not author the planned string, so rejecting a valid query because sqlglot cannot re-parse the transpiler's dialect-specific output would be a regression. A mutating node in output that *does* parse is still rejected.

Failing closed on input costs nothing — SQL sqlglot cannot parse is already rejected downstream with `INVALID_SQL`, with strict mode on or off. Verified before relying on it.

Rejections raise the new `ErrorCode.BLOCKED_STATEMENT` with phase `SQL_POLICY_CHECK`.

## Verification

Same fixtures before and after, every "after" figure read from an independent `mysql` client:

| | result |
|---|---|
| `SELECT a FROM t` (with the change) | returns `1`, `2` |
| `DROP TABLE t` — **before** | no error, table gone |
| `DROP TABLE t` — **after** | `BLOCKED_STATEMENT` / `SQL_POLICY_CHECK`, table intact, 2 rows |

Also confirmed after the change: a view whose statement is `DROP TABLE public.customer` plans to `WITH evil AS (DROP TABLE public.customer) SELECT * FROM evil`, and that planned SQL is now rejected.

The new tests were **mutation-checked rather than just run** — disabling `validate_read_only_ast` fails 17 of them, so they are load-bearing.

Both CI jobs run locally:

- `ruff format --check src/` + `ruff check src/` — clean
- `pytest tests/unit/ --ignore=tests/unit/test_memory.py --ignore=tests/unit/test_mcp_server.py` — **1172 passed, 1 skipped**

The wider `tests/` tree has 15 failures (Snowflake connector, memory embeddings) and 24 collection errors from missing optional connector extras. All of them reproduce identically on the base commit with these changes stashed — checked by diffing the failure lists, not assumed.

## Notes

- `VALUES (1), (2)` is allowed: it roots at `exp.Values` and reads no table, so rejecting it would be a false positive with nothing gained.
- `DESCRIBE` / `DESC` and `SHOW` are rejected even though they are read-only. Say if you would rather they were allowed — it is one entry in the allow-list.
- The check classifies statement **shape**, not effects: a pure `SELECT` calling a side-effecting function (`pg_terminate_backend`, `setval`, …) still passes. `denied_functions` is the control for that surface. Out of scope here, but worth knowing the boundary.
- `context.py` already dry-plans every view statement when validating a project, so mutating view statements are now reported there as a side effect of this change.

## Summary by CodeRabbit

* **Security Enhancements**
  * Enforced read-only SQL validation for all queries.
  * Blocked write operations, data-modifying statements, session controls, DDL, and multiple statements.
  * Added validation of generated SQL to prevent injected mutations.
  * Added a dedicated blocked-statement error code.
* **Bug Fixes**
  * Ensured statement safety checks run before stricter table and function policy checks.
  * Improved protection against unsafe SQL introduced during query planning.


goldmedal:
Good catch — fixed in 9096b2cd. `exp.Lock` is now in `_FORBIDDEN_NODES`, and `SELECT ... FOR UPDATE` / `FOR SHARE` moved from the allowed cases to the blocked ones.

One correction on the reasoning, since I measured it before changing anything: through the semantic layer the row mark does **not** take a blocking lock. The planner wraps the query, so `FOR UPDATE` lands on a generated CTE rather than the base table, and an independent writer with `lock_timeout='1s'` updates the same row without waiting — identically after a plain `SELECT` and after the locking read:

```
after plain SELECT             -> outside writer: SUCCEEDED
after SELECT ... FOR UPDATE    -> outside writer: SUCCEEDED
```

So the availability harm as described doesn't materialise on this path. I'm taking the change anyway, on two grounds that don't depend on that result:

- Asking for write-intent locks isn't a read-only request, whatever the planner happens to do with it.
- Its harmlessness is an accident of the current CTE shape. The connector holds a long-lived connection it never commits, so a row mark that *did* reach a base table would be held for the life of that connection — and leaning on "the rewriter happens to wrap it" is exactly the kind of incidental protection this PR exists to replace.

Worth noting for the record: my original justification for allowing it was wrong in a different way. I'd written that autocommit releases the lock immediately, but the PostgreSQL connector never commits — the backend sits `idle in transaction` after any query. The code comment now records the measurement rather than a conclusion, so the next reader can see why the node is on the list.

Lint clean; `pytest tests/unit/` (with CI's ignores) 1173 passed / 1 skipped.

douenergy (approved):

PR #2680, opened 2026-08-18
fix(wren-ai-service): resolve all open dependabot alerts (39 high, 12 medium)

yichieh-lu:
Closes every open dependabot alert on `wren-ai-service/poetry.lock` — 39 high and 12 medium, all of which sit on 10 packages. The last commit to touch this lock was e42b8d05 (2026-05-04), so the backlog is ~3 months of advisories.

## What changes for a user

Nothing in the service's behaviour, with one exception: litellm 1.95 logs `register_model: model=... not in built-in cost map` whenever a `Router` is constructed. That string does not exist in 1.83.14, so it is new log noise, not a behaviour change.

## Versions

Declared deps:

| package | from | to | advisory |
|---|---|---|---|
| sqlparse | 0.5.5 | 0.6.0 | lexer/grouping/comment quadratic DoS |
| aiohttp | 3.13.4 | 3.14.3 | OOB heap read in C HTTP response parser |
| litellm | 1.83.14 | 1.95.1 | MCP auth bypass via OAuth2 passthrough |
| nltk | 3.9.3 | 3.10.3 | path traversal, SSRF filter bypass, ReDoS |
| gitpython | 3.1.45 | 3.1.59 | git option injection / arbitrary file write |
| pillow | 12.1.1 | 12.3.0 | heap OOB writes, decompression-bomb DoS |

Newly pinned transitive deps, each placed in the group where its consumer lives so the pin adds nothing to the runtime image:

| package | from | to | group | consumers |
|---|---|---|---|---|
| pyasn1 | 0.6.2 | 0.6.4 | main | google-auth, rsa, pyasn1-modules |
| h2 | 4.3.0 | 4.4.1 | main | httpx[http2] via qdrant-client |
| tornado | 6.5.2 | 6.5.8 | dev | streamlit, ipykernel, jupyter-client |
| setuptools | 80.9.0 | 84.0.0 | dev | deepeval, locust |

Dependabot reads `poetry.lock`, which records one version per package regardless of group, so the dev-group pins close their alerts while `poetry install --without dev,eval,test` (`docker/Dockerfile:15`) keeps tornado and setuptools out of the image, as before this change. Confirmed via the `groups` metadata in the lock.

The lock moves 12 packages in total: the 10 above plus `hpack` (h2's dep) and `defusedxml` (new nltk dep). Nothing else changed.

## Verification

`pytest --ignore tests/pytest/test_usecases.py` against a throwaway Qdrant v1.15.0 and a live `OPENAI_API_KEY`: **95 passed, 4 skipped, 0 failed**. The 4 skips are pre-existing `@pytest.mark.skip` in `test_ask.py`.

Because the suite does not reach every upgraded package, each was checked against the surface this repo actually uses:

- **litellm** — the suite covers neither streaming nor the `Router` fallback, so both were driven against the API directly. Streaming produced chunks that `connect_chunks` joined correctly; a `Router` with `mock_testing_fallbacks` served from the fallback model; the normal `Router` path served from the primary. The response fields `build_message`/`build_chunk` read (`model`, `index`, `finish_reason`, `usage`) are all still populated, and `aembedding`'s `response.data[i]["embedding"]` subscript still works.
- **aiohttp** — the repo uses only `ClientSession`, `ClientTimeout`, `ClientError` and `aiohttp.request`; 3.14's removals are server-side (`web.BaseRequest.host` getfqdn fallback), plus the Python 3.9 drop and header control-char tightening (the only custom header here is the static `x-wren-fallback_disable`). `WrenUI` and `WrenIbis` were round-tripped against an in-process server: query args and JSON bodies unchanged, the header still arrives, a slow response still raises `asyncio.TimeoutError` and returns the same tuple, and a refused connection still raises an `aiohttp.ClientError` subclass so the `backoff` decorators still retry.
- **sqlparse** — 21 statements × 3 call shapes diffed between 0.5.5 and 0.6.0. The only difference is `ALTER TABLE ... ROW_FORMAT=`, which 0.6.0 now recognises as a keyword; this service formats SELECT queries. `split()` is not used, so the splitter rewrite is not on any path.
- **nltk** — `word_tokenize` output is identical on 5 statements. Only `word_tokenize` and `download("punkt_tab")` are used, both in `eval/metrics/spider`.
- **gitpython** — these CVEs are fixed by *rejecting* unsafe git options, which is itself a behaviour change, but the only usage is `Repo(search_parent_directories=True)` with `active_branch` and `head.commit` in `eval/prediction.py`, which passes no options.
- **setuptools 84 removes `pkg_resources`.** Every reference to it under site-packages sits behind a `try/except ImportError`; litellm's is the only one on a runtime path and prefers `importlib.resources`, which works on 3.12. litellm, tiktoken, pytz, numpy, pandas, pyarrow, wrapt, werkzeug, matplotlib, sentry_sdk, locust, streamlit and dspy all import cleanly.
- **pillow / tornado / pyasn1 / h2** have zero direct imports anywhere in the repo. litellm's only PIL reference is a function-local guarded import in `_load_image_from_url`, used for Bedrock/Vertex image conversion. `qdrant_client` defaults `http2=False` and the repo never overrides it, so h2 is installed but not executed.

### Not covered

- Only the OpenAI provider was exercised. litellm 1.83 → 1.95 spans 12 minor releases and most of its churn is in per-provider parameter translation, so Anthropic / Bedrock / Vertex / Azure / Ollama paths are untested here.
- `tests/pytest/test_usecases.py` was not run; `just test` ignores it and it needs a full stack.
- The lock was generated with poetry 2.3.2 on macOS. CI installs it with poetry 1.8.3 on ubuntu-22.04, which is the main thing this PR's CI run adds.

### Pre-existing, left alone

`tests/pytest/eval/test_metrics.py` still fails to collect: `eval/metrics/accuracy.py` imports `TestResult` from `deepeval.evaluate`, which has lived in `deepeval.evaluate.types` since deepeval 3.x. deepeval is 3.8.8 both before and after this change, so the break predates it (introduced by e7db92e4).

### DIFF
--- a/wren-ai-service/poetry.lock
+++ b/wren-ai-service/poetry.lock
@@ -7,6 +7,7 @@ description = "Simple DNS resolver for asyncio"
 optional = false
 python-versions = ">=3.9"
 groups = ["main"]
+markers = "sys_platform != \"android\" and sys_platform != \"ios\""
 files = [
     {file = "aiodns-3.5.0-py3-none-any.whl", hash = "sha256:6d0404f7d5215849233f6ee44854f2bb2481adf71b336b2279016ea5990ca5c5"},
     {file = "aiodns-3.5.0.tar.gz", hash = "sha256:11264edbab51896ecf546c18eb0dd56dff0428c6aa6d2cd87e643e07300eb310"},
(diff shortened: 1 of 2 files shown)


yichieh-lu:
## CI note: the `pytest` failure is pre-existing and unrelated to this change

The `pytest` job fails at collection, not on a test:

```
eval/metrics/accuracy.py:7: in <module>
    from deepeval.evaluate import TestResult
E   ImportError: cannot import name 'TestResult' from 'deepeval.evaluate'
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
```

Evidence that this predates the branch:

- That import line is unchanged at the base commit (`origin/legacy/v1`); this PR touches only `pyproject.toml` and `poetry.lock`.
- `deepeval` resolves to 3.8.8 both before and after this change, so nothing here moved it. `TestResult` has lived at `deepeval.evaluate.types` since deepeval 3.x, i.e. since e7db92e4 bumped deepeval to 3.0.
- The last `ai-service-test` run that was not skipped before this one was on 2026-05-04 against `main`, and it failed the same way.

Because it is a *collection* error, pytest aborts before running anything — so **none of the 99 tests executed**, on this PR or on any run since May.

### What CI did verify

The `Install the project dependencies` step passed, i.e. `poetry install --without eval` resolves and installs this lock cleanly with **poetry 1.8.3 on ubuntu-22.04**. That was the main gap in the local verification, which used poetry 2.3.2 on macOS. What CI could not add is a clean-room run of the suite itself.

The local run reported in the PR description (95 passed, 4 skipped, 0 failed, plus the per-package checks) used `--ignore tests/pytest/eval`, which is why it did not hit this.

### Fixing it is a separate change

`eval/metrics` has two independent breaks, not one:

1. `from deepeval.evaluate import TestResult` → `from deepeval.evaluate.types import TestResult`
2. with that fixed, the next error is `ModuleNotFoundError: No module named 'deprecated'` — `eval/metrics/accuracy.py:10` imports `Deprecated`, which appears in neither `pyproject.toml` nor `poetry.lock`. It was presumably a transitive dependency lost in the same deepeval 3.0 bump.

With both applied locally, collection succeeds and yields 105 tests. That was deliberately left out of this PR: it needs a new dependency added to the lock, which has no place in a security bump, and those 6 `test_metrics.py` tests have not executed in months — once collection succeeds they will actually run against wren-engine and OpenAI, so the collection error may simply become test failures. That deserves its own PR rather than riding along here.

cyyeh (approved):

PR #2681, opened 2026-08-18
fix(wren): report a validation error for non-scalar model/view names

### LINKED ISSUE #2673: validate_project crashes with TypeError on a non-scalar model or view name
## Summary

`validate_project()` raises `TypeError: unhashable type: 'list'` instead of reporting a validation error when a model or view `name` parses as a list or a mapping. `wren context validate` shows a raw traceback.

A validation command should report malformed input, not crash on it — the same principle applied in #2604 and #2616.

## Where

`core/wren/src/wren/context.py:972` on `main` (`6db8bed5`):

```python
if name in model_names:      # <- TypeError when name is a list/dict
    errors.append(
        ValidationError("error", src_path, f"duplicate model name '{name}'")
    )
model_names.add(name)
```

`name` comes from the project's own YAML via `load_models()` and is only checked for truthiness (`if not name: ... continue`), so any non-scalar value reaches the set operations. Views have the same shape with `view_names`, and crash the same way.

## Reproduction

```yaml
# models/orders/metadata.yml
name: [a, b]
table_reference: {table: orders}
columns:
  - {name: id, type: INTEGER}
```

```console
$ wren context validate
...
│ ❱  972 │   │   if name in model_names:                                       │
TypeError: unhashable type: 'list'
```

Exit code is 1, but the user gets a traceback and no indication of which file or field is wrong. A view whose `metadata.yml` contains `name: [a, b]` reproduces it via `view_names`.

`name: 3` does not crash — an int is hashable — so the trigger is specifically a non-scalar (`list` / `dict`) name.

## Expected

An error such as `models/orders/metadata.yml: model name must be a string, got list`, reported alongside the other validation errors, with the model skipped afterwards.

## Notes

The check belongs next to the existing `if not name` guard so both the models and the views loop are covered. Note that #2614, if merged first, adds a second site that hits this earlier (`raw_model_columns[mname]`), so a fix should cover the raw-columns side table too.

AmirF194:
## Root cause

`validate_project()` (`core/wren/src/wren/context.py`) only checks a model or
view `name` for truthiness before using it as a `set` member for duplicate
detection: `if name in model_names:` (models, line 972) and `if name in
view_names or name in model_names:` (views, line 1223). `name` comes straight
from the project's own YAML via `load_models()`/`load_views()` with no type
check, so a hand-edited `name: [a, b]` or `name: {x: 1}` (both unhashable)
reaches the set membership test and raises an unhandled `TypeError`, crashing
`wren context validate`/`wren context build` with a raw traceback instead of
reporting a validation error.

## Fix

Add a type guard next to the existing `if not name` check in both loops: a
`list`/`dict` name is now reported as a clean `ValidationError` and the entry
is skipped, mirroring the "must be a mapping, got X" pattern already used for
malformed relationship/column entries in this same file. A hashable
non-string name (e.g. `name: 3`) is unaffected, matching current behavior.

## Verification

- Docker-reproduced on unmodified HEAD (`7830cc7`, `python:3.11-slim`,
  `wren.context.__file__` provenance-checked): `TypeError: unhashable type:
  'list'` at `context.py:972`, matching the issue.
- 5 new tests in `test_context.py` (list/dict for both models and views, plus
  an int-name control) fail on `main` with the same `TypeError` and pass on
  this branch; the full `tests/unit/` suite (excluding `test_memory.py`/
  `test_mcp_server.py`, matching CI's `test-unit` job) is 1150 passed, 2
  skipped, 0 failed. `ruff format --check` / `ruff check` on `src/` clean.
- Not verified: line coverage via `pytest-cov`/`coverage.py` — both panic on
  this repo's `wren_core` PyO3 extension (`env_logger::init should not be
  called after logger initialized`) independent of this diff. Checked by hand
  instead: every added line has a test that enters the branch (list/dict) and
  a test that skips it (string/int name, the existing duplicate-name test).

Fixes #2673

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for model and view names provided as lists or dictionaries.
  * Validation now reports clear scalar-value errors before proceeding with duplicate-name and other checks.
  * Preserved support for valid numeric model names.

### DIFF
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -968,6 +968,15 @@ def validate_project(project_path: Path) -> list[ValidationError]:
         if not name:
             errors.append(ValidationError("error", src_path, "model missing 'name'"))
             continue
+        if isinstance(name, (list, dict)):
+            errors.append(
+                ValidationError(
+                    "error",
+                    src_path,
+                    f"model 'name' must be a scalar value, got {type(name).__name__}",
+                )
+            )
+            continue
 
         if name in model_names:
             errors.append(
@@ -1217,6 +1226,15 @@ def validate_project(project_path: Path) -> list[ValidationError]:
             errors.append(
                 ValidationError(
                     "error", f"views/{src_dir}/metadata.yml", "view missing 'name'"
+                )
+            )
+            continue
+        if isinstance(name, (list, dict)):
+            errors.append(
+                ValidationError(
+                    "error",
+                    f"views/{src_dir}/metadata.yml",
+                    f"view 'name' must be a scalar value, got {type(name).__name__}",
                 )
             )
             continue
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -965,9 +965,6 @@ def validate_project(project_path: Path) -> list[ValidationError]:
         src = model.get("_source_dir", f"models[{i}]")
         src_path = f"models/{src}/metadata.yml"
         name = model.get("name")
-        if not name:
-            errors.append(ValidationError("error", src_path, "model missing 'name'"))
-            continue
         if isinstance(name, (list, dict)):
             errors.append(
                 ValidationError(
@@ -976,6 +973,9 @@ def validate_project(project_path: Path) -> list[ValidationError]:
                     f"model 'name' must be a scalar value, got {type(name).__name__}",
                 )
             )
+            continue
+        if not name:
+            errors.append(ValidationError("error", src_path, "model missing 'name'"))
             continue
 
         if name in model_names:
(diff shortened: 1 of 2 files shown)


AmirF194:
CodeRabbit's finding was correct: an empty list/dict is falsy, so the pre-existing 'if not name' check ran first and reported 'missing name' instead of the scalar-type error the new guard exists for. Moved the isinstance check ahead of the truthiness check in both loops, added regression tests for the empty-list/empty-dict case on both, verified they fail against the old ordering and pass against this one (046aafe6).

AmirF194:
No rush, just checking in on this one. CI is green and CodeRabbit's one finding has been addressed (moved the type check ahead of the truthiness check, added regression tests). Happy to rebase or adjust scope if useful.

goldmedal (approved):
Thanks @AmirF194 👍

AmirF194:
Thanks for merging it.

PR #2683, opened 2026-08-19
fix(wren): connect with autocommit so a failed statement cannot poison the connection

### LINKED ISSUE #2669: Bug: PostgresConnector never rolls back after a failed query, poisoning the shared connection for the rest of the process
### Summary

`wren/connector/postgres.py` opens a single, long-lived `psycopg` connection per `PostgresConnector` instance and never calls `.rollback()` after a failed query. Since `WrenEngine._get_connector()` caches this connector for the engine's lifetime, and `ServeContext.engine` in the MCP server is a single instance shared across the whole server process, **one failing query from any client leaves the shared Postgres connection in an aborted-transaction state for every subsequent query from every client**, until the process is restarted.

### Version

`wrenai[postgres]==0.13.2` (latest on PyPI at time of filing)

### Where

- `wren/connector/postgres.py` — `PostgresConnector.__init__` creates one `psycopg.connect(...)` and stores it as `self.connection`. `query()` and `dry_run()` catch exceptions and re-raise as `WrenError`, but never call `self.connection.rollback()`.
- `wren/engine.py` — `_get_connector()` caches the connector: `if self._connector is None: self._connector = get_connector(...)`, reused for the lifetime of the `WrenEngine` instance.
- `wren/mcp_server.py` — `ServeContext` is documented in-code as "Shared state captured once at startup and used by every tool handler", i.e. one `Engine`/connector for the whole MCP server process.

### Steps to reproduce

1. Start `wren serve mcp` against a Postgres data source.
2. Call `run_sql` (or `dry_run`) with SQL that fails at execution time on Postgres itself (not at Wren's SQL-planning stage), e.g. a query referencing a column that doesn't exist:
   ```sql
   SELECT nonexistent_column FROM some_table
   ```
   This returns the expected error.
3. Immediately call `run_sql` again with a completely valid query, e.g.:
   ```sql
   SELECT COUNT(*) FROM some_table
   ```

### Expected

Step 3 succeeds normally.

### Actual

Step 3 fails with:
```
[GENERIC_USER_ERROR] current transaction is aborted, commands ignored until end of transaction block
```
Every subsequent query on that server process fails the same way — from any client, any session — until the container/process is restarted.

### Why this matters

Because the connector/engine is shared per-process (not per-request, per-session, or pooled), this isn't a caller-side mistake to recover from — a single bad query from one MCP client degrades the server for every other concurrent client.

### Suggested fix

Roll back on failure before re-raising, e.g. in `PostgresConnector.query()` / `dry_run()`:
```python
except Exception as e:
    try:
        self.connection.rollback()
    except Exception:
        pass
    raise WrenError(...) from e
```
A more robust fix would move to short-lived connections or an actual connection pool (e.g. `psycopg_pool`) with reset-on-return semantics, so a single connector instance doesn't hold one Postgres session's transaction state indefinitely.

### Environment

- Reproduced via `wren serve mcp --transport http`, connecting with an MCP client (`mcp` Python SDK, `streamablehttp_client` / `ClientSession`).
- Postgres backend (local dev instance).

lucifer726:
## What changes

`PostgresConnector` now connects with `autocommit=True` (via `kwargs.setdefault`, so an
explicit caller-supplied value still wins). Without it, psycopg opens an implicit
transaction per statement and a failed statement leaves the session
`idle in transaction (aborted)` — every later statement on that connection fails until
the process restarts.

Closes #2669.

## Why this shape

psycopg is the only connector in this repo that never set autocommit — `canner.py:240`,
`redshift.py:44` and `mysql.py:518` all do. Measured against a real Postgres 16 in a
clean container:

    autocommit=False   after success=INTRANS   after failure=INERROR   next stmt=InFailedSqlTransaction
    autocommit=True    after success=IDLE      after failure=IDLE      next stmt=succeeds

Two things follow:

1. The poisoned connection is not something to clean up after the fact — it is something
   not to create.
2. With autocommit off, even a **successful** statement leaves the session
   `idle in transaction`. This connector is long-lived and process-shared
   (`WrenEngine._get_connector` caches it; the MCP server's `ServeContext` hands one
   engine to every tool handler), so it held one snapshot open for its entire life. A
   rollback-on-failure fix does not touch that.

An earlier revision of this PR took the rollback-on-failure route. @AmirF194's review
showed it was patching the symptom and missing the success path, so this revision drops
`_rollback_after_failure` and its six except-branch call sites in favour of the one
`setdefault`.

## Verification

- Reproduced the original failure on a real Postgres 16 (testcontainers) against the
  unfixed connector: `2 failed`, `[GENERIC_USER_ERROR] current transaction is aborted,
  commands ignored until end of transaction block`. With the fix: `2 passed`.
- The integration tests are **unchanged from the previous revision** and still pass — they
  assert the behaviour (a valid query after a failed one succeeds), not the mechanism.
- New unit tests (`tests/unit/`, runs in the default unit job): autocommit defaults on; an
  explicit `autocommit=False` in `connection_info.kwargs` still wins; other kwargs
  survive. Confirmed 2 of the 3 fail without the source change.
- `tests/unit/` 1153 passed; `tests/connectors/test_postgres.py -m postgres` 37 passed;
  `ruff format --check src/` and `ruff check src/` clean.


lucifer726:
Checked this against the repo's actual test layout — I don't think the stub can
leak in either configuration, so I've left it as is:

- **With the `postgres` extra installed** (the only configuration where
  `tests/connectors/test_postgres.py` can run at all): pytest imports that module
  during collection, and its top-level `import psycopg` puts the real module in
  `sys.modules` before any test executes. `_ensure_psycopg_stub()` then hits its
  `if "psycopg" in sys.modules: return` guard and never installs the fake.
  Verified by running both in one process:

      pytest tests/unit/test_postgres_rollback.py \
             tests/connectors/test_postgres.py::TestPostgresConnectorTransactionRecovery -m ""
      → 8 passed

- **Without the extra** (the `test-unit` job): `tests/connectors/test_postgres.py`
  cannot be imported at all — both `psycopg` and `testcontainers.postgres` are
  missing — so there is nothing for the stub to shadow. The two CI jobs are
  separate runners anyway (`wren-ci.yml`: `test-unit` vs `test-connector`).

The stub follows the existing pattern in
`tests/unit/test_postgres_semicolon_unlimited.py`, whose module docstring records
the same rationale. Happy to scope it with a fixture if a maintainer prefers that
regardless.

core/wren/src/wren/connector/postgres.py
```diff
@@ -301,6 +301,25 @@ def __init__(self, connection_info):
         )
         self._closed = False
 
+    def _rollback_after_failure(self) -> None:
```
AmirF194: I read this against `33cfc93c` and traced where `_get_connector()` is used outside this file.

`ServeContext` in `mcp_server.py` holds one `engine` (and so one `PostgresConnector`, one
`psycopg` connection) for the whole process, and `build_server()` can run the MCP server on
`transport="streamable-http"`. The tool functions registered there (`query`, `dry_run`, etc.)
are plain `def`, not `async def`, which the MCP SDK runs off the event loop in a worker
thread, so two concurrent HTTP tool calls can be executing on separate threads at the same
time, both going through the same connector.

If that happens, `_rollback_after_failure()` calls `self.connection.rollback()` on the shared
connection while a second thread's `cursor.execute()` on that same connection may still be
in flight, which is unsafe with psycopg (a connection is documented as not usable
concurrently from multiple threads). Before this PR a failed statement left every later
caller failing with "transaction is aborted" until restart; after it, a failed statement can
now actively cancel a different, unrelated in-flight query's transaction instead.

I have not built a repro for this since it needs the streamable-http transport running with
two real concurrent tool calls, so this is a question rather than a claim: is there a
serialization point (a lock, a connection-per-request, single-flight dispatch) between
`ServeContext.engine` and the transport that I'm missing by reading the code alone? If not,
the rollback in `_rollback_after_failure` (postgres.py:304) is only safe under the same
single-flight assumption the pre-existing sharing already depended on, and it would be worth
saying so in a comment there.

lucifer726:
Good catch — I went and read it rather than assuming, and there is no serialization
point I can point you to.

`mcp_server.py` at `33cfc93c`:
- `ServeContext` is a dataclass holding one `engine`, documented as "Shared state
  captured once at startup and used by every tool handler".
- The tool functions are sync `def` (`dry_run`, `query_cube`, `dry_plan`,
  `list_models`) — none are `async def`.
- `grep -nE "Lock|Semaphore|threading|single.?flight"` over that file returns nothing.
- `transport == "http"` runs `mcp.run(transport="streamable-http")`.

So your read holds, and the consequence is worth stating plainly rather than arguing:
before this PR a failed statement left every later caller failing until restart; with
it, a failed statement can call `rollback()` while another thread's `cursor.execute()`
is in flight. That is a different failure mode, not strictly a smaller one.

I did try the fuller fix — a `threading.Lock` around the execute/rollback block — and
backed it out for scope reasons rather than doctrinal ones:

- It breaks two existing tests (`tests/unit/test_postgres_semicolon_unlimited.py`),
  which build the connector via `PostgresConnector.__new__` and so never receive the
  new instance attribute. Fixing that means touching three test files' mock helpers
  alongside the connector.
- The unsafe sharing predates this PR — the connector was never usable concurrently —
  so serialising it is a separate change with its own design question (lock vs.
  per-request connection vs. `psycopg_pool` with reset-on-return, which #2669 already
  floats).

Instead I did what you suggested: `_rollback_after_failure`'s docstring now records
that it inherits the same single-flight assumption the existing sharing relied on, and
names where that assumption comes from (`85a85bac`).

Happy to open a separate issue for the concurrency problem, or leave it to you since
your trace of the transport/threading path is more precise than mine — whichever you
prefer. And if a maintainer would rather this PR carry the lock and the test updates,
say the word and I'll put it back.

core/wren/src/wren/connector/postgres.py
AmirF194: Agreed, that is a reasonable scope call. Not pushing for a separate issue from my side, the docstring note is enough to flag the assumption for whoever picks up the pooling work later.

core/wren/src/wren/connector/postgres.py
```diff
@@ -301,6 +301,33 @@ def __init__(self, connection_info):
...
```
AmirF194: I checked psycopg's autocommit behavior against a real Postgres 16 in a clean container rather than trusting the driver docs from memory.

`self.connection = psycopg.connect(...)` at line 290 never sets `autocommit`, unlike every other connector in this codebase (`redshift.py:44`, `canner.py:240`, `mysql.py:518`). With autocommit off, a failed statement leaves the session `idle in transaction (aborted)`, which is exactly the bug this PR patches with three new `_rollback_after_failure()` call sites. With autocommit on, a failed statement never poisons the next one, no rollback needed:

```
autocommit=False: second (valid) statement FAILED: InFailedSqlTransaction: current transaction is aborted, commands ignored until end of transaction block
autocommit=True: second (valid) statement SUCCEEDED: (1,)
```

That also covers a gap this PR leaves open: a successful query does not commit either, so today this long-lived, process-shared connector holds one continuously open transaction for its whole life except when a failure now triggers the new rollback. Setting `autocommit=True` in `__init__` would remove `_rollback_after_failure` and the three added except branches entirely, and match the rest of the connectors.

lucifer726:
You're right, and thanks — this is a better fix than what I had.

I checked both claims independently rather than taking them on faith. The connector
inconsistency is real (`canner.py:240`, `redshift.py:44`, `mysql.py:518` all set
autocommit; psycopg was the only one that didn't), and a clean Postgres 16 container
reproduces your numbers:

    autocommit=False   after success=INTRANS   after failure=INERROR   next=InFailedSqlTransaction
    autocommit=True    after success=IDLE      after failure=IDLE      next=succeeds

The part I had missed entirely is your second point: with autocommit off a *successful*
statement also leaves the session `INTRANS`, so this process-shared connector was holding
one snapshot open for its whole life, and rollback-on-failure never touched that. That is
what settles it — my version was patching the symptom.

Pushed the rewrite: `_rollback_after_failure` and its six except-branch call sites are
gone, replaced by `kwargs.setdefault("autocommit", True)` so an explicit caller value
still wins. The connector diff is now +10 lines instead of +25 with a new method, and the
concurrency question you raised earlier goes away with it — there is no longer a rollback
that can land on another thread's in-flight statement.

Worth noting the integration tests carried over unchanged and still pass: they assert that
a valid query after a failed one succeeds, not that `rollback()` was called. They ended up
validating a replacement implementation for free.

goldmedal (approved):
Thanks @lucifer726, make sense to me

goldmedal:
Thanks @AmirF194 for reviewing

PR #2684, opened 2026-08-19
feat(pydantic): forward session properties to the engine

### LINKED ISSUE #2638: wren-pydantic: session properties are not forwarded → RLAC-protected models are unusable from the SDK
# wren-pydantic: session properties are not forwarded → RLAC-protected models are unusable from the SDK

## Summary

`WrenToolkit.query()` / `dry_plan()` and the Pydantic AI tools `wren_query` / `wren_dry_plan` never pass **session properties** to the engine. Any project that enables row-level access control (RLAC) becomes unusable from the SDK: every query errors because the required session property is missing, and there is no way to supply it per request. This blocks the SDK for any multi-tenant / permissioned deployment.

## Environment

- `wren-pydantic` 0.1.x (`sdk/wren-pydantic`, main branch)
- `wrenai` 0.11.0
- Under the hood, `wren.engine.WrenEngine.query(sql, limit, properties)` **already supports** `properties` — the SDK just doesn't forward it.

## Repro

MDL with an RLAC-protected model:

```json
"rowLevelAccessControls": [
  {
    "name": "only_own",
    "requiredProperties": [{ "name": "session_user_id", "required": true }],
    "condition": "user_id = @session_user_id"
  }
]
```

Calling the engine directly:

```python
engine.dry_plan("SELECT * FROM orders",
                properties={"session_user_id": "'u_42'"})
# ✅ WITH properties:
#   ... WHERE orders.user_id = 'u_42'   (predicate injected)

engine.dry_plan("SELECT * FROM orders")
# ❌ WITHOUT properties:
#   [INVALID_SQL] session property session_user_id is required for
#   'only_own' rule but not found in headers
```

But the SDK never lets you pass `properties`:

```python
toolkit = WrenToolkit.from_project("./my_project")
toolkit.query("SELECT * FROM orders")   # no `properties` kwarg
```

So with RLAC enabled, `toolkit.query()` always hits the "required session property not found" error — there is **no way to run even a permitted query**.

## Root cause

- `sdk/wren-pydantic/src/wren_pydantic/_toolkit.py`
  - `query(self, sql, limit=None)` → `engine.query(sql, limit=limit)` (no `properties`)
  - `dry_plan(self, sql)` / `dry_run(self, sql)` → same, no `properties`
- `sdk/wren-pydantic/src/wren_pydantic/_tools.py`
  - `wren_query` / `wren_dry_plan` tools call `toolkit.query(sql)` / `toolkit.dry_plan(sql)` — no way to inject per-user session properties

## Expected behavior

- `WrenToolkit.query(sql, limit=None, properties=None)` (and `dry_plan` / `dry_run`) forward `properties` to the engine.
- The LLM-facing `wren_query` / `wren_dry_plan` tools can receive session properties, ideally sourced from Pydantic AI `RunContext` / `deps` — session properties are **per-request / per-user**, not per-project.

## Suggested minimal patch

```python
def query(self, sql: str, limit: int | None = None, properties: dict | None = None):
    ...
    result = engine.query(sql, limit=limit, properties=properties)
```

And thread per-request properties into the tools via `takes_ctx=True` + `ctx.deps`, or as a `properties` tool argument.

## Why it matters

RLAC (`rowLevelAccessControls`) is the core multi-tenant isolation mechanism in the OSS engine. Without `properties` support in `wren-pydantic`, the SDK integration path is a hard blocker for any permissioned deployment — it either errors on protected models or (if a model is mistakenly unprotected) silently queries the full table.

Related note: the `wren query` / `wren dry-plan` CLI commands have the same gap (no `--property` flag), so the same fix could cover both entry points.

Happy to open a PR if the approach looks right.

lucifer726:
## What changes

`WrenToolkit.query()`, `dry_plan()` and `dry_run()` now accept `properties` and forward
it to the engine, so a caller can supply MDL session properties per call.

`WrenEngine.query/dry_plan/dry_run` already accept `properties` (`core/wren/src/wren/engine.py`)
— the toolkit simply never passed it, so a model guarded by row-level access control could
not be read through the SDK at all: planning fails when a required session property is
missing, and there was no way to supply one.

Closes #2638.

## Why the Pydantic AI tools are unchanged

`wren_query` / `wren_dry_plan` keep their current signatures on purpose. Their `ctx` is
documented as ignored because "the toolkit already captures all required state"
(`_tools.py`), and turning a session property into a model-fillable tool argument would let
the LLM choose the identity that RLAC is keyed on — the model would be deciding which rows
it is allowed to see. Binding session state for agent runs is a separate design question
(toolkit- or run-scoped), so this PR only opens the direct Python API that the issue asks
for.

## How this was verified

Scope of the claim: these tests assert the parameter reaches `WrenEngine`. I did not stand
up a live RLAC project against a real data source, so this PR does not claim an end-to-end
RLAC observation — the engine-side behaviour is the existing, already-tested path.

- 3 new tests assert `properties` is forwarded by `query` / `dry_plan` / `dry_run`.
- 3 existing delegation assertions updated to record the new call contract
  (`properties=None` by default), since they pinned the exact kwargs.
- Confirmed the new tests fail without the source change (6 failed with `_toolkit.py`
  reverted, 8 passed with it).
- `pytest tests/unit` → 103 passed. `ruff check .` and `ruff format --check .` clean.

## Summary by CodeRabbit

* **New Features**
  * Added optional session properties to query, plan preview, and dry-run operations.
  * Session properties are forwarded to the Wren engine, enabling them to influence planning and row-level access control predicates.

* **Bug Fixes**
  * Preserved existing connector caching behavior for queries.
  * Added coverage ensuring custom session properties are passed through unchanged.

### DIFF
--- a/sdk/wren-pydantic/src/wren_pydantic/_toolkit.py
+++ b/sdk/wren-pydantic/src/wren_pydantic/_toolkit.py
@@ -118,24 +118,42 @@ class WrenToolkit:
 
     # ── Direct Python API (sync only — see module docstring) ──────────────
 
-    def query(self, sql: str, limit: int | None = None) -> pa.Table:
-        """Execute SQL through the Wren context layer. Returns a pyarrow Table."""
+    def query(
+        self,
+        sql: str,
+        limit: int | None = None,
+        properties: dict[str, Any] | None = None,
+    ) -> pa.Table:
+        """Execute SQL through the Wren context layer. Returns a pyarrow Table.
+
+        ``properties`` carries MDL session properties and is forwarded to the
+        engine's planning path. A model guarded by row-level access control
+        needs the value its rule declares required — without it planning fails,
+        so RLAC-protected models cannot be read at all.
+        """
         engine = self._build_engine()
         try:
-            result = engine.query(sql, limit=limit)
+            result = engine.query(sql, limit=limit, properties=properties)
         finally:
             self._connector_cache = engine._connector
         return result
 
-    def dry_plan(self, sql: str) -> str:
-        """Plan SQL through MDL and return the expanded SQL in target dialect."""
-        return self._build_engine().dry_plan(sql)
+    def dry_plan(self, sql: str, properties: dict[str, Any] | None = None) -> str:
+        """Plan SQL through MDL and return the expanded SQL in target dialect.
 
-    def dry_run(self, sql: str) -> None:
-        """Validate SQL by planning and asking the DB to plan it without executing."""
+        ``properties`` carries MDL session properties (see :meth:`query`); RLAC
+        predicates are injected during planning, so they apply here too.
+        """
+        return self._build_engine().dry_plan(sql, properties=properties)
+
+    def dry_run(self, sql: str, properties: dict[str, Any] | None = None) -> None:
+        """Validate SQL by planning and asking the DB to plan it without executing.
+
+        ``properties`` carries MDL session properties (see :meth:`query`).
+        """
         engine = self._build_engine()
         try:
-            engine.dry_run(sql)
+            engine.dry_run(sql, properties=properties)
         finally:
             self._connector_cache = engine._connector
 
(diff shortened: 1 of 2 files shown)


goldmedal (approved):
Thanks @lucifer726

PR #2690, opened 2026-08-21
fix(context): reject lossy v1 view upgrades

### LINKED ISSUE #2687: `wren context upgrade` silently deletes v1 views that the loader dropped
## Summary

`wren context upgrade` writes the v1→v2 view directories from `_load_views_v1()`, which **drops** malformed entries, and then unconditionally `unlink()`s `views.yml`. Anything the loader dropped is therefore deleted from disk with no error, no warning, and exit code 0. In the worst case the entire `views.yml` is destroyed.

Dropping malformed entries in the *loader* is correct and intentional (`validate_project` is the designated reporter). Doing it on a *migration path that deletes the source file* is data loss.

Measured on `main` @ `ce9513d6`, `wren 0.13.3`.

## Repro A — individual views vanish

```yaml
# views.yml
views:
  - name: kept
    statement: SELECT 1
  - statement: SELECT id FROM orders     # no name
  - just_a_bare_string                   # not a mapping
```

```console
$ wren context upgrade --path "$P"
Upgrading project from schema_version 1 -> 5...
  + models/orders/metadata.yml
  + views/kept/metadata.yml
  ...
  - views.yml
  * wren_project.yml (schema_version 1 -> 5)

Upgrade complete. Run `wren context validate` to check the result.
$ echo $?
0
$ ls views/
kept
```

`SELECT id FROM orders` and the bare string are gone permanently. The plan output never mentions them.

## Repro B — the whole file is destroyed

A `views.yml` hand-written as a mapping instead of a list — a very plausible edit:

```yaml
# views.yml
views:
  revenue:
    statement: SELECT sum(amount) FROM orders
  churn:
    statement: SELECT 1
```

```console
$ wren context upgrade --path "$P"
  ...
  - views.yml
Upgrade complete. Run `wren context validate` to check the result.
$ echo $?
0
$ ls views/ 2>&1
ls: views/: No such file or directory
```

Both view definitions are gone and no `views/` directory was created. Only git can recover them.

## Root cause

- `context.py:612-614` — `_load_views_v1()` returns `[]` for a non-list `views:` value and filters non-mapping entries out of the list.
- `context.py:1708-1711` — `_apply_v1_to_v2()` iterates that already-filtered list and additionally `continue`s on any view without a `name`.
- `context.py:1734-1737` — `views.yml` is then unlinked regardless of how many views were actually written.

## Why nothing warns you

`wren context validate` *does* report the underlying problems on the v1 project and exits 1:

```console
[ERROR] views.yml > views[2]: view entry must be a mapping, got str
[ERROR] views/views[1]/metadata.yml: view missing 'name'
```

But `wren context upgrade` does not run validation, and `--dry-run` gives no hint either — it prints `Would create: views/kept/metadata.yml` / `Would delete: views.yml` with nothing to indicate that two views are being dropped. A user reading the plan cannot tell the difference between this and a clean migration.

## Expected

`wren context upgrade` should never delete a source file whose content it did not carry across. Either:

1. **Abort in the preflight.** `_plan_v1_to_v2()` already rejects other unmigratable input (e.g. non-string view statements) before the first write. Extending it to reject malformed `views.yml` shape, non-mapping entries, and nameless views would fit that shape and keep the source intact.
2. **Don't filter on the migration path.** Have `_apply_v1_to_v2()` read the raw YAML rather than going through the normalising loader, so migration is byte-preservingly lossless and `validate_project` keeps reporting the mistake afterwards in the v2 layout.

Option 2 is the more general fix: it closes this class of bug for every field at once, rather than requiring a new preflight rule each time a loader learns to filter something.

## Scope note

This is the same bug class as the model `columns` case currently being fixed in #2614, but for views, and it is already live on `main` — #2614's head and `main` behave identically here. Whichever way #2614 lands, the views path will still need this, so it is worth tracking separately.

Cubes are not affected: `_load_cubes_v1()` leaves a non-mapping cube file's source in place.

## Minor, possibly separate

`wren context validate` prints two `[ERROR]` lines above and then summarises as `3 warning(s), 0 errors.` The exit code is correct (1), so this is a cosmetic tally bug in the summary line, but it does make the errors easy to miss.

NNoradrenaline:
## Summary

Abort the v1→v2 context upgrade before deleting `views.yml` when the legacy file contains content that `_load_views_v1()` would drop.

This follows the same migration-safety pattern as the model-columns fix in #2614: loader normalisation remains unchanged for runtime/build consumers, while the migration path performs a raw-source preflight before any write.

`_load_views_v1()` intentionally filters malformed entries and returns `[]` for a non-list `views:` container. The upgrade path then migrated only the surviving views and unconditionally deleted the original `views.yml`.

That meant hand-edited content such as a nameless view, a bare scalar entry, or a mapping-shaped `views:` container could be permanently discarded with exit code 0.

## Changes

- Add `_reject_malformed_v1_views()` to inspect the raw v1 `views.yml` before migration.
- Reject malformed `views:` containers, non-mapping list entries, and nameless views with `UpgradeError`.
- Call the preflight from `_plan_v1_to_v2()`, which also makes `_apply_v1_to_v2()` re-check the source before the first filesystem mutation.
- Update the existing malformed-view upgrade tests so they assert abort + source preservation instead of successful lossy migration.
- Add regression coverage for the exact mixed valid/nameless/bare-entry case from #2687.

## Test plan

Wren SDK CI passes on the final commit, including:

- lint
- unit tests
- MCP tests
- UI tests
- memory tests
- PostgreSQL tests
- MySQL tests

Focused regression coverage also verifies:

- `plan_upgrade()` rejects malformed legacy views before changing the project.
- The original `views.yml` remains intact.
- A valid plan followed by late corruption is re-checked by `apply_upgrade()`.
- The model source and schema version remain unchanged when the upgrade aborts.

## Duplicate check

No open PR found for #2687 / malformed v1 view upgrade data loss.

Fixes #2687

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved v1-to-v2 upgrades by validating `views.yml` before applying changes.
  * Rejects malformed files, duplicate YAML keys, unsupported root settings, invalid view collections, unnamed views, and non-mapping entries with a clear upgrade error.
  * Preserves source files when invalid view definitions are detected.
  * Handles empty YAML documents safely by removing only the obsolete configuration file without creating invalid view files.

### DIFF
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -1713,7 +1713,16 @@ def _reject_malformed_v1_views(project_path: Path) -> None:
             f"views.yml: must be a mapping with a 'views' key, got {type(raw).__name__}"
         )
     else:
-        raw_views = raw.get("views") if isinstance(raw, dict) else None
+        raw_views = None
+        if isinstance(raw, dict):
+            unexpected_keys = set(raw) - {"views"}
+            if unexpected_keys:
+                problems.append(
+                    "views.yml: unsupported root keys would be discarded: "
+                    + ", ".join(sorted(str(key) for key in unexpected_keys))
+                )
+            raw_views = raw.get("views")
+
         if raw_views is not None and not isinstance(raw_views, list):
             problems.append(
                 f"views.yml > views: must be a list, got {type(raw_views).__name__}"
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -1493,6 +1493,86 @@ class UpgradeError(Exception):
     """Raised when a project upgrade cannot proceed."""
 
 
+class _UniqueKeySafeLoader(yaml.SafeLoader):
+    """SafeLoader variant that rejects explicit duplicate mapping keys."""
+
+    def construct_mapping(self, node, deep=False):
+        """Reject explicit duplicates without breaking YAML merge-key semantics."""
+        seen = set()
+        for key_node, _ in node.value:
+            # SafeConstructor.flatten_mapping() handles << later. Checking the
+            # merge node itself here would try to construct the special merge
+            # tag too early and reject YAML that yaml.safe_load() accepts.
+            if key_node.tag == "tag:yaml.org,2002:merge":
+                continue
+            key = self.construct_object(key_node, deep=deep)
+            try:
+                if key in seen:
+                    raise yaml.constructor.ConstructorError(
+                        "while constructing a mapping",
+                        node.start_mark,
+                        f"found duplicate YAML key {key!r}",
+                        key_node.start_mark,
+                    )
+                seen.add(key)
+            except TypeError as exc:
+                raise yaml.constructor.ConstructorError(
+                    "while constructing a mapping",
+                    node.start_mark,
+                    f"found unhashable YAML key {key!r}",
+                    key_node.start_mark,
+                ) from exc
+        return super().construct_mapping(node, deep=deep)
+
+
+def _collect_yaml_node_ids(node, seen=None) -> set[int]:
+    """Collect node identities below a composed YAML node, guarding cycles."""
+    if seen is None:
+        seen = set()
+    node_id = id(node)
+    if node_id in seen:
+        return seen
+    seen.add(node_id)
+
+    if isinstance(node, yaml.nodes.MappingNode):
+        for key_node, value_node in node.value:
+            _collect_yaml_node_ids(key_node, seen)
+            _collect_yaml_node_ids(value_node, seen)
+    elif isinstance(node, yaml.nodes.SequenceNode):
+        for item in node.value:
+            _collect_yaml_node_ids(item, seen)
+    return seen
+
+
+def _v1_views_anchor_root_keys(text: str) -> set[str]:
+    """Return extra root keys whose value node is aliased inside views."""
+    root = yaml.compose(text, Loader=yaml.SafeLoader)
+    if not isinstance(root, yaml.nodes.MappingNode):
+        return set()
+
+    views_node = None
+    for key_node, value_node in root.value:
+        if (
+            isinstance(key_node, yaml.nodes.ScalarNode)
+            and key_node.value == "views"
+        ):
+            views_node = value_node
+            break
+    if views_node is None:
+        return set()
+
+    referenced_ids = _collect_yaml_node_ids(views_node)
+    anchor_keys: set[str] = set()
+    for key_node, value_node in root.value:
+        if (
+            isinstance(key_node, yaml.nodes.ScalarNode)
+            and key_node.value != "views"
+            and id(value_node) in referenced_ids
+        ):
+            anchor_keys.add(key_node.value)
+    return anchor_keys
+
+
 def _resolve_upgrade_directory(
     project_path: Path,
     collection: str,
@@ -1555,6 +1635,19 @@ def _resolve_upgrade_file(target_directory: Path, filename: str) -> Path:
         ) from exc
     return resolved_file
 
+
+def _record_unique_upgrade_target(
+    seen: dict[str, str], target: str, source_kind: str
+) -> None:
+    """Reject exact and case-insensitive v1→v2 destination collisions."""
+    key = target.casefold()
+    previous = seen.get(key)
+    if previous is not None:
+        detail = f"'{target}'" if previous == target else f"'{previous}' and '{target}'"
+        raise UpgradeError(
+            f"Cannot upgrade: multiple legacy {source_kind} map to {detail}"
+        )
+    seen[key] = target
 
 def _validate_upgrade_view_statement(name: str, statement: Any) -> str | None:
     """Reject malformed legacy view statements before migration."""
@@ -1705,25 +1798,10 @@ def _reject_malformed_v1_views(project_path: Path) -> None:
     if not views_file.exists():
         return
 
-    class _UniqueKeySafeLoader(yaml.SafeLoader):
-        """SafeLoader variant that rejects explicit duplicate mapping keys."""
-
-        def construct_mapping(self, node, deep=False):
-            """Reject duplicate keys before normal SafeLoader construction."""
-            seen = set()
-            for key_node, _ in node.value:
-                key = self.construct_object(key_node, deep=deep)
-                if key in seen:
-                    raise UpgradeError(
-                        f"Cannot upgrade: views.yml contains duplicate YAML key {key!r}"
-                    )
-                seen.add(key)
-            return super().construct_mapping(node, deep=deep)
-
+    source_text = views_file.read_text(encoding="utf-8")
     try:
-        raw = yaml.load(
-            views_file.read_text(encoding="utf-8"), Loader=_UniqueKeySafeLoader
-        )
+        anchor_root_keys = _v1_views_anchor_root_keys(source_text)
+        raw = yaml.load(source_text, Loader=_UniqueKeySafeLoader)
     except (TypeError, yaml.YAMLError) as exc:
         raise UpgradeError(f"Cannot upgrade: invalid views.yml: {exc}") from exc
 
@@ -1736,10 +1814,10 @@ def _reject_malformed_v1_views(project_path: Path) -> None:
     else:
         raw_views = None
         if isinstance(raw, dict):
-            unexpected_keys = set(raw) - {"views"}
+            unexpected_keys = set(raw) - {"views"} - anchor_root_keys
             if unexpected_keys:
                 problems.append(
-                    "views.yml: unsupported root keys would be discarded: "
+                    "views.yml: unsupported root keys are not preserved by migration: "
                     + ", ".join(sorted(str(key) for key in unexpected_keys))
                 )
             raw_views = raw.get("views")
@@ -1778,6 +1856,7 @@ def _plan_v1_to_v2(project_path: Path) -> tuple[list[str], list[str]]:
     _reject_malformed_v1_views(project_path)
 
     # Models: flat files → directories
+    seen_model_targets: dict[str, str] = {}
     models = _load_models_v1(project_path)
     for model in models:
         source_dir = model.pop("_source_dir", None)
@@ -1790,12 +1869,17 @@ def _plan_v1_to_v2(project_path: Path) -> tuple[list[str], list[str]]:
             created.append(ref_sql_file.relative_to(project_root).as_posix())
 
         metadata_file = _resolve_upgrade_file(model_dir, "metadata.yml")
-        created.append(metadata_file.relative_to(project_root).as_posix())
+        metadata_target = metadata_file.relative_to(project_root).as_posix()
+        _record_unique_upgrade_target(
+            seen_model_targets, metadata_target, "model files"
+        )
+        created.append(metadata_target)
 
         if source_dir:
             deleted.append(f"models/{source_dir}.yml")
 
     # Views: single file → directories
+    seen_view_targets: dict[str, str] = {}
     views = _load_views_v1(project_path)
     for view in views:
         name = view.get("name")
@@ -1809,14 +1893,16 @@ def _plan_v1_to_v2(project_path: Path) -> tuple[list[str], list[str]]:
             created.append(sql_file.relative_to(project_root).as_posix())
 
         metadata_file = _resolve_upgrade_file(view_dir, "metadata.yml")
-        created.append(metadata_file.relative_to(project_root).as_posix())
+        metadata_target = metadata_file.relative_to(project_root).as_posix()
+        _record_unique_upgrade_target(seen_view_targets, metadata_target, "views")
+        created.append(metadata_target)
 
     views_file = project_path / "views.yml"
     if views_file.exists():
         deleted.append("views.yml")
 
     # Cubes: flat files → directories
-    seen_cube_targets: set[str] = set()
+    seen_cube_targets: dict[str, str] = {}
     cubes = _load_cubes_v1(project_path)
     for cube in cubes:
         source_file = cube.pop("_source_file", None)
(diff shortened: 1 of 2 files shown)


NNoradrenaline:
@coderabbitai review

goldmedal:
## Review

The direction is right and the test hygiene is good — asserting source preservation and re-checking in `apply_upgrade()` is exactly what this bug class needs. CI is green and all 195 `test_context.py` tests pass locally at `3739e91`. But the duplicate-key loader introduces a regression, and the preflight still leaves one live instance of the very data-loss class it is closing.

### 🔴 1. `_UniqueKeySafeLoader` breaks YAML merge keys — regression vs `main`

`construct_mapping` calls `self.construct_object(key_node)` on every raw key node *before* delegating to `SafeConstructor.construct_mapping`, which is what runs `flatten_mapping()`. At that point a `<<` key still carries `tag:yaml.org,2002:merge`, and `SafeConstructor` has no constructor for it.

```yaml
# views.yml — no extra root keys
views:
  - name: a
    <<: {statement: SELECT 1}
```

| | result |
|---|---|
| `main` | migrates cleanly → `views/a/metadata.yml` |
| this PR | `UpgradeError: Cannot upgrade: invalid views.yml: could not determine a constructor for the tag 'tag:yaml.org,2002:merge'` |

`yaml.safe_load()` handles this file fine, so `_load_views_v1()` and `validate_project()` still accept it — only the upgrade path now rejects it, with an error that misattributes the cause.

Minimal fix (verified locally — merge keys migrate again, and all 33 duplicate/views tests still pass):

```python
for key_node, _ in node.value:
    if key_node.tag == "tag:yaml.org,2002:merge":
        continue
    key = self.construct_object(key_node, deep=deep)
```

Skipping the merge node rather than calling `flatten_mapping()` first also avoids flagging legitimate merge-overridden keys as duplicates.

### 🔴 2. Colliding view target names are still silently lossy

The preflight validates *shape*, not *target uniqueness*. Two views resolving to the same directory still produce exit 0, a deleted `views.yml`, and one destroyed view — the exact failure mode of #2687:

```yaml
views:
  - name: revenue
    statement: SELECT 1
  - name: revenue
    statement: SELECT 2
```

```
plan created: ['views/revenue/metadata.yml', 'views/revenue/metadata.yml']
apply OK.  views.yml exists? False
  revenue -> statement: SELECT 2      # SELECT 1 is gone
```

Worse on a case-insensitive filesystem (macOS/Windows default) — `name: Revenue` plus `name: revenue` yields a single `views/Revenue/` directory containing the *`revenue`* view, while the plan claims both were created.

`_apply_v1_to_v2()` already has the right precedent for cubes:

```python
if target in seen_cube_targets:
    raise UpgradeError(f"Cannot upgrade: multiple legacy cube files map to '{target}'")
```

Views have no equivalent. For what it's worth, models have the same gap — two `models/*.yml` files sharing a `name:` collapse into one directory and both sources are deleted — so a shared target-collision check in the preflight would cover all three, and would sit earlier than the cube check, which currently fires only *after* model files have already been written and their sources unlinked.

### 🟡 3. The root-key rule false-positives on anchor-definition blocks

```yaml
base: &b
  statement: SELECT 1
views:
  - name: a
    <<: *b
```

`main` migrates this losslessly — the anchor is fully expanded into the view before anything is written. The new rule rejects it as `unsupported root keys would be discarded: base`, which is inaccurate: nothing is discarded, the content is carried across by expansion. Worth either excluding keys consumed only as anchor sources, or softening the message so it does not assert data loss it cannot establish.

### 📝 Notes, not findings

- The CodeRabbit `Merge Risk: 🟡 Moderate` verdict cites `.github/workflows/address-coderabbit-2690.yml` and a write-capable checkout credential. That file is **not** in the branch at `3739e91` — the verdict is stale and should not block.
- #2687 recommends Option 2 (lossless raw migration) over Option 1 (per-field preflight) precisely because Option 1 needs a new rule per filter. Findings 1–3 are all instances of that cost. Option 1 is a defensible, consistent choice given #2614, but it is worth an explicit maintainer decision rather than defaulting.

---

Verified against `3739e91` in a clean worktree: `uv sync` + `pytest tests/unit/test_context.py` (195 passed). Each repro above was run as a script comparing actual `main` and PR behaviour; the fix in finding 1 was applied locally and confirmed to keep the existing tests green.

*Baseline note (edited after posting):* findings 1–3 were re-verified against current `main` at `bfc2b9ab` — my first pass used a stale local checkout. The comparisons are unchanged: a `views.yml` using a `<<:` merge key still migrates cleanly on `main` and is still rejected by this branch, and all three findings stand as described.

goldmedal:
## Correction to my earlier review — finding 4 was wrong

I claimed the inline `from wren.context import UpgradeError as _UE  # noqa: PLC0415` lines were redundant because `UpgradeError` is already imported at module scope. That is incorrect, and the comment on those lines is right.

`test_context.py:718` and `:748` call `importlib.reload(ctx)`. `reload()` re-executes the module into the *same* module dict, so `wren.context.UpgradeError` is rebound to a new class object. The test module's top-level `UpgradeError` still points at the old class, while `plan_upgrade` — also captured at import time, but resolving its globals through that same mutated dict — raises the *new* one. The two stop matching.

Removing the inline imports and running the file: **57 failed, 138 passed**. The pattern is load-bearing; please disregard finding 4.

Findings 1–3 are unaffected. I have since re-verified finding 1 against current `main` (`bfc2b9ab` — my earlier baseline was a stale local checkout): a `views.yml` using a `<<:` merge key still migrates cleanly there and is still rejected by this branch, so the regression stands as described.

goldmedal:
## Re-review — `38309b78`

All three findings from the previous round are fixed, and verified so:

| | |
|---|---|
| Merge keys (`<<:`) | ✅ migrate again — `views/a/metadata.yml` written |
| Anchor-holder root keys | ✅ allowed; unreferenced root keys still rejected |
| Colliding targets | ✅ blocked, exact **and** case-only, extended to models and cubes |

The anchor solution is careful work — walking the composed node graph to test whether a root key's value is actually aliased inside `views`, with a cycle guard, is the right way to distinguish "holder for an anchor that gets expanded" from "real data that would be dropped". `test_apply_upgrade_v1_to_v2_preserves_merge_override` shows the merge-override semantics were thought through too: the shadowed `SELECT 1` is genuinely not part of the document's effective value, so letting it go is correct rather than a hole.

### 🔴 CI is red — two one-line fixes

**1. `lint` — `ruff format --check src/`** (the job only lints `src/`, per `wren-ci.yml`). Three spots in `context.py`, verbatim from the job log:

- over-wrapped `if` in `_v1_views_anchor_root_keys` → one line
- missing second blank line before `_validate_upgrade_view_statement`
- over-wrapped `_record_unique_upgrade_target(seen_cube_targets, target, "cube files")` call

`ruff format src/` fixes all three.

**2. `unit tests` — `NameError: name 'yaml' is not defined`** at `test_context.py:1707`, in the new `test_apply_upgrade_v1_to_v2_preserves_merge_override`. `test_context.py` has no module-level `yaml`; existing tests import it function-locally (`:923`, `:945`) or as `_yaml` (`:2317`). Adding `import yaml  # noqa: PLC0415` inside the test matches the file's convention.

Applying just those two locally: `ruff format --check src/` and `ruff check src/` clean, `tests/unit/test_context.py` 202 passed.

### 🟡 Residual — the collision guard closes case, not Unicode normalization

Same failure mode one axis over. `casefold()` makes `Revenue`/`revenue` equal, but a case-insensitive filesystem is usually normalization-insensitive too:

```yaml
views:
  - name: café      # NFC:  caf\xc3\xa9
    statement: SELECT 1
  - name: café      # NFD:  cafe\xcc\x81
    statement: SELECT 2
```

```
-> MIGRATED (no error). one directory on disk, holding SELECT 2. views.yml gone.
```

Distinct Python strings, distinct casefold keys, one file on APFS. Keying on `unicodedata.normalize("NFC", target).casefold()` closes it — I applied that one line and confirmed the NFD case blocks while all four earlier collision cases still block correctly.

Worth being explicit that this is a heuristic either way: no string comparison can fully predict a given filesystem's equivalence rules. Two axes covered is materially better than one, and this is a low-probability case (non-ASCII view names with mixed normalization), so entirely reasonable to defer — but it is the same class as the finding it descends from, so better decided than discovered.

Minor: when the two names normalize together, the message renders them identically — `multiple legacy views map to 'views/café/metadata.yml' and 'views/café/metadata.yml'`. Using `repr()` on the two targets would make that legible.

### 🟢 Minor — the apply-side cube guard is now inconsistent

`_plan_v1_to_v2` routes all three collections through `_record_unique_upgrade_target` (casefold `dict`), but `_apply_v1_to_v2` still keeps its own `seen_cube_targets: set[str]` with exact-match comparison. Not a bug today — `_apply_v1_to_v2` calls `_plan_v1_to_v2` first, so the stricter guard always fires — but it leaves a weaker duplicate of the same rule that would silently become the only one if that call were ever dropped. Routing it through the same helper would remove the divergence.

### 📝 Notes

- This now overlaps **#2696**, which adds exact-match view dedupe only. This PR is a strict superset (case-insensitive, plus models and cubes), so #2696 becomes redundant if this lands first — worth deciding the order rather than merging both.
- The branch is 15 commits behind `main` (`bfc2b9ab`) and `mergeStateStatus` is `BLOCKED`. A rebase before merge would also re-run CI against current `main`.

---

Verified against `main` at `bfc2b9ab` and PR head `38309b78`. Both CI failures reproduce locally; the two one-line fixes above were applied locally and confirmed to give clean lint and 202 passed.

goldmedal:
Thanks @NNoradrenaline, due to #2696 merge, there are some conflicts. There are some lint failures, too.

goldmedal:
## Follow-up after #2696 merged (`992f1da2`)

First, the important part: **the core of this PR is untouched by that merge.** #2696 landed exact-plus-case-insensitive collision guards for models and views only. The entire malformed-views preflight here — shape checks, root keys, duplicate YAML keys, merge-key handling, the anchor-node analysis — has no counterpart on `main`. What changed is that the collision half of this branch went from "new capability" to "improvement plus the missing cube case", not that it became redundant.

Four things outstanding.

### 1. CI is still red — two one-line fixes

Nothing has been pushed since `38309b78`, so both failures from my last comment still stand:

- **`lint`** — `ruff format src/` fixes all three spots in `context.py` (over-wrapped `if` in `_v1_views_anchor_root_keys`, missing second blank line before `_validate_upgrade_view_statement`, over-wrapped `_record_unique_upgrade_target(...)` call in the cube loop).
- **`unit tests`** — `NameError: name 'yaml' is not defined` at `test_context.py:1707` in the new `test_apply_upgrade_v1_to_v2_preserves_merge_override`. The file has no module-level `yaml`; existing tests import it function-locally (`:923`, `:945`) or as `_yaml` (`:2317`). A local `import yaml  # noqa: PLC0415` matches the convention.

Applying just those two locally gives clean `ruff format --check src/` / `ruff check src/` and 202 passed.

### 2. Rebase onto `992f1da2` — 4 conflict hunks, all expected

I trial-rebased `38309b78` onto the new `main`. Every conflict is in the model and view loops of `_plan_v1_to_v2`, and every one is the same shape:

```
<<<<<<< main (from #2696)
    seen_model_targets: set[str] = set()
=======
    seen_model_targets: dict[str, str] = {}
>>>>>>> this branch
```

Resolve by keeping **this branch's** `_record_unique_upgrade_target` form in all four. It is a strict superset of what landed: same case-folding, but it also covers cubes, names both colliding targets in the message instead of only the second, and carries the rule once instead of three times.

### 3. What this branch still uniquely adds after the rebase

| | on `main` now |
|---|---|
| malformed-views preflight | absent |
| **cube case guard** | absent — `_plan_v1_to_v2:1922` **and** `_apply_v1_to_v2:2038` are both still exact-match |
| error naming both colliding targets | names only one |
| the rule as one helper | inlined three times, one copy divergent |

Cubes are the notable one: after #2696, models and views are case-guarded and cubes are not, so a case-only cube collision still deletes both source files and keeps one. This branch already closes that.

### 4. Two findings from my last review still open

- 🟡 **NFC/NFD.** `café` (NFC) and `café` (NFD) are distinct strings with distinct casefold keys but one directory on APFS, so that collision is still silently lossy. `unicodedata.normalize("NFC", target).casefold()` as the key closes it — I verified that exact change on this branch: the NFD case blocks and all four earlier collision cases still behave.
- 🟢 **`_apply_v1_to_v2` still keeps its own `seen_cube_targets: set[str]`** with exact-match comparison, now diverging from the plan-side helper. Harmless while `_apply_v1_to_v2` calls `_plan_v1_to_v2` first, but it is a weaker duplicate of the same rule.

### An alternative worth considering

You could also drop the collision commit (`2bb23c3e`) entirely and keep only the preflight. The rebase then has almost nothing to resolve, the PR returns to its original single purpose, and the cube case gap becomes a separate issue.

The trade-off is that the cube gap and the message/helper improvements get deferred. But given this branch is currently red, 15 commits behind, and carrying two unrelated bodies of work, narrowing it back to one purpose may well be the faster route to merge. Either way is reasonable — worth a deliberate choice rather than defaulting to the larger rebase.

---

Verified against `main` at `992f1da2` and branch head `38309b78`: both CI failures reproduce locally, the rebase conflict count and shape are from an actual trial rebase, and the NFC fix and the two CI fixes were each applied locally and confirmed.

AmirF194 (commented):
I compared this branch against current `main`, not this repo's CI, since the question here is about rebase conflicts rather than correctness of the diff in isolation.

This branch's last commit is `38309b78` (2026-08-31T04:57:36Z). PR #2696 merged into `main` at 2026-08-31T05:55:24Z, about an hour later, and it already added case-folded collision checks for model and view targets inline in `_plan_v1_to_v2` (`seen_model_targets` / `seen_view_targets`, both keyed on `target.casefold()`). This PR reintroduces the same checks through a new `_record_unique_upgrade_target()` helper, so a rebase will conflict in `_plan_v1_to_v2` around the model and view loops:

```
$ gh api repos/Canner/WrenAI/contents/core/wren/src/wren/context.py --jq '.content' | base64 -d \
  | sed -n '1855,1880p'
    seen_model_targets: set[str] = set()
    ...
        if target.casefold() in seen_model_targets:
            raise UpgradeError(
                f"Cannot upgrade: multiple legacy models map to '{target}'"
            )
        seen_model_targets.add(target.casefold())
```

One thing worth keeping from this PR's version rather than dropping outright: `_record_unique_upgrade_target()` is shared across models, views and cubes, so it also casefolds the cube check. Main's cube check is still a plain `target in seen_cube_targets` with no casefold (confirmed at the same file, line 1922), a gap #2696 left open on purpose as out of scope for that PR.

If you rebase onto current main, the model/view portions of this diff are redundant with what's already there, but `_reject_malformed_v1_views()` (the actual fix for #2687) is not on main at all and is still needed. Might be worth splitting this into just the malformed-views preflight plus, if you want it, extending the cube loop to `target.casefold()` the same way models/views already do.

I have not run this branch; this is a static read against current main's source, not a Docker repro.

goldmedal:
@NNoradrenaline, to clear my review queue, I'll convert this PR to a draft. Feel free to request a review after addressing the comments. Thanks.

PR #2692, opened 2026-08-21
fix(langchain): forward session properties to the engine

### LINKED ISSUE #2691: wren-langchain: session properties are not forwarded, same gap as #2638
# wren-langchain: session properties are not forwarded, same gap as #2638

## Summary

`WrenToolkit.query()` / `dry_plan()` / `dry_run()` in `sdk/wren-langchain` never
accept or forward session properties to the engine. `wren.engine.WrenEngine`
already threads `properties` through its whole planning path, so any project
with row-level access control (RLAC) enabled cannot be queried through this
SDK at all: planning fails with a missing-session-property error, and there
is no parameter to supply one.

This is the same defect reported in #2638 for `wren-pydantic`; the two
toolkits share the same method signatures and the gap is identical in both.

## Root cause

- `core/wren/src/wren/engine.py`
  - `dry_plan(self, sql, properties=None)`
  - `query(self, sql, limit=None, properties=None)` calls `self.dry_plan(sql, properties)`
  - `dry_run(self, sql, properties=None)` same shape
- `sdk/wren-langchain/src/wren_langchain/_toolkit.py`
  - `query(self, sql, limit=None)` calls `engine.query(sql, limit=limit)`, no `properties`
  - `dry_plan(self, sql)` calls `engine.dry_plan(sql)`, same gap
  - `dry_run(self, sql)` same

## Repro

```python
toolkit = WrenToolkit.from_project("./my_project")
toolkit.query("SELECT * FROM orders")   # no `properties` kwarg available
```

With an RLAC rule requiring `session_user_id`, this always fails with a
missing-session-property planning error, and there is no way to pass one
through the toolkit.

## Fix

Opening a PR now that adds the `properties` kwarg to all three methods and
forwards it to the engine, mirroring the fix shape already proposed for
`wren-pydantic` in #2684.

AmirF194:
## What changes

`WrenToolkit.query()`, `dry_plan()` and `dry_run()` now accept `properties`
and forward it to the engine, so a caller can supply MDL session properties
per call, the same fix already needed on the `wren-pydantic` side (#2684).

`WrenEngine.query/dry_plan/dry_run` already accept `properties`
(`core/wren/src/wren/engine.py`); the langchain toolkit simply never passed
it through, so a model guarded by row-level access control could not be
read through this SDK at all: planning fails when a required session
property is missing, and there was no way to supply one.

Fixes #2691 (companion issue for wren-langchain; see #2638 for the identical wren-pydantic gap)

## Why the LangChain tools are unchanged

`wren_query` / `wren_dry_plan` keep their current signatures on purpose.
Turning a session property into a model-fillable tool argument would let the
LLM choose the identity that RLAC is keyed on, which is a separate design
question (toolkit- or run-scoped binding), so this PR only opens the direct
Python API.

## How this was verified

- 3 new tests assert `properties` reaches `WrenEngine.query/dry_plan/dry_run`.
- 3 existing delegation tests updated for the new `properties=None` default
  call contract.
- Confirmed the new tests fail against the unmodified toolkit (6 failed) and
  pass with the fix (6 passed), same Docker image both times.
- Full suite: `pytest -v` on Python 3.11 and 3.12 (this module's CI matrix),
  119 passed / 2 deselected (the `slow` marker) on both. `ruff check .` and
  `ruff format --check .` clean on both.
- Not verified: an actual RLAC-protected project end to end. The engine-side
  behavior is the existing, already-tested path; this PR only closes the gap
  in what the toolkit forwards to it.

## Summary by CodeRabbit

* **New Features**
  * Added optional session properties to query, plan, and validation operations.
  * Supports row-level access-control predicates through session properties.

* **Tests**
  * Added coverage for default and custom session properties across query, planning, and validation workflows.

### DIFF
--- a/sdk/wren-langchain/src/wren_langchain/_toolkit.py
+++ b/sdk/wren-langchain/src/wren_langchain/_toolkit.py
@@ -58,24 +58,42 @@ class WrenToolkit:
 
     # ── Direct Python API ──────────────────────────────────────────────────
 
-    def query(self, sql: str, limit: int | None = None) -> pa.Table:
-        """Execute SQL through the Wren context layer. Returns a pyarrow Table."""
+    def query(
+        self,
+        sql: str,
+        limit: int | None = None,
+        properties: dict[str, Any] | None = None,
+    ) -> pa.Table:
+        """Execute SQL through the Wren context layer. Returns a pyarrow Table.
+
+        ``properties`` carries MDL session properties and is forwarded to the
+        engine's planning path. A model guarded by row-level access control
+        needs the value its rule declares required, so without it planning
+        fails and the model cannot be read through this method at all.
+        """
         engine = self._build_engine()
         try:
-            result = engine.query(sql, limit=limit)
+            result = engine.query(sql, limit=limit, properties=properties)
         finally:
             self._connector_cache = engine._connector
         return result
 
-    def dry_plan(self, sql: str) -> str:
-        """Plan SQL through MDL and return the expanded SQL in target dialect."""
-        return self._build_engine().dry_plan(sql)
+    def dry_plan(self, sql: str, properties: dict[str, Any] | None = None) -> str:
+        """Plan SQL through MDL and return the expanded SQL in target dialect.
 
-    def dry_run(self, sql: str) -> None:
-        """Validate SQL by planning and asking the DB to plan it without executing."""
+        ``properties`` carries MDL session properties (see :meth:`query`); RLAC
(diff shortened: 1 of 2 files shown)


AmirF194:
No rush, just checking in on this one. CI is green and it has been open a week with no review yet, happy to rebase or adjust scope if that helps.

goldmedal (approved):
Thanks @AmirF194 👍

AmirF194:
Thanks for the quick review and merge, appreciate it.

PR #2696, opened 2026-08-22
fix(context): reject v1 view upgrades that map two views to the same target

### LINKED ISSUE #2695: v1 to v2 project upgrade silently discards a view when two views.yml entries share the same name
**Describe the bug**

`plan_upgrade`/`apply_upgrade` (v1 to v2 project upgrade) silently discards a view when two entries in `views.yml` share the same `name`. `_apply_v1_to_v2` resolves each view's target directory from `name` alone and writes both views into it, so the second write clobbers the first, and `views.yml`, the only other copy, is deleted right after. The upgrade exits 0 with one view's definition gone.

The same class of bug already has a guard for cubes (`seen_cube_targets` in `_plan_v1_to_v2`, added when duplicate cube files were found to collide the same way). Views never got the equivalent check.

**To Reproduce**

Given a v1 project with `views.yml`:

```yaml
views:
  - name: summary
    statement: SELECT 'first' AS marker
  - name: summary
    statement: SELECT 'second' AS marker
```

Run the v1 to v2 upgrade (`wren upgrade` CLI, or `plan_upgrade`/`apply_upgrade` directly). It succeeds. `views/summary/metadata.yml` ends up containing only the second view's statement, and `views.yml` no longer exists, so the first view's definition is gone with no error or warning at any step.

**Expected behavior**

The upgrade should reject the project up front, the same way it already does for two legacy cube files that resolve to the same target (`Cannot upgrade: multiple legacy cube files map to '...'`), rather than silently dropping data.

**Additional context**

Traced this in `core/wren/src/wren/context.py`. Have a fix ready mirroring the existing cube guard, will open a PR referencing this issue.

AmirF194:
Fixes #2695

**Root cause**

`_plan_v1_to_v2`'s cube loop already refuses to upgrade when two legacy cube files resolve to the same target directory (`seen_cube_targets`, raises `UpgradeError` before any write). The views loop right above it has no equivalent check: it resolves each view's target purely from `view.get("name")`, so two `views.yml` entries sharing the same `name` map to the same `views/<name>/` directory. `_apply_v1_to_v2` then writes both views into that directory (the second write overwrites the first) and deletes `views.yml`, the only other copy, right after.

**Fix**

Mirror the existing cube guard for views: track resolved `metadata.yml` targets in a `seen_view_targets` set while planning, and raise `UpgradeError` on a repeat before `_apply_v1_to_v2` performs any filesystem write (it calls `_plan_v1_to_v2` first for exactly this reason).

**Verification**

- New test `test_plan_upgrade_v1_to_v2_rejects_duplicate_view_names` (`tests/unit/test_context.py`), mirroring the existing `..._rejects_duplicate_cube_targets` test: fails on main (`DID NOT RAISE UpgradeError`) and passes on this branch, confirmed both ways in the same Docker image.
- Also reproduced end to end outside the test suite: built a v1 project with two `views.yml` entries named `summary`, ran `plan_upgrade`/`apply_upgrade` on main and got a silently clobbered `views/summary/metadata.yml` with `views.yml` deleted; same input on this branch raises `UpgradeError` before any file is touched.
- Full `tests/unit/` suite (excluding `test_memory.py`/`test_mcp_server.py`, matching the CI job's own scope): 1191 passed, 2 skipped.
- `ruff format --check` and `ruff check` clean on both changed files.
- Not run: the connector/UI/memory/mcp CI jobs, since this change touches neither their code paths nor their extras.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented v1-to-v2 upgrade plans from proceeding when multiple legacy views or models target the same destination.
  * Added detection for name conflicts that differ only by letter case.
  * Ensured failed upgrade plans leave the schema version and source files unchanged, without creating partial destination files or directories.

* **Tests**
  * Added regression coverage for duplicate view and model mappings, including case-insensitive conflicts and prevention of partial migrations.

### DIFF
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -1705,6 +1705,7 @@ def _plan_v1_to_v2(project_path: Path) -> tuple[list[str], list[str]]:
     _reject_malformed_v1_model_columns(project_path)
 
     # Models: flat files → directories
+    seen_model_targets: set[str] = set()
     models = _load_models_v1(project_path)
     for model in models:
         source_dir = model.pop("_source_dir", None)
@@ -1717,7 +1718,17 @@ def _plan_v1_to_v2(project_path: Path) -> tuple[list[str], list[str]]:
             created.append(ref_sql_file.relative_to(project_root).as_posix())
 
         metadata_file = _resolve_upgrade_file(model_dir, "metadata.yml")
-        created.append(metadata_file.relative_to(project_root).as_posix())
+        target = metadata_file.relative_to(project_root).as_posix()
+        # Case-folded: two names differing only in case resolve to the same
+        # directory on a case-insensitive filesystem (macOS APFS default,
+        # Windows), so treat them as the same target everywhere, not just on
+        # a literal string match.
+        if target.casefold() in seen_model_targets:
+            raise UpgradeError(
+                f"Cannot upgrade: multiple legacy models map to '{target}'"
+            )
+        seen_model_targets.add(target.casefold())
+        created.append(target)
 
         if source_dir:
             deleted.append(f"models/{source_dir}.yml")
@@ -1738,11 +1749,12 @@ def _plan_v1_to_v2(project_path: Path) -> tuple[list[str], list[str]]:
 
         metadata_file = _resolve_upgrade_file(view_dir, "metadata.yml")
         target = metadata_file.relative_to(project_root).as_posix()
-        if target in seen_view_targets:
+        # Case-folded for the same reason as the model guard above.
+        if target.casefold() in seen_view_targets:
             raise UpgradeError(
                 f"Cannot upgrade: multiple legacy views map to '{target}'"
             )
-        seen_view_targets.add(target)
+        seen_view_targets.add(target.casefold())
 
         created.append(target)
 
--- a/core/wren/src/wren/context.py
+++ b/core/wren/src/wren/context.py
@@ -1723,6 +1723,7 @@ def _plan_v1_to_v2(project_path: Path) -> tuple[list[str], list[str]]:
             deleted.append(f"models/{source_dir}.yml")
 
     # Views: single file → directories
+    seen_view_targets: set[str] = set()
     views = _load_views_v1(project_path)
     for view in views:
         name = view.get("name")
@@ -1736,7 +1737,14 @@ def _plan_v1_to_v2(project_path: Path) -> tuple[list[str], list[str]]:
             created.append(sql_file.relative_to(project_root).as_posix())
 
         metadata_file = _resolve_upgrade_file(view_dir, "metadata.yml")
-        created.append(metadata_file.relative_to(project_root).as_posix())
+        target = metadata_file.relative_to(project_root).as_posix()
+        if target in seen_view_targets:
+            raise UpgradeError(
+                f"Cannot upgrade: multiple legacy views map to '{target}'"
+            )
+        seen_view_targets.add(target)
+
+        created.append(target)
 
     views_file = project_path / "views.yml"
     if views_file.exists():
--- a/core/wren/tests/unit/test_context.py
+++ b/core/wren/tests/unit/test_context.py
@@ -1495,6 +1495,111 @@ def test_plan_upgrade_v1_to_v2_rejects_duplicate_view_names(tmp_path):
         ) == source_contents[relative_path]
 
 
+def test_plan_upgrade_v1_to_v2_rejects_case_insensitive_duplicate_view_names(
+    tmp_path,
+):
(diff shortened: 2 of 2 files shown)


AmirF194:
Checking in on this one, no rush. Pushed a small test strengthening per CodeRabbit's suggestion (assert views.yml content is unchanged after the rejected upgrade, not just that the file exists).

goldmedal:
## Review

The fix is correct, minimal, and lands in the right place — `_plan_v1_to_v2`, before any write, mirroring the existing cube guard, with `_apply_v1_to_v2` inheriting it via its own `_plan_v1_to_v2` call. The test's source-preservation assertions (schema version, `views.yml` byte-identical, no `views/` directory, all other v1 sources untouched) are exactly the right shape for this bug class.

Verified independently against current `main` (`bfc2b9ab`, not the branch's 13-commits-older base): the new test fails there with `DID NOT RAISE UpgradeError` and passes on this branch; `tests/unit/test_context.py` is 182 passed on the branch. The PR description's verification claims check out.

Two gaps of the same class remain — both pre-existing, neither introduced here.

### 🟡 1. Case-only collisions still slip through

`seen_view_targets` keys on the literal `views/<name>/metadata.yml` string, so two names differing only in case are distinct keys — but on a case-insensitive filesystem (macOS APFS default, Windows) they are the same directory:

```yaml
views:
  - name: Revenue
    statement: SELECT 1
  - name: revenue
    statement: SELECT 2
```

```
-> MIGRATED (no error). views.yml gone: True
   views/Revenue: name: revenue | statement: SELECT 2
```

One directory, named for the first view, containing the second — and `views.yml` deleted. This is the PR's own stated failure mode ("two views map to the same target"), just resolved by the filesystem rather than by string equality. It will not reproduce on Linux CI, which is why the existing suite cannot catch it.

Keying the set on `target.casefold()` closes it in one line. That does reject a project that would migrate fine on a case-sensitive filesystem, but for a one-way migration that deletes the source, refusing an ambiguous project seems clearly preferable to data loss that depends on which machine runs the upgrade. The cube guard has the same gap, for what it's worth.

### 🟡 2. Models have the identical gap, ~15 lines above

The model loop in the same function resolves `models/<name>/` from `model.get("name")` with no dedupe, so two `models/*.yml` files sharing a `name:` collapse the same way — and here *both* source files are deleted:

```
models/a.yml (name: dup, column x)  +  models/b.yml (name: dup, column y)
-> MIGRATED. models/dup/metadata.yml has only column y
   leftover model files: []          # both sources gone
```

The same case-collision variant applies. Understandable if you would rather keep this PR scoped to views and file models separately, but it is the same three lines in the same loop, and models are the more likely collection to hit it.

### 🟢 3. Minor — guard placement asymmetry

Cubes are guarded in both `_plan_v1_to_v2` and `_apply_v1_to_v2`; views now only in plan. That is sufficient today because `_apply_v1_to_v2` calls `_plan_v1_to_v2` first, so this is not a bug — but the file already has a `..._rechecks_..._before_writing` test convention for exactly this invariant, and a duplicate-views case there would pin it against future refactors.

### 📝 Note

This is complementary to #2690, which is also open and also touches the views section of `_plan_v1_to_v2` (a raw-source preflight for malformed and nameless views). Neither subsumes the other — that one rejects shape problems, this one rejects target collisions — and they touch different hunks, so no textual conflict is expected. Worth landing both.

AmirF194:
Thanks for the careful read, this is exactly the kind of check I should have run myself before opening the PR.

Pushed a follow-up commit that folds both the view and model target guards through casefold(), so Revenue/revenue (or any case-only variant) now collides the same way an exact duplicate does, and it raises before any write. I also added the missing duplicate-target guard to the model loop itself, since the collision there deletes the source file the same way and it turned out to be the same three lines in the same style. New tests cover a case-insensitive duplicate view name and both an exact and case-insensitive duplicate model name, each asserting the source files come back byte-identical and no partial directory gets created, same shape as the existing duplicate-view test. Full tests/unit/ run at 1194 passed, 2 skipped; ruff format and check are clean on both touched files (the 11 pre-existing check findings elsewhere in test_context.py are unchanged from main).

Left the cube guard's case sensitivity alone since it predates this PR and is a separate function, happy to open a follow-up for it if that's useful. Same for item 3, the guard is only in _plan_v1_to_v2 for now, which is sufficient since _apply_v1_to_v2 calls plan first, so I skipped adding the extra pinning test to keep the diff focused, but can add it if you'd rather have it here.

Good call on #2690 being complementary, nothing here should conflict with it.

goldmedal (approved):
Thanks @AmirF194 👍

AmirF194:
Thanks for the careful review across all three rounds, appreciate you catching the case-insensitivity gap and the missing model guard. Glad it's in.

Issue #2688, opened 2026-08-21
Relationship handles are not first-class join aliases


goldmedal:
## Summary

A relationship column in MDL is a *handle*: a join named locally by the model that declares it.
`docs/core/reference/mdl.md` documents it that way — a handle `customer` on `orders` pointing at
the model `customers`.

The engine does not treat it as an alias. It uses the **related model** as the identity of the
join and discards the handle name. #2661 makes the immediate symptom go away by rewriting the
handle to the model name before planning, which is the right minimal fix under the current
architecture, but the identity is still lost one step later. Several separate symptoms follow
from that one design point, so this issue tracks the root cause rather than any of them
individually.

## Root cause

The handle name survives only as long as the MDL text. Everything downstream is keyed by model:

- `Lineage::collect_required_fields` records required fields against the related model's
  `TableReference`, and the relationship graph is `Graph<Dataset, DatasetLink>` whose node
  identity is the model itself (`Dataset::Model(Arc<Model>)`) —
  `core/wren-core/core/src/mdl/lineage.rs`, `core/wren-core/core/src/mdl/dataset.rs:11`.
- `Dataset::to_qualified_schema` hardcodes the schema qualifier to the model name
  (`core/wren-core/core/src/mdl/dataset.rs:37`), so a joined relation has no way to carry an
  alias.
- `RelationChain::with_chain` walks the graph with `find_edge(prev, next)`
  (`core/wren-core/core/src/logical_plan/analyze/relation_chain.rs:85`), which yields a single
  edge per node pair — parallel edges between the same two models are structurally unusable
  even if they were kept.

## Symptoms

1. **Fixed by #2661.** A calculated field whose handle name differs from the related model fails
   to plan (`column 'name' not found in 'customer'`).
2. **Two handles onto the same model collapse into one join — silently wrong results.** Repro
   below.
3. **Relationship traversal in user SQL fails regardless of the handle name.**
   `SELECT orders.customer.name FROM orders` has nothing to resolve against, because
   `Model::get_physical_columns` keeps only columns where `relationship.is_none()`, so the handle
   never reaches the schema a model registers with DataFusion. Reported in #2658 and analysed in
   the last section of #2661.
4. **`docs/core/reference/mdl.md:186`** states that `orders.customer.first_name` is valid SQL,
   which symptom 3 contradicts. Either the engine gains that capability or the page needs a
   caveat.

## Repro for symptom 2

[...]

PR #2697, opened 2026-08-23
fix(core): reject collapsed relationship joins

primorLee:
## Summary

- Detect different DatasetLink values when calculated-column graphs merge the same model pair.
- Return a planning error containing both join conditions instead of silently keeping whichever relationship was merged first.
- Preserve deduplication when the incoming relationship is genuinely identical.

This is the short-term containment described in #2688. It prevents plausible-but-wrong SQL while full first-class relationship aliases remain a larger architectural change.

The regression test builds bill-to and ship-to edges from orders to addresses with different join conditions, then merges the two graphs.

Before this change:

    cargo test -p wren-semantic-core --lib rejects_distinct_relationships_between_the_same_models -- --nocapture

failed with:

    distinct relationships must not collapse into one join: ()

The unit value shows merge_graph returned Ok and silently discarded the incoming ship-to edge. In the role-playing-dimension query from #2688, that causes both calculated cities to use the bill-to join.

After this change, the same graph merge returns a plan error that names both conflicting conditions.

## How is it tested?

- Targeted regression: 1 passed.
- Full core library suite: 153 passed, 0 failed.
- cargo clippy -p wren-semantic-core --all-targets -- -D warnings: passed.
- The test is a normal core library unit test and runs in the default Rust test target.

## Duplicate check

Searched open PRs for #2688, relationship handles, and distinct join aliases; the issue still has no linked PR. #976 was also reviewed and is an unrelated, service-only semantics-enrichment alias change that does not touch wren-core or relationship planning.

Addresses #2688.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved query planning when multiple relationships connect the same datasets.
  * Distinct join conditions are now preserved and reported as planning errors instead of being silently ignored.

### DIFF
--- a/core/wren-core/core/src/logical_plan/analyze/plan.rs
+++ b/core/wren-core/core/src/logical_plan/analyze/plan.rs
@@ -943,15 +943,71 @@ fn merge_graph(
         };
         let source = node_map.get(&source).unwrap();
         let target = node_map.get(&target).unwrap();
-        // Skip duplicate edges between the same pair of nodes: the same
-        // relationship may appear in multiple calc-col sub-graphs (e.g. two
-        // calc cols traversing the same relationship), and adding parallel
-        // edges would produce redundant joins downstream.
-        if graph.find_edge(*source, *target).is_none() {
+        // The same relationship may appear in multiple calculated-column
+        // sub-graphs, in which case the duplicate edge is safe to skip. Two
+        // different relationships between the same models cannot be merged,
+        // though: downstream graph traversal can use only one edge and would
+        // silently generate a join with the wrong condition.
+        if let Some(existing_edge) = graph.find_edge(*source, *target) {
+            let existing_link = &graph[existing_edge];
+            let incoming_link = &new_graph[edge];
+            if existing_link != incoming_link {
+                return plan_err!(
+                    "Multiple relationships from '{}' to '{}' require distinct join aliases; \
+                     cannot merge '{}' with '{}'",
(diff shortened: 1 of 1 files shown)


AmirF194 (commented):
I checked this against head 5eed46a in a clean container (rust:1.90-slim, RUST_MIN_STACK bumped for an unrelated pre-existing stack-overflow in test_access_model that reproduces on base too and is untouched by this diff).

Applying only the new test to base f2841bc, before the fix, confirms the regression:

    cargo test -p wren-semantic-core --lib rejects_distinct_relationships_between_the_same_models
    thread '...' panicked at core/src/logical_plan/analyze/plan.rs:992:14:
    distinct relationships must not collapse into one join: ()
    test result: FAILED. 0 passed; 1 failed

At head the same test passes, and the full core suite is clean:

    test result: ok. 1 passed; 0 failed
    test result: ok. 153 passed; 0 failed; 0 ignored

The equality check is safe against false positives too: get_dataset_link_revers_if_need builds `condition` from `rs.condition.clone()`, the same Relationship's verbatim string every time a handle is traversed, so two sub-graphs hitting the same relationship still produce a byte-identical DatasetLink and keep deduplicating. Only genuinely distinct relationships (different `condition`) hit the new error path.

goldmedal (approved):
Thanks @primorLee, look great 👍

goldmedal:
Thanks @AmirF194 for reviewing

PR #2703, opened 2026-08-27
fix(wren): forget query_history rows deleted from knowledge/sql on reindex

### LINKED ISSUE #2702: wren memory index only upserts, never forgets deletions from knowledge/sql
**Describe the bug**

`wren memory index` (LanceDB backend) treats `knowledge/sql/*.md` as upsert-only: it
never removes a `query_history` row whose natural-language query is no longer present
in the markdown. Deleting or renaming an example under `knowledge/sql/` does not clear
it from the semantic index, so it keeps being recalled as few-shot context for SQL
generation.

`wren memory check` already detects this drift and calls it out as stale, but its own
suggested fix does not work:

```
2 user pair(s) indexed without markdown — stale index, run `wren memory index`.
```

Re-running `wren memory index` only upserts the pairs currently in the markdown; it
never forgets the ones that dropped out. The same gap exists in `wren memory watch`'s
auto-reindex loop, whose own docstring says it keeps recall from serving stale
content.

**To Reproduce**

```bash
wren memory index                       # index two examples
rm knowledge/sql/count-orders.md        # delete one example
wren memory check                       # -> "1 user pair(s) indexed without markdown"
wren memory index                       # per check's own advice
wren memory recall -q "count orders"    # still returns the deleted example
```

**Expected behavior**

After deleting an example from `knowledge/sql/` and re-running `wren memory index`
(or letting `wren memory watch` pick up the change), the deleted example is dropped
from `query_history` and no longer recalled.

**Additional context**

`core/wren/src/wren/memory/store.py`'s `load_queries(pairs, upsert=True)` only
upserts rows whose `nl_query` is present in the current batch; it has no notion of
"rows that used to be there and now aren't." The three call sites that treat
`knowledge/sql/*.md` as the complete source of truth (`cli.py`'s `index` and `watch`
commands, and `index_backend.py`'s `LanceDBIndex.rebuild`) all inherit this gap, since
they all pass the current markdown pairs straight through the same upsert-only call.

AmirF194:
## Root cause

`MemoryStore.load_queries(pairs, upsert=True)` upserts every pair in the batch by
`nl_query`, but never removes a row whose `nl_query` is no longer in the batch. The
three call sites that treat `knowledge/sql/*.md` as the complete source of truth for
`query_history` (`cli.py`'s `index` and `watch` commands, and `index_backend.py`'s
`LanceDBIndex.rebuild`) all pass the current markdown pairs straight through this
upsert-only call, so a deleted or renamed example is never forgotten and keeps
surfacing in semantic recall. `check()` already computes exactly this drift (`stale =
indexed_user - md_nls`) and tells the user to fix it by running `index`, but `index`
does not actually clear it.

## Fix

Add `MemoryStore.sync_markdown_queries(pairs)`: upserts as before, then lists the
current rows and forgets any whose source is not `seed`/`view` and whose `nl_query` is
absent from `pairs`, mirroring `check()`'s own "stale" definition on the write path
instead of only the read-only report. The three call sites above now use it.

## Verification

- New regression tests in `tests/unit/test_memory.py` (`TestMarkdownSourcedIndex`):
  deleting a markdown example and re-syncing forgets the row and it no longer recalls;
  seed/view rows survive a sync even though they have no markdown file; deleting every
  markdown example forgets every markdown-sourced row; `wren memory index` and `wren
  memory watch --reindex-on-start` both forget a deleted pair end to end through the
  CLI. Each fails on unmodified `main` and passes on this branch (Docker, Python 3.11,
  `WREN_EMBEDDING_MODEL=paraphrase-MiniLM-L3-v2`).
- `pytest tests/unit/test_memory.py`: 102 passed.
- `pytest tests/unit/ --ignore=tests/unit/test_memory.py --ignore=tests/unit/test_mcp_server.py`:
  1236 passed, 3 pre-existing failures in `test_served_content_guard.py` unrelated to
  this change (confirmed identical on unmodified `main`).
- `ruff format --check src/` and `ruff check src/`: clean.
- Not run: the `postgres`/`mysql`/`ui` CI legs (unaffected by this diff) and the
  `mcp` extra's tests.

Fixes #2702

## Summary by CodeRabbit

* **Bug Fixes**
  * Markdown-based memory indexes now remove stale queries when source files are deleted.
  * Empty Markdown sources correctly clear previously indexed queries.
  * Seed, legacy, view, and manually loaded queries remain preserved during synchronization.
  * Markdown queries now correctly take precedence when overlapping with seed queries.
  * Indexing and watch reindexing consistently reflect additions, updates, and deletions.
  * Memory checks no longer incorrectly flag legacy or non-Markdown entries as stale.
  * Source-based filtering works correctly for entries with additional metadata.
  * Synchronization results report loaded, updated, and forgotten queries, including stale-query removals.
  * Legacy YAML imports are now identified consistently during export and synchronization.


AmirF194:
Good catch, thanks. sync_markdown_queries called load_queries(pairs, upsert=True), whose delete step keyed only on nl_query, not source, so a markdown pair could clobber a seed or view row with the same nl. Fixed in f075ee7: markdown pairs are now filtered against existing seed/view nl_query values before the upsert, with a regression test that reproduces the collision and fails without the fix.

goldmedal (changes requested):
**Verdict: request changes — 1 blocking.**

The bug in #2702 is real and commit `b4de068f` fixes it correctly. The follow-up commit `f075ee77` ("protect seed/view rows from a markdown-sync nl collision") introduces a worse regression than the one it fixes, and the PR's own test encodes the regression as intended behaviour.

### 🔴 Blocking — the seed-collision pre-filter permanently drops user-authored examples

`sync_markdown_queries` filters markdown pairs against existing seed/view `nl_query` values *before* the upsert. A `knowledge/sql/*.md` file whose NL matches a generated seed NL (seeds are formulaic: `List all {model}`, `Total {col} in {model}`) is then never indexed, and any previously-indexed row for it is deleted as "stale".

Reproduced against this branch (`f075ee77`), a project with one `orders` model:

```
$ wren memory index
Indexed 2 schema items, 1 seed queries.

$ wren memory store --nl "List all orders" --sql "SELECT id FROM orders WHERE status <> 'test'"
Stored: knowledge/sql/list-all-orders.md

$ wren memory index
Indexed 0 pair(s) from knowledge/sql/. Forgot 1 stale pair(s).
```

The pair a user just confirmed is deleted on the next routine reindex and reported as *stale* while its markdown file is sitting on disk. Consequences:

- **`wren memory check` enters the same unfixable loop this PR set out to fix, inverted.** It reports `1 not indexed — run 'wren memory index'` on every run, and `index` can never clear it (verified across 3 consecutive `index` runs).
- **Recall silently serves the wrong SQL under the user's own filename.** `_annotate_markdown_paths` matches on exact NL, so the surviving *seed* row gets annotated with `path = knowledge/sql/list-all-orders.md` while carrying `SELECT * FROM orders LIMIT 100` — not what that file contains.

Same scenario on `main` and on this PR's first commit `b4de068f`: `Indexed 1 pair(s) from knowledge/sql/.`, `check` → `In sync.`, recall returns the user's SQL. So the regression is entirely the 8-line pre-filter added in `f075ee77`.

`test_sync_preserves_seed_row_whose_nl_collides_with_markdown_pair` asserts `result["loaded"] == 0` and `total == 1` — i.e. it locks in the dropped markdown pair as correct, which is why the rest of the suite stays green.

**Suggested fix: drop the pre-filter (revert the `store.py` hunk of `f075ee77`).** The premise of that commit — that the upsert "silently deleted a protected row" — overstates the harm: seed rows are regenerated from the manifest by `index_schema` on every `index`/`watch` reindex, so a clobbered seed is self-healing, whereas a dropped markdown pair is permanent. Explicit user content should win over an auto-generated seed, which is what `main` does today. If seed precedence really is wanted, it needs to at minimum (a) not leave `check` in a permanent unfixable state and (b) tell the user their file was skipped.

### 🟡 The fix's own premise breaks when a legacy `queries.yml` is present

In `index`, `sync_markdown_queries` runs *before* the legacy `queries.yml` loader in the same command. Those pairs are not markdown-backed, so every run deletes and re-embeds them:

```
--- index run 2 ---
Indexed 1 pair(s) from knowledge/sql/. Forgot 1 stale pair(s).
Loaded 1 pair(s) from queries.yml (legacy) (0 skipped).
--- index run 3 ---            # identical, forever
```

`Forgot 1 stale pair(s)` is reported on every run for a pair that is not stale and is re-added two lines later, plus an unnecessary embedding round-trip each time. `check` still reports the drift afterwards, so the claim that `index` now clears what `check` reports does not hold here. Consider running the sync *after* the legacy load, feeding the yml pairs into the sync's current set, or tagging them `source:legacy` and adding that to `_NON_MARKDOWN_SOURCES`.

### 🔵 Minor

1. **`_tag_source` duplicates `cli._parse_source` verbatim**, and `_NON_MARKDOWN_SOURCES` duplicates `check()`'s inline `("seed", "view")`. The whole design rests on the write path mirroring `check()`'s read path exactly — share one helper and one constant so they cannot drift.
2. **`wren memory load` (YAML import) becomes ephemeral.** Its rows are not markdown-backed, so the next `index`/`watch` silently deletes them. Defensible under "markdown is the source of truth", but it is currently undocumented and silent for a supported command — worth a note in `docs/cli.md` and/or a warning.
3. **Performance**: a sync now materialises the whole `query_history` table via `to_pandas()` 4–5 times (two `list_queries(limit=1_000_000)` + `_existing_pairs_index` + `forget_queries_by_ids`). `watch` runs this on every detected change; the stale ids could come from a single snapshot.
4. `limit=1_000_000` as an "all rows" idiom silently truncates past 1M (same idiom as `check`); an explicit no-limit path would be clearer.
5. `sync_markdown_queries` rebinds its `pairs` parameter — minor readability.
6. When only `forgotten` is non-zero the message reads `Indexed 0 pair(s) from knowledge/sql/. Forgot N stale pair(s).` — slightly awkward phrasing.
7. `LanceDBIndex.rebuild()` has no production callers (tests only), so its return-shape change is safe; the PR description's "three call sites" is really two live ones.
8. The branch is based on `56e007da`, ~15 commits behind `main` (`4000bea0`). Still mergeable, but worth a rebase.

**Verification performed**: `pytest tests/unit/test_memory.py` → 103 passed; `ruff format --check src/` and `ruff check src/` clean; behavioural repro of findings 1 and 2 run against `main`, `b4de068f`, and `f075ee77` (Python 3.11, `WREN_EMBEDDING_MODEL=paraphrase-MiniLM-L3-v2`).

AmirF194:
You are right about the blocking issue. Reverted the `store.py` hunk from `f075ee77` as suggested: the pre-filter is gone, a markdown pair now wins over a colliding seed row again, and `test_sync_preserves_seed_row_whose_nl_collides_with_markdown_pair` is rewritten as `test_sync_lets_a_markdown_pair_win_over_a_colliding_seed_row` to assert that instead of the regression.

Also fixed the queries.yml ordering problem: `index()` now tags legacy pairs `source:legacy` and `_NON_MARKDOWN_SOURCES` includes `legacy`, so `sync_markdown_queries` stops deleting and re-embedding them every run. `check()` had its own hardcoded `("seed", "view")` tuple for the same exclusion, so a legacy pair would still have read as permanent drift there even after this fix; it now imports the shared constant. Added a CLI test that runs `index` then `check` twice on a project with only a legacy `queries.yml` and asserts `In sync.` both times, plus a store-level test for the tagging itself.

Took your minor point 1 too: `cli._parse_source` now delegates to `store._tag_source` instead of duplicating it.

Left the rest of the minor list (`to_pandas()` call count, the `limit=1_000_000` idiom, `load`'s ephemeral-row behavior, the message phrasing, `pairs` rebinding) as filed, not fixed: none of them change behavior, and bundling them in risked losing the actual regression fix in more diff to review.

Rebased onto current `main` and re-ran `pytest tests/unit/test_memory.py tests/unit/test_memory_watch.py tests/unit/test_index_backend.py tests/unit/test_memory_markdown.py` (142 passed) plus `ruff format --check` / `ruff check` on `src/` and the touched test file, all clean, before pushing.

goldmedal (changes requested):
**Verdict: request changes — 1 blocking.**

The root-cause analysis is right, and the markdown-backed path works — verified end to end (`wren memory store` → `wren memory index` keeps the row). The earlier seed pre-filter regression is correctly reverted in `b4c1a980`. The blocking problem is the *scope* of the new delete.

### 🔴 Blocking — `sync_markdown_queries` deletes rows that were never markdown-backed

`sync_markdown_queries` forgets every row whose source is not in `_NON_MARKDOWN_SOURCES = {seed, view, legacy}` and whose `nl_query` is absent from `knowledge/sql/`. But `source:user` is not a synonym for "came from markdown" — it is just the default source for *any* pair that doesn't say otherwise. Two live paths produce `source:user` rows with no markdown file, and both now lose data on the next `wren memory index`.

**1. `wren memory load <file>.yml`** — a documented, non-deprecated command that writes no markdown and defaults `source` to `"user"` (`sources = Counter(p.get("source", "user") ...)` → `load_queries` → `tags = "source:user"`).

```
# this branch
$ wren memory load pairs.yml
Loaded 1 pair(s) (1 new).
$ wren memory index
Indexed 0 pair(s) from knowledge/sql/. Forgot 1 stale pair(s).
$ wren memory list
No pairs found.

# identical steps on main (8d23753e)
$ wren memory index
Indexed 0 schema items.
$ wren memory list
Revenue by region | SELECT region, SUM(amount) FROM o GROUP BY region | source:user
Showing 1-1 of 1 pairs.
```

The YAML file may live anywhere, or have been deleted after the import — there is nothing for a later `index` to restore from.

**2. Existing users whose `queries.yml` has already been consumed.** The new `source: legacy` tag only helps while `queries.yml` is still on disk. A user who imported it on an older version (rows land as `source:user`) and then removed the file loses those rows on their *first* `index` after upgrading:

```
# index on main with queries.yml present  → row stored as source:user
# user removes queries.yml, upgrades, runs index on this branch:
Indexed 0 pair(s) from knowledge/sql/. Forgot 1 stale pair(s).
$ wren memory list
No pairs found.
```

The in-place upgrade path where `queries.yml` *is* still present does self-heal (forget → reload as `source:legacy`; second run clean, `check` reports "In sync.") — I verified that separately. It is only the consumed-and-removed case that is unrecoverable, and that is the expected end state for a file the code itself describes as transitional.

Worth naming explicitly: `check()` previously only *reported* these as drift. Turning that report into an unconditional delete is the actual change in blast radius, and neither the PR description nor the CLI output warns that anything is being destroyed permanently.

Suggested directions, in order of preference:

- **Track provenance instead of inferring it.** Tag rows written by the markdown sync distinctly (e.g. `source:markdown`, or record the originating `path` — `load_query_pairs` already returns it) and forget only rows carrying that provenance. "Don't delete what you didn't write" removes the whole class of bug rather than enumerating exemptions.
- **Minimal stopgap** (still leaves a hole for an explicit `source: user` inside a loaded YAML): default `wren memory load` to a non-markdown source and add it to `_NON_MARKDOWN_SOURCES`.
- Either way, consider making the destructive step opt-in for one release (`index --prune`, or list what would be dropped and require a flag), since existing indexes contain `source:user` rows of unknowable provenance.

Verification behind the above: real CLI runs on this branch and on `main` with `WREN_MEMORY_BACKEND=lancedb` and `WREN_EMBEDDING_MODEL=paraphrase-MiniLM-L3-v2`; `pytest tests/unit/test_memory.py` → 105 passed on this branch.

AmirF194:
You're right. source:user isn't evidence of markdown backing, it's also what wren memory load defaults to and what a pre-source:legacy queries.yml import already carried.

Went with your first option. sync_markdown_queries's own upsert now tags every row it writes with an explicit provenance marker separate from the source: tag, and the forget scan is scoped to that marker instead of the seed/view/legacy exclusion list, so it only ever deletes a row it wrote itself. wren memory load rows and pre-upgrade legacy imports never carry it.

That marker broke list_queries/count_queries_by_source/dump_queries/forget_queries_by_source, which matched tags by exact string equality against "source:X" and would have gone blind to every markdown-synced row with --source user. Fixed all four to parse the source token instead.

Skipped the opt-in-for-a-release suggestion: the provenance fix already narrows the delete to what this method wrote, which was the actual bug, so I left that bigger change for you to weigh in on.

Tested: reproduced your wren memory load repro on unpatched code (same "Forgot 1 stale pair(s)" then "No pairs found"), confirmed it survives on the fix. Full tests/unit/test_memory.py: 110 passed, including the seed-collision test unchanged. Added coverage for a view row, the load scenario, a pre-upgrade queries.yml row with the file gone, and --source filtering against a tagged row. ruff clean, CI green.

Not checked: wren memory check still infers "not markdown-backed" the old way, so it can call a load row stale even though index is now a no-op for it. Left that for a follow-up. Also haven't tried this outside Linux.

core/wren/src/wren/memory/cli.py
```diff
@@ -532,16 +541,19 @@ def check(
...
+    # seed/view/legacy rows are derived or externally sourced (same set
+    # sync_markdown_queries excludes from its own stale scan).
     indexed_user = {
         r.get("nl_query")
         for r in indexed
-        if _parse_source(r.get("tags")) not in ("seed", "view")
+        if _parse_source(r.get("tags")) not in _NON_MARKDOWN_SOURCES
     }
 
     # Compare user pairs only — seed/view rows aren't markdown-backed.
```
goldmedal: **Blocking 1 — this is the regression.** `check()` infers markdown-backing from the `source:` name, but the write path records it with `_MARKDOWN_SYNC_TAG`. Two definitions of one property, and they disagree in both directions.

`wren memory export` preserves each row's source into the frontmatter, so with the new `source:legacy` a migrated pair lands in `knowledge/sql/` as `source: legacy`. It is markdown-backed like any other file, but `legacy` is excluded here, so it counts as `missing` — forever:

```
main:  knowledge/sql: 1 pair(s); index: 1 pair(s)   In sync.
PR:    knowledge/sql: 1 pair(s); index: 1 pair(s)     1 not indexed — run `wren memory index`.
```
(`queries.yml` → `index` → `export` → `index` → `check`; the row is `'source:legacy origin:markdown-sync'`, i.e. correctly synced. Repeating `index` never clears it.)

The same mismatch is why a `wren memory load` pair reads as permanently `stale` while `index` provably cannot clear it — the very loop this PR set out to close. Judging by the provenance tag fixes both at once:

```suggestion
    from wren.memory.store import _MARKDOWN_SYNC_TAG, _has_tag  # noqa: PLC0415

    md_nls = {p["nl"] for p in load_query_pairs(project)}
    mem_store = idx.store
    indexed, _ = mem_store.list_queries(limit=1_000_000)
    indexed_nls = {r.get("nl_query") for r in indexed}
    # Markdown-backed means "written by sync_markdown_queries", which is what
    # the provenance tag records — the same predicate that method's own forget
    # pass uses. Deriving it from `source` instead would put this report out of
    # step with the write path in both directions: a `source:legacy`/`view`
    # pair exported into knowledge/sql/ would read as permanently "not
    # indexed", and a `wren memory load` pair would read as permanently
    # "stale" even though no `index` run is able to clear it.
    indexed_user = {
        r.get("nl_query")
        for r in indexed
        if _has_tag(r.get("tags"), _MARKDOWN_SYNC_TAG)
    }

    # markdown vs. the rows the sync derived from it
```

Apply together with the `_NON_MARKDOWN_SOURCES` deletion in `store.py` (this drops its only consumer).

I kept the name `indexed_user` so the two lines just below still resolve — a suggestion can't reach past the hunk. Worth renaming to `indexed_md` yourself, refreshing those two trailing comments, and changing the message at line 573 to `indexed pair(s) without markdown` since it no longer means "user" pairs.

goldmedal (commented):
Reviewed against `origin/main` (4d167fc0) vs head 79980f10, with the memory extras installed and the CLI exercised end to end. The root cause is correctly identified and `sync_markdown_queries` is the right home for the fix — the headline behaviour works (`Indexed 1 pair(s) from knowledge/sql/. Forgot 1 stale pair(s).`, and the deleted pair stops surfacing in `recall`).

Two blocking issues, both from one structural gap: **the write path records markdown provenance via `_MARKDOWN_SYNC_TAG`, while `check()` still infers it from the `source:` name.** One is a user-visible regression on the documented `wren memory export` migration path; the other silently widens what `wren memory forget --source user` deletes. Details and suggested changes are inline — five of them compose into one coherent end state, so please apply them together.

Two regression tests are still missing; these both fail on this branch as it stands and pass with the inline changes applied:

```python
    def test_cli_export_then_index_reports_in_sync(self, tmp_path, monkeypatch):
        """`wren memory export` is the documented one-time migration, and it
        preserves each row's source into the markdown frontmatter -- so a
        legacy queries.yml pair lands in knowledge/sql/ as `source: legacy`.
        That file is markdown-backed like any other, so `check` must not read
        it as "not indexed".
        """
        pytest.importorskip("lancedb", reason="wren[memory] extras not installed")
        pytest.importorskip(
            "sentence_transformers", reason="wren[memory] extras not installed"
        )
        from typer.testing import CliRunner  # noqa: PLC0415

        from wren.cli import app  # noqa: PLC0415

        monkeypatch.setenv("WREN_PROJECT_HOME", str(tmp_path))
        monkeypatch.setenv("WREN_MEMORY_BACKEND", "lancedb")
        (tmp_path / "target").mkdir()
        (tmp_path / "target" / "mdl.json").write_text("{}", encoding="utf-8")
        (tmp_path / "queries.yml").write_text(
            "pairs:\n  - nl: Total revenue\n    sql: SELECT SUM(amount) FROM o\n",
            encoding="utf-8",
        )

        cli = CliRunner()
        assert cli.invoke(app, ["memory", "index"]).exit_code == 0
        export = cli.invoke(app, ["memory", "export"])
        assert export.exit_code == 0, export.output
        assert "source: legacy" in next(
            (tmp_path / "knowledge" / "sql").glob("*.md")
        ).read_text(encoding="utf-8")

        for _ in range(2):
            index_result = cli.invoke(app, ["memory", "index"])
            assert index_result.exit_code == 0, index_result.output
            check_result = cli.invoke(app, ["memory", "check"])
            assert check_result.exit_code == 0, check_result.output
            assert "In sync." in check_result.output

    def test_source_filters_ignore_rows_carrying_no_source_tag(self, memory_store):
        """`wren memory store` and the MCP `store_query` tool pass the caller's
        own free-form labels through to `tags` -- "revenue,finance", or nothing
        at all -- so a real index holds rows with no `source:` token. A
        `--source user` filter must not claim them, least of all `forget`.
        """
        memory_store.store_query(
            nl_query="A user-tagged pair", sql_query="SELECT 1", tags="revenue,finance"
        )
        memory_store.store_query(
            nl_query="An untagged pair", sql_query="SELECT 2", tags=None
        )
        memory_store.store_query(
            nl_query="A source-tagged pair", sql_query="SELECT 3", tags="source:user"
        )

        rows, total = memory_store.list_queries(limit=100, source="user")
        assert total == 1
        assert rows[0]["nl_query"] == "A source-tagged pair"
        assert memory_store.count_queries_by_source("user") == 1
        assert [r["nl_query"] for r in memory_store.dump_queries(source="user")] == [
            "A source-tagged pair"
        ]

        assert memory_store.forget_queries_by_source("user") == 1
        survivors, _ = memory_store.list_queries(limit=100)
        assert sorted(r["nl_query"] for r in survivors) == [
            "A user-tagged pair",
            "An untagged pair",
        ]
```

With every inline change plus both tests applied: `ruff format --check src/` and `ruff check src/` clean, and `pytest tests/unit/ --ignore=tests/unit/test_mcp_server.py` → **1359 passed, 1 skipped**. (For what it's worth I don't reproduce the 3 `test_served_content_guard.py` failures the description mentions — they pass here.)

Two smaller notes, no action needed. `forget_queries_by_source` swapped an atomic `table.delete(where)` for a full-table read-modify-overwrite; that's defensible now that it shares the `forget_queries_by_ids` mechanism, but it loses atomicity and `load_queries(overwrite=True)` now does one full rewrite per distinct source — worth a line saying why that's acceptable. And `LanceDBIndex.rebuild()` has no production caller (tests only), so "three call sites" in the description is really two plus an interface method.

core/wren/src/wren/memory/store.py
```diff
@@ -26,12 +26,40 @@
...
     """Escape single quotes for LanceDB where-clause literals."""
     return value.replace("'", "''")
 
 
+def _tag_source(tags: str | None) -> str:
+    """Extract the ``source:`` value from a query_history tags string."""
+    for part in (tags or "").split():
+        if part.startswith("source:"):
+            return part[len("source:") :]
+    return "user"
```
goldmedal: **Blocking 2 — this default silently widens what `forget` deletes.** Moving the `--source` filters from `tags == f"source:{source}"` to `_tag_source(tags) == source` is right for the new `source:user origin:markdown-sync` rows, but the `"user"` fallback makes the filter claim rows that carry *no* `source:` token at all.

That is not hypothetical: both real write paths pass the caller's own labels straight into `tags`. Verified with the actual CLI —

```
$ wren memory store --nl "Total revenue" --sql "…" --tags revenue,finance
$ wren memory store --nl "Count orders"  --sql "…"
    'Total revenue' | 'revenue,finance'
    'Count orders'  | ''
```

so `--source user` now sweeps them up:

```
                       main   PR
  list --source user      1     3
  FORGET --source user    1     3   ← rows deleted
```
(index seeded with `'revenue,finance'`, `None`, `'source:user'`, `'source:seed'`)

Reporting only an explicit token keeps the filters matching the new sync-tagged rows while leaving `forget`'s blast radius exactly where `main` had it:

```suggestion
def _tag_source(tags: str | None) -> str | None:
    """Extract the explicit ``source:`` value from a tags string, else ``None``.

    ``None`` means the row carries no ``source:`` token at all -- which is what
    ``store_query`` writes for a `wren memory store` / MCP ``store_query`` pair,
    whose ``tags`` hold the caller's own free-form labels (``"revenue,finance"``)
    or nothing. Those rows must not be swept up by a ``--source user`` filter,
    least of all by ``forget``, so this stays distinct from the ``"user"``
    *display* default applied in ``cli._parse_source``.
    """
    for part in (tags or "").split():
        if part.startswith("source:"):
            return part[len("source:") :]
    return None
```

core/wren/src/wren/memory/store.py
```diff
@@ -26,12 +26,40 @@
 _SCHEMA_TABLE = "schema_items"
 _QUERY_TABLE = "query_history"
 
+# query_history sources that are not backed by knowledge/sql/*.md, so a
+# markdown sync must never forget them: seed/view are derived from the
+# manifest, legacy comes from a project's queries.yml.
+_NON_MARKDOWN_SOURCES = frozenset({"seed", "view", "legacy"})
+
```
goldmedal: Dead code once `check()` uses the provenance tag — this frozenset is its only consumer, and `sync_markdown_queries` already scopes its forget pass by `_MARKDOWN_SYNC_TAG` rather than by this list. Deleting it also removes the second, competing definition of "markdown-backed", which is what let the two paths drift apart in the first place.

```suggestion
```

core/wren/src/wren/memory/cli.py
```diff
@@ -882,10 +893,9 @@ def forget(
 
 def _parse_source(tags: str | None) -> str:
     """Extract source value from a (possibly null/empty) tags string."""
-    for part in (tags or "").split():
-        if part.startswith("source:"):
-            return part[len("source:") :]
-    return "user"
+    from wren.memory.store import _tag_source  # noqa: PLC0415
+
+    return _tag_source(tags)
```
goldmedal: Paired with the `_tag_source` change: the `"user"` default belongs here, on the display/export path, not in the predicate the `--source` filters share. This keeps `memory list` / `export` / `dump` output identical to today (and `test_parse_source_handles_null_tags` passing).

```suggestion
def _parse_source(tags: str | None) -> str:
    """Source value for display/export, defaulting an untagged row to "user".

    Distinct from ``store._tag_source``, which reports ``None`` for a row with
    no ``source:`` token so that ``--source`` filters do not match it.
    """
    from wren.memory.store import _tag_source  # noqa: PLC0415

    return _tag_source(tags) or "user"
```

core/wren/src/wren/memory/cli.py
```diff
@@ -255,7 +259,12 @@ def index(
                 raw = queries_file.read_text(encoding="utf-8")
                 doc = yaml.safe_load(raw)
                 if doc and isinstance(doc, dict) and doc.get("pairs"):
-                    load_result = mem_store.load_queries(doc["pairs"], upsert=False)
+                    # Tag legacy=queries.yml explicitly so a later
+                    # sync_markdown_queries() (not markdown-backed, so it
+                    # would otherwise treat these as stale and re-embed them
+                    # on every run) leaves them alone, same as seed/view rows.
```
goldmedal: The stated reason isn't the operative one: `sync_markdown_queries` scopes its forget pass to rows carrying `_MARKDOWN_SYNC_TAG`, so a `queries.yml` row was never at risk regardless of its source. Keeping the tag is still worth it — just for a different reason, and the accurate comment matters here because the stale one is what makes `check()`'s source-name exclusion list look justified.

```suggestion
                    # Tag legacy=queries.yml explicitly so these rows are
                    # distinguishable from markdown-backed ones in
                    # `memory list/dump/forget --source`. They are already
                    # safe from sync_markdown_queries, whose forget pass is
                    # scoped to rows carrying its own provenance tag.
```

core/wren/tests/unit/test_memory.py
```diff
@@ -1477,6 +1477,408 @@ def test_load_query_pairs_index_is_convergent(self, memory_store, tmp_path):
...
+        )
+
+        cli = CliRunner()
+        load_result = cli.invoke(app, ["memory", "load", str(pairs_file)])
+        assert load_result.exit_code == 0, load_result.output
+        assert "1 new" in load_result.output
+
+        index_result = cli.invoke(app, ["memory", "index"])
+        assert index_result.exit_code == 0, index_result.output
+        assert "Forgot" not in index_result.output
```
goldmedal, line 1829: This pins the half that looks right — `index` leaves the row alone — but not what `check` then says about it, which is where the contract actually breaks. Its legacy sibling (`test_cli_check_does_not_flag_legacy_queries_yml_pairs_as_stale`) does assert `In sync.`; without the same assertion here, `check` reporting drift that `index` cannot clear stays invisible to the suite.

Worth noting how insensitive the current suite is to the root cause: I switched `check()` to the provenance tag and **all 110 `test_memory.py` tests still passed**, so nothing today covers the divergence between the two definitions.

```suggestion
        assert "Forgot" not in index_result.output

        # ...and `check` must agree, rather than reporting drift that `index`
        # has just proven it cannot clear.
        check_result = cli.invoke(app, ["memory", "check"])
        assert check_result.exit_code == 0, check_result.output
        assert "In sync." in check_result.output
```

AmirF194:
Applied all five inline suggestions together plus both regression tests, pushed to 807d7d47. Renamed indexed_user to indexed_md as you suggested and updated the two trailing comments and the stale-count message wording to match. Verified against the pushed commit: ruff format/check clean, tests/unit/test_memory.py 112/112 (your two new tests included). Left the two non-blocking notes as is since you said no action needed there.

goldmedal (approved):
Thanks @AmirF194 👍

AmirF194:
Appreciate you catching the seed-collision regression and then the source:user gap, tracking provenance directly instead of inferring it is a much cleaner fix than what I had.

PR #2707, opened 2026-08-28
feat(wren): add a torch-free onnx embedding backend for wren memory

### LINKED ISSUE #2643: feat(wren memory): torch-free ONNX embedding backend to cut install size and cold start
### Problem

`wrenai[memory]` is currently `["lancedb>=0.6", "sentence-transformers>=3.0.0"]`. That pulls in `torch`, and on `linux-x86_64` the default index resolves the **CUDA build**: torch plus the whole `nvidia-*` stack, roughly 2.7 GB of wheels and 6–7 GB installed, where a CPU-only build is about 200 MB. It also imposes a per-process floor before the first result: importing `torch` and loading the model weights, which is the dominant cost in #2611.

A CPU-only pin is not available to us: **an extra cannot pin a package index**, so `wrenai[memory]` cannot express "torch, but the CPU wheel". Every user on Linux pays the CUDA download unless they know to reinstall torch from the PyTorch CPU index afterwards. Telling users to do that by hand is a workaround, not a fix.

### Proposal

Add a torch-free embedding backend based on `onnxruntime`, selectable at install and runtime, and keep `sentence-transformers` as the existing option.

- New extra, e.g. `wrenai[memory-onnx]`: `lancedb` + `onnxruntime` + tokenizer, no torch.
- Runtime selection via an env var, e.g. `WREN_EMBEDDING_BACKEND=onnx|sentence-transformers`, defaulting to whichever is installed (ONNX preferred when both are).
- Slot it in behind the existing embedding abstraction in `core/wren/src/wren/memory/embeddings.py` so `MemoryStore` is unchanged: the backend only has to provide `compute_source_embeddings` / `compute_query_embeddings`.
- **Keep the default model's 384 dimensions and multilingual coverage** (`paraphrase-multilingual-MiniLM-L12-v2` equivalent). Existing LanceDB tables are typed with a fixed-size 384 vector, so preserving the dimension means existing indexes keep working; a different dimension forces every user to `wren memory reset && wren memory index`.

### What the evaluation needs to answer

- Cold-start time to first embedding, compared with the current sentence-transformers path.
- Install footprint of the ONNX extra vs. the current one, on `linux-x86_64` specifically (where the CUDA resolution happens).
- Retrieval parity: the same set of `fetch` / `recall` queries should return comparable results — this must not quietly degrade Chinese/multilingual retrieval quality.
- Where the ONNX weights come from and how they are cached offline (the local-first resolution behaviour must be preserved).
- Platform coverage: linux/macOS/Windows, x86_64 and arm64.

### Acceptance criteria

- `pip install wrenai[memory-onnx]` installs no torch and no `nvidia-*` packages.
- `wren memory index` / `fetch` / `recall` work end-to-end on the ONNX backend, at 384 dimensions, against an index built by either backend.
- Model resolution stays local-first, so a cached model needs no network access.
- Backend selection is explicit and reported by `wren memory status`.
- The evaluation numbers above are recorded on this issue before the default is changed.

### Context

Follow-up from #2611. This is the item that removes both the multi-GB install and the per-process load floor, rather than working around them.

audi0417:
Closes #2643.

## Why

`wrenai[memory]` pulls sentence-transformers and therefore torch. Resolving that extra for `linux-x86_64` produces torch, triton and **sixteen `nvidia-*` packages** — verified with `uv pip compile --python-platform x86_64-unknown-linux-gnu`. As the issue notes, an extra cannot pin a package index, so there is no way to say "torch, but the CPU wheel" from inside `pyproject.toml`.

## What

A `wrenai[memory-onnx]` extra — lancedb, onnxruntime, tokenizers — as an *alternative* to `memory` rather than an addition, selected at runtime with `WREN_EMBEDDING_BACKEND=onnx|sentence-transformers` and defaulting to onnx when it is importable. It is deliberately excluded from `all`, which keeps installing the sentence-transformers backend.

The weights are the ONNX export **published in the same HF repo as the torch weights** (`onnx/model.onnx`), so nothing is re-exported or converted, and the existing local-first HF cache resolution carries over unchanged. The backend reproduces the sentence-transformers pipeline: tokenize to the model's `max_seq_length`, encode, attention-masked mean pooling, L2 normalize.

It slots in behind the existing abstraction — `MemoryStore` only ever calls `compute_source_embeddings` / `compute_query_embeddings`, so it is untouched.

## Evaluation

Against the questions the issue asked:

| | sentence-transformers | onnx |
|---|---|---|
| Cold start to first embedding (cached model, median of 3) | 20.3 s | **3.0 s** |
| Installed size, macOS arm64 | 817 MB | **133 MB** |
| `linux-x86_64` resolution | torch + triton + 16 × `nvidia-*` | none of them |
| Vector dimension | 384 | 384 |

**Retrieval parity.** Worst-case cosine between the two backends is `1.00000000`, max absolute delta `6.11e-07` — float32 rounding — over English, Traditional Chinese, Japanese and Korean inputs. The vectors are equal, not merely close, so **existing LanceDB tables need no reindex**. End-to-end recall against a real store resolves cross-lingual queries correctly (`哪些客戶花最多錢` → `top customers by net spend`).

**Weights and caching.** `hf_hub_download(..., local_files_only=True)` first, falling back to an online fetch on `LocalEntryNotFoundError` — the same local-first shape, and the same `OSError` branch, as the sentence-transformers adapter.

**Platform coverage.** onnxruntime publishes wheels for linux/macOS/Windows on x86_64 and arm64. Verified locally on macOS arm64; the new CI job covers linux-x86_64.

## Two things that would have degraded silently

**Normalization.** LanceDB's `SentenceTransformerEmbeddings` defaults to `normalize=True` and the existing adapter does not override it, so every vector already in a store is L2-normalized. My first implementation skipped this and the parity test caught it. Unnormalized output would have left old and new rows on different scales and skewed distance ranking — while still looking fine in any cosine-based smoke test, because cosine is scale-invariant.

**Pooling mode.** Only mean pooling is implemented. A CLS-pooled model would still yield a 384-vector, so `1_Pooling/config.json` is checked and anything else raises instead of being indexed with quietly wrong vectors.

## What the change requires

Two edits outside the issue's text that the feature does not work without, flagging them so they are not mistaken for drive-by changes:

1. **`_extra_available()`** in `index_backend.py` tested for `sentence_transformers` by name, so a `memory-onnx` install would have been silently downgraded to the Grep backend — the extra would have done nothing.
2. **Three sentence-transformers-specific tests** went through `get_embedding_function`, which now dispatches to onnx when both extras are installed. Left alone they exercise the wrong backend, and the concurrency one hangs waiting on a fake model that is never constructed. They build the adapter directly now. `TestLocalFirstEmbeddings` also needed its `importorskip` moved into an autouse fixture — it monkeypatches `sentence_transformers` by string, which imports the module before an in-body skip can fire.

## Also included — happy to split these out

3. **Test gating.** The store and `WrenMemory` suites gated on `sentence_transformers`, so they skipped wholesale under `memory-onnx` — exactly where running them proves something. Regated on either backend: 53 skips become 104 executed tests on an onnx-only install.
4. **A CI job** installing `memory-onnx`, asserting torch is absent, and running the memory suite. Without it nothing would notice torch creeping back into the resolution. This is the one piece that stands alone cleanly; note that dropping 3 as well would leave it running very little.

## Verification

| Environment | Result |
|---|---|
| `memory` only (mirrors the existing CI job) | 111 passed, 1 skipped |
| `memory-onnx` only, torch absent | 105 passed, 7 skipped |
| Both extras installed | 112 passed |
| `uvx ruff format --check src/` and `ruff check src/` | clean |
| `uv lock --check` | clean — adds only `onnxruntime` and `flatbuffers` |
| No extras installed | `wren.memory.embeddings` still imports; backend resolves to Grep |

`tests/unit/test_served_content_guard.py` has 3 failures locally, identical on `main` — unrelated to this change.

## Summary by CodeRabbit

* **New Features**
  * Added a PyTorch-free ONNX embedding backend for vector search.
  * Added the `memory-onnx` installation option.
  * Added automatic embedding backend selection and configuration.
  * Memory status now displays the active embedding backend and model.
  * ONNX embeddings remain compatible with existing vectors.

* **Bug Fixes**
  * Improved command-line reporting for embedding configuration and loading errors.

* **Tests**
  * Added ONNX embedding, backend detection, and vector compatibility coverage.
  * Added CI checks for torch-free installation and embedding parity.

### DIFF
--- a/.github/workflows/wren-ci.yml
+++ b/.github/workflows/wren-ci.yml
@@ -266,6 +266,35 @@ jobs:
           WREN_EMBEDDING_MODEL: paraphrase-MiniLM-L3-v2
         run: uv run --no-sync pytest tests/unit/test_memory.py -v
 
+  test-embedding-parity:
+    name: embedding backend parity
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: core/wren
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
+      - name: Install both embedding extras
+        # Each of the two jobs above installs one backend, so the parity test
+        # importorskips out of both. It is the assertion existing stores
+        # depend on — a store written by one backend has to stay readable by
+        # the other — so it needs a job where both are present.
+        run: uv sync --locked --extra memory --extra memory-onnx
+      - name: Cache models
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/huggingface
+          key: hf-parity-paraphrase-MiniLM-L3-v2
+      - name: Compare backends
+        env:
+          WREN_EMBEDDING_MODEL: paraphrase-MiniLM-L3-v2
+        run: uv run --no-sync pytest tests/unit/test_memory.py -v -k TestOnnxVectorParity
+
   test-mcp:
     name: mcp tests
     runs-on: ubuntu-latest
--- a/.github/workflows/wren-ci.yml
+++ b/.github/workflows/wren-ci.yml
@@ -213,6 +213,59 @@ jobs:
           WREN_EMBEDDING_MODEL: paraphrase-MiniLM-L3-v2
         run: uv run --no-sync pytest tests/unit/test_memory.py -v
 
+  test-memory-onnx:
+    name: memory tests (onnx, torch-free)
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: core/wren
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - name: Install uv
+        uses: astral-sh/setup-uv@v4
+      - name: Cache Cargo
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            ./core/wren-core-py/target/
+          key: ${{ runner.os }}-cargo-wren-${{ hashFiles('**/Cargo.lock') }}
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Build wren-core-py wheel
+        working-directory: core/wren-core-py
+        run: |
+          uv sync --locked --no-install-project
+          uv run --no-sync maturin build
+      - name: Install dependencies
+        run: |
+          uv sync --locked --extra memory-onnx
+          uv pip install --reinstall --no-index --find-links ../wren-core-py/target/wheels/ wren-core-py
+      - name: Assert the install is torch-free
+        # The whole point of this extra: on linux-x86_64 the sentence-transformers
+        # path resolves the CUDA build of torch. If torch reappears here, the
+        # extra has regressed.
+        run: |
+          if uv run --no-sync python -c "import importlib.util as u; raise SystemExit(0 if u.find_spec('torch') else 1)"; then
+            echo "::error::torch was installed by the memory-onnx extra"
+            exit 1
+          fi
+      - name: Cache ONNX model
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/huggingface
+          key: hf-onnx-paraphrase-MiniLM-L3-v2
+      - name: Run memory tests
+        env:
+          WREN_EMBEDDING_MODEL: paraphrase-MiniLM-L3-v2
+        run: uv run --no-sync pytest tests/unit/test_memory.py -v
+
   test-mcp:
     name: mcp tests
     runs-on: ubuntu-latest
--- a/.github/workflows/wren-ci.yml
+++ b/.github/workflows/wren-ci.yml
@@ -281,6 +281,8 @@ jobs:
         working-directory: core/wren
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
       - uses: actions/setup-python@v5
         with:
           python-version: "3.11"
--- a/.github/workflows/wren-ci.yml
+++ b/.github/workflows/wren-ci.yml
@@ -298,10 +298,8 @@ jobs:
         uses: actions/cache@v4
         with:
           path: ~/.cache/huggingface
-          key: hf-parity-paraphrase-MiniLM-L3-v2
+          key: hf-parity-paraphrase-multilingual-MiniLM-L12-v2
       - name: Compare backends
-        env:
-          WREN_EMBEDDING_MODEL: paraphrase-MiniLM-L3-v2
         run: uv run --no-sync pytest tests/unit/test_memory.py -v -k TestOnnxVectorParity
 
   test-mcp:
--- a/.github/workflows/wren-ci.yml
+++ b/.github/workflows/wren-ci.yml
@@ -249,11 +249,18 @@ jobs:
           uv pip install --reinstall --no-index --find-links ../wren-core-py/target/wheels/ wren-core-py
       - name: Assert the install is torch-free
         # The whole point of this extra: on linux-x86_64 the sentence-transformers
-        # path resolves the CUDA build of torch. If torch reappears here, the
-        # extra has regressed.
+        # path resolves the CUDA build of torch plus triton and sixteen nvidia-*
+        # packages. Checking torch alone would miss a resolution that drops
+        # torch but keeps the CUDA payload, which is where the gigabytes are.
         run: |
           if uv run --no-sync python -c "import importlib.util as u; raise SystemExit(0 if u.find_spec('torch') else 1)"; then
             echo "::error::torch was installed by the memory-onnx extra"
+            exit 1
+          fi
+          cuda=$(uv pip list --format=freeze | grep -iE '^(nvidia-|triton==)' || true)
+          if [ -n "$cuda" ]; then
+            echo "::error::CUDA packages were installed by the memory-onnx extra"
+            echo "$cuda"
             exit 1
           fi
       - name: Cache ONNX model
--- a/core/wren/src/wren/memory/embeddings.py
+++ b/core/wren/src/wren/memory/embeddings.py
@@ -191,6 +191,31 @@ def _hf_file(repo_id: str, filename: str) -> str:
         return hf_hub_download(repo_id, filename)
 
 
+def _is_absent_from_repo(error: OSError) -> bool:
+    """True when the hub answered "no such file", as opposed to not answering.
+
+    A remote 404 is the only response that means the repo does not publish the
+    file. ``LocalEntryNotFoundError`` is the opposite -- nothing cached *and*
+    the hub unreachable -- and a rate limit or dropped connection say nothing
+    about the file at all. Reporting either as absence is how a transient
+    fault acquires a permanent-sounding explanation.
+
+    Tested against the pair rather than ``RemoteEntryNotFoundError``, which
+    only exists from huggingface-hub 1.x; the pair discriminates identically
+    on both. Needs >= 0.25 all the same, which is where they moved into
+    ``huggingface_hub.errors`` -- before that this import raises inside the
+    caller's ``except`` handler, replacing the error being classified.
+    """
+    from huggingface_hub.errors import (  # noqa: PLC0415
+        EntryNotFoundError,
+        LocalEntryNotFoundError,
+    )
+
+    return isinstance(error, EntryNotFoundError) and not isinstance(
+        error, LocalEntryNotFoundError
+    )
+
+
 def _read_json(repo_id: str, filename: str) -> dict | None:
     """Read a small JSON config from the model repo, or None when absent.
 
@@ -206,17 +231,7 @@ def _read_json(repo_id: str, filename: str) -> dict | None:
     try:
         path = _hf_file(repo_id, filename)
     except OSError as e:
-        from huggingface_hub.errors import (  # noqa: PLC0415
-            EntryNotFoundError,
-            LocalEntryNotFoundError,
-        )
-
-        # A remote 404 is the only answer that means "this repo does not
-        # publish the file". LocalEntryNotFoundError is the opposite: nothing
-        # cached *and* the hub unreachable, so presence is unknown.
-        if isinstance(e, EntryNotFoundError) and not isinstance(
-            e, LocalEntryNotFoundError
-        ):
+        if _is_absent_from_repo(e):
             return None
         raise OnnxBackendError(
             f"'{filename}' for '{repo_id}' could not be fetched: {e}. Whether "
@@ -335,16 +350,24 @@ class OnnxEmbeddings:
             except OSError as e:
                 # onnx is selected on importability alone, so a
                 # WREN_EMBEDDING_MODEL that works under sentence-transformers
-                # reaches this. Not falling back: a network blip and a repo
-                # with no export both arrive as OSError, and quietly loading
-                # torch on a bad connection is the outcome this extra exists
-                # to avoid. Name the override and let the caller choose.
-                raise MissingOnnxExportError(
+                # reaches this. Only a 404 means the export is really missing;
+                # anything else and the remedy is to retry, not to switch
+                # backends. Deliberately not falling back on the 404 either:
+                # quietly loading torch is the outcome this extra exists to
+                # avoid, so name the override and let the caller choose.
+                if _is_absent_from_repo(e):
+                    raise MissingOnnxExportError(
+                        f"'{self._repo_id}' does not publish '{onnx_file}', "
+                        "which the onnx embedding backend needs. Set "
+                        "WREN_EMBEDDING_BACKEND=sentence-transformers to use "
+                        "this model, or point WREN_ONNX_MODEL_FILE at the "
+                        "right path."
+                    ) from e
+                raise OnnxBackendError(
                     f"'{onnx_file}' could not be fetched for '{self._repo_id}': "
-                    f"{e}. The onnx embedding backend needs a repo that "
-                    "publishes an ONNX export. Set "
-                    "WREN_EMBEDDING_BACKEND=sentence-transformers to use this "
-                    "model, or point WREN_ONNX_MODEL_FILE at the right path."
+                    f"{e}. Retry when the Hugging Face cache or network is "
+                    "available, or set "
+                    "WREN_EMBEDDING_BACKEND=sentence-transformers."
                 ) from e
 
             session = onnxruntime.InferenceSession(
--- a/core/wren/src/wren/memory/embeddings.py
+++ b/core/wren/src/wren/memory/embeddings.py
@@ -34,6 +34,13 @@ ONNX_BACKEND = "onnx"
 SENTENCE_TRANSFORMERS_BACKEND = "sentence-transformers"
 
 
+class UnsupportedPoolingError(RuntimeError):
+    """The onnx backend cannot reproduce this model's pooling mode.
+
+    Not a ``ValueError``: the CLI reports those as a malformed manifest.
+    """
+
+
 def _disable_transformers_progress_bar() -> None:
     # Imported lazily: transformers ships with the optional `memory` extra,
     # so this module must stay importable when that extra is not installed.
@@ -200,7 +207,7 @@ def _require_mean_pooling(repo_id: str) -> None:
         for key, value in config.items()
         if key.startswith("pooling_mode_") and value
     )
-    raise ValueError(
+    raise UnsupportedPoolingError(
         f"The onnx embedding backend implements mean pooling, but '{repo_id}' "
         f"pools with {active or ['an unrecognized mode']}. Set "
         "WREN_EMBEDDING_BACKEND=sentence-transformers to use this model."
--- a/core/wren/src/wren/memory/embeddings.py
+++ b/core/wren/src/wren/memory/embeddings.py
@@ -194,16 +194,36 @@ def _hf_file(repo_id: str, filename: str) -> str:
 def _read_json(repo_id: str, filename: str) -> dict | None:
     """Read a small JSON config from the model repo, or None when absent.
 
-    A corrupt file is not treated as absent. ``_require_mean_pooling`` reads
-    "absent" as "assume mean pooling", so swallowing the error there would
-    wave a CLS-pooled model through with quietly wrong vectors -- and
-    ``json.JSONDecodeError`` is a ``ValueError``, so letting it escape would
-    put it behind the CLI's "Malformed manifest:" instead.
+    "Absent" means the hub answered 404. A cache miss that could not be
+    checked online, a rate limit and a dropped connection are not absence, and
+    neither is a corrupt file. The distinction matters because
+    ``_require_mean_pooling`` reads "absent" as "assume mean pooling": report
+    any of those as absence and a network blip waves a CLS-pooled model
+    through with quietly wrong vectors. ``json.JSONDecodeError`` is also a
+    ``ValueError``, so letting it escape would put it behind the CLI's
+    "Malformed manifest:".
     """
     try:
         path = _hf_file(repo_id, filename)
-    except OSError:
-        return None
+    except OSError as e:
+        from huggingface_hub.errors import (  # noqa: PLC0415
+            EntryNotFoundError,
+            LocalEntryNotFoundError,
+        )
+
+        # A remote 404 is the only answer that means "this repo does not
+        # publish the file". LocalEntryNotFoundError is the opposite: nothing
+        # cached *and* the hub unreachable, so presence is unknown.
+        if isinstance(e, EntryNotFoundError) and not isinstance(
+            e, LocalEntryNotFoundError
+        ):
+            return None
+        raise OnnxBackendError(
+            f"'{filename}' for '{repo_id}' could not be fetched: {e}. Whether "
+            "this model is usable here could not be determined; retry with the "
+            "Hugging Face cache or network available, or set "
+            "WREN_EMBEDDING_BACKEND=sentence-transformers."
+        ) from e
     try:
         return json.loads(Path(path).read_text(encoding="utf-8"))
     except (OSError, json.JSONDecodeError) as e:
@@ -225,7 +245,15 @@ def _max_seq_length(repo_id: str) -> int:
     the default model reports 512 there against a real length of 128, so
     reading it would diverge from the sentence-transformers backend.
     """
-    config = _read_json(repo_id, "sentence_bert_config.json") or {}
+    try:
+        config = _read_json(repo_id, "sentence_bert_config.json") or {}
+    except OnnxBackendError:
+        # Truncation length is soft where pooling mode is not. An ONNX-native
+        # mirror legitimately ships no sentence_bert_config.json, so an
+        # offline run with the rest of the model cached should fall back here
+        # rather than fail -- getting the length wrong degrades recall,
+        # getting the pooling wrong corrupts the index.
+        config = {}
     value = config.get("max_seq_length")
     if isinstance(value, int) and value > 0:
         return value
--- a/core/wren/src/wren/memory/embeddings.py
+++ b/core/wren/src/wren/memory/embeddings.py
@@ -29,6 +29,9 @@ _DEFAULT_MODEL = os.getenv(
 )
 _DEFAULT_DIM = 384
 _DEFAULT_MAX_SEQ_LENGTH = 128
+# Matches SentenceTransformer.encode's default batch_size, so peak memory
+# per encode call is the same on both backends.
+_ENCODE_BATCH_SIZE = 32
 
 ONNX_BACKEND = "onnx"
 SENTENCE_TRANSFORMERS_BACKEND = "sentence-transformers"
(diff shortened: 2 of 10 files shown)


audi0417:
Pushed `cbbf3b6..9b9ed3f` covering the review points.

**Parity test never ran** — correct, and it was the assertion existing stores depend on. `test-memory` installs sentence-transformers without onnxruntime and `test-memory-onnx` does the reverse, so `TestOnnxVectorParity` `importorskip`ed out of both. Added a `test-embedding-parity` job that installs both extras and runs it. The two existing jobs are unchanged.

**`wren memory status` did not name the backend** — also correct. Which backend is live is now a real choice (env var, install shape, or fallback) and the only symptoms of getting the wrong one were indirect: a slow cold start, or an 800 MB venv instead of 130 MB.

```
$ WREN_EMBEDDING_BACKEND=onnx wren memory status
Backend: lancedb
  embeddings: onnx (paraphrase-MiniLM-L3-v2)

$ WREN_EMBEDDING_BACKEND=sentence-transformers wren memory status
Backend: lancedb
  embeddings: sentence-transformers (paraphrase-MiniLM-L3-v2)
```

Reported from `resolve_embedding_backend()`, not from a constructed model, so `status` still never loads one — `TestMemoryStoreLazyModelLoad` already asserts that and still passes. The grep backend embeds nothing, so it prints no embedding line; there's a test for each.

**CUDA packages, not just torch** — the job asserted `torch` was absent, which is the package the issue names but not where the gigabytes are. A resolution that dropped torch while keeping triton and the sixteen `nvidia-*` wheels would have passed. It now fails on those too.

**Silent fallback** (from the merge-risk note) — kept the fallback, since both backends emit the same vectors so failing hard would be worse, but an unhonored `WREN_EMBEDDING_BACKEND` now logs a warning naming the extra to install. An empty or unrecognized value still means auto-detect and stays quiet.

**Docstring coverage** — added them to the onnx backend's remaining surface (`compute_source_embeddings` / `compute_query_embeddings`, `_encode`, the availability probes); `embeddings.py` is now 21/22. I did not add them to the 137 test methods the check also counts: `test_memory.py` is 12/149 on `main` and documents intent in comments, so docstringing only the new tests would fork the file's convention for a metric.

**Verification** (macOS arm64, both extras installed): 1357 passed, 3 skipped across `tests/unit`; `test_memory.py` + `test_index_backend.py` 127 passed; the new status and parity tests pass under `WREN_EMBEDDING_BACKEND=onnx` and `=sentence-transformers`; `ruff format --check` and `ruff check` clean on the touched files. The 3 `test_served_content_guard.py` failures I noted in the description do not reproduce in a clean venv — disregard that line.

audi0417:
Fixed in `a9a1fcb` — `persist-credentials: false` on the parity job's checkout.

Worth recording why this one is a fix rather than a decline: 7 of the 8 checkout steps in `wren-ci.yml` do *not* set it, so on a first look the finding reads as breaking the file's convention. But `git log -S persist-credentials` puts its only occurrence in #2681 (2026-08-26), the commit that added `test-mcp` — the newest job in the file, and the one directly below the job I added. So the older jobs predate the practice rather than opting out of it, and a new job should follow the newer convention.

It also isn't cosmetic here: the parity job runs `uv sync` and `pytest`, so dependency and project code executes with whatever the checkout left in `.git/config`.

The seven older checkouts want the same treatment. I've left them alone — that's a workflow-wide change with its own reviewable shape, not something to bury in an embedding-backend PR.

Note the Wren SDK CI runs on this PR are still `action_required` (fork PR pending approval), so none of the new jobs — `test-memory-onnx`, `test-embedding-parity` — have actually executed upstream yet. Everything I've reported is from local runs.

core/wren/src/wren/memory/embeddings.py
```diff
@@ -74,14 +102,213 @@ def get_embedding_model(self):
...
+    """
+    config = _read_json(repo_id, "1_Pooling/config.json")
+    if config is None or config.get("pooling_mode_mean_tokens"):
+        return
+    active = sorted(
+        key.removeprefix("pooling_mode_")
+        for key, value in config.items()
+        if key.startswith("pooling_mode_") and value
+    )
+    raise ValueError(
```
goldmedal: **This `ValueError` reaches the user labelled as a manifest problem.**

`index_schema` is called from `core/wren/src/wren/memory/cli.py` inside an `except ValueError` that prints `Malformed manifest: {e}`, so this guard — the one that exists precisely to avoid a silent wrong-vector index — tells the user their MDL is broken.

Reproduced on this branch with a CLS-pooled 384-dim model:

```
$ WREN_EMBEDDING_MODEL=BAAI/bge-small-en-v1.5 wren memory index -m mdl.json
Malformed manifest: The onnx embedding backend implements mean pooling, but
'BAAI/bge-small-en-v1.5' pools with ['cls_token']. Set
WREN_EMBEDDING_BACKEND=sentence-transformers to use this model.
```

The message itself is good; the prefix sends people to inspect the wrong thing. Raising a dedicated exception type here (and letting `cli.py` report it on its own terms) keeps the diagnosis intact. The `except ValueError` handler predates this PR, but this is a new `ValueError` source flowing into it.

Worth noting the ordering here is right: the guard runs before the tokenizer and session are built, so the failure costs one 8 KB config fetch and not the ~470 MB graph — I confirmed the HF cache for that model contained only `1_Pooling/config.json` afterwards.

.github/workflows/wren-ci.yml
```diff
@@ -213,6 +213,97 @@ jobs:
...
+        # the other — so it needs a job where both are present.
+        run: uv sync --locked --extra memory --extra memory-onnx
+      - name: Cache models
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/huggingface
+          key: hf-parity-paraphrase-MiniLM-L3-v2
+      - name: Compare backends
+        env:
+          WREN_EMBEDDING_MODEL: paraphrase-MiniLM-L3-v2
```
goldmedal: **This override means the parity gate never compares the model that actually ships.**

`_DEFAULT_MODEL` is `os.getenv("WREN_EMBEDDING_MODEL", "paraphrase-multilingual-MiniLM-L12-v2")` evaluated at import, so with this env set the job compares `paraphrase-MiniLM-L3-v2` — an English BERT/WordPiece model. The shipped default is XLM-R/SentencePiece: different tokenizer family, no `token_type_ids` input, pad token id 1 rather than 0. The job log confirms what ran:

```
WREN_EMBEDDING_MODEL: paraphrase-MiniLM-L3-v2
TestOnnxVectorParity::test_onnx_matches_sentence_transformers PASSED
1 passed, 115 deselected in 21.02s
```

So the load-bearing claim of the PR — same vectors, existing LanceDB tables need no reindex — is gated on a model no user gets by default.

To be clear about the risk: I ran the parity test against the real default locally and **it passes**, including a >128-token row that forces truncation:

| input | cosine | max abs delta |
|---|---|---|
| `哪些客戶花最多錢` | 1.0000000058 | 8.9e-08 |
| `月別の売上推移` | 1.0000000458 | 1.8e-07 |
| `월별 매출 추이` | 1.0000000461 | 1.8e-07 |
| 400-token English (truncated) | 1.0000000303 | 1.4e-07 |

I also wrote a store with one backend and read it with the other, in both directions: identical top-3 ordering on five cross-lingual queries, distances agreeing to the sixth decimal.

So this is a coverage gap rather than a defect — but the property is currently true and nothing keeps it true. Dropping this one line makes the job exercise the default (the other two jobs can keep the small model for speed); `onnx/model.onnx` for that repo is ~470 MB, so give the job its own `hf-` cache key and it is a one-time fill. The job currently finishes in 1m9s, so there is room.

goldmedal (commented):
Two things I'd fix before merge; everything else I found is follow-up material. CI is green and the design is right — details inline.

audi0417:
Both fixed.

**Pooling guard (`embeddings.py`)** — `_require_mean_pooling` now raises `UnsupportedPoolingError`. It subclasses `RuntimeError` rather than `ValueError` so the "ValueError means the manifest" convention that `schema_indexer`'s docstring spells out keeps its single meaning, and `cli.py` handles it before the `ValueError` branch and reports it on its own terms:

```
$ WREN_EMBEDDING_MODEL=BAAI/bge-small-en-v1.5 wren memory index -m mdl.json
The onnx embedding backend implements mean pooling, but 'BAAI/bge-small-en-v1.5'
pools with ['cls_token']. Set WREN_EMBEDDING_BACKEND=sentence-transformers to use
this model.
$ echo $?
1
```

The ordering you noted survives the change — the HF cache for that repo held only `1_Pooling/config.json`, 8 KB, afterwards.

**Parity job (`wren-ci.yml`)** — override dropped, so the job exercises the shipped default; the cache key follows the model. Against the real default locally: `1 passed, 115 deselected`.

PS on the cache — it is a larger fill than 470 MB. The job needs *both* backends' weights for that repo, so the onnx graph plus the torch weights: my `~/.cache/huggingface` entry for `paraphrase-multilingual-MiniLM-L12-v2` measures 906 MB. Still a one-time fill under its own key, but the first run will be noticeably longer than the current 1m9s.

goldmedal:
Re-reviewed at `f64608a`. Both items I asked for are fixed, and fixed in a way I could check rather than take on trust.

**Pooling guard.** `UnsupportedPoolingError` subclasses `RuntimeError`, and `cli.py` handles it ahead of the `ValueError` branch, so the guard no longer reaches the user labelled as a manifest problem. Your reasoning for not making it a `ValueError` is the right one.

**Parity gate.** I read the job log rather than the config: the override is gone and the job now runs the shipped default — `1 passed, 130 deselected in 29.96s`, cache saved under `hf-parity-paraphrase-multilingual-MiniLM-L12-v2` at 871 MB, job total 1m2s. So the "existing tables need no reindex" claim is now gated on the model users actually get. Your PS predicted the first run would be noticeably longer; it wasn't, and the cache is a one-time fill under its own key.

All 12 checks green. Five further findings below. **None of them block merge** — no data format, no API, no migration, and nothing that mis-reports to the user, which was the bar the last two items cleared. I'd still like the first one in this PR, for cost rather than severity; more on that at the end.

---

### 1. `_encode` sends the whole corpus through one `session.run`, where sentence-transformers chunks at 32

`_encode` does no chunking, and neither caller pages: `store.py:203` embeds every record from `extract_schema_items` in one call, and `store.py:530` every NL query in `_prepare_query_records`. Batch size is therefore the size of the manifest, with no upper bound.

The backend this replaces does not behave that way. lancedb's adapter calls `encode(list(texts), convert_to_numpy=True, normalize_embeddings=self.normalize)` with no `batch_size`, and `SentenceTransformer.encode` defaults to `batch_size=32` — checked against the signature in sentence-transformers 6.0.1. That is the asymmetry that matters: on the sentence-transformers path peak memory is bounded by 32 rows regardless of manifest size, on the onnx path it is bounded by the manifest. The same `MemoryStore` call goes from O(1) to O(n) in peak memory purely by switching backend.

Looking at the return value understates this, because attention is quadratic in sequence length. For the default model (12 layers, 12 heads, hidden 384, intermediate 1536, `max_seq_length` 128), at 4 000 texts padded to 128:

| tensor | shape | bytes |
|---|---|---|
| pooled output (the return value) | batch × 384 | ~6 MB |
| `last_hidden_state` | batch × seq × 384 | ~786 MB |
| attention scores, per layer | batch × heads × seq² | ~3.1 GB |
| FFN intermediate, per layer | batch × seq × 1536 | ~3.1 GB |

onnxruntime plans and reuses buffers across nodes and may fuse attention, so the peak is not twelve layers summed — but several GB-scale allocations are live at once. To be clear about what this is: an argument from the code and the tensor shapes, **not** a measured OOM. I did not push a large manifest through onnxruntime, and the largest manifest I have to hand is 18 models / 91 columns, which is ~110 records and entirely harmless. This is about scaling, not about anything in the repo's own fixtures.

Two things make the batch worse than the record count suggests:

- `extract_schema_items` emits one record per model, **per column**, per relationship, per view, and per cube/measure/dimension — so the count tracks total field count, not table count.
- `enable_padding()` is called without `length`, so the batch pads to its own longest row (capped at 128). One verbose column description drags the entire batch to the pessimistic row of that table; on the sentence-transformers path the same row only inflates its own group of 32.

Nothing in the suite can see this: the largest batch under test is two rows, the parity test uses three, and `_runtime` is monkeypatched in the unit tests, so no test gets slower or fatter as batch size grows.

The fix is a chunk loop in `_encode` at 32 or 64. It is cheap for two reasons worth stating: mean pooling and L2 normalization are both per-row, so **the vectors are bit-identical** and the parity gate keeps covering them unchanged; and chunking also confines padding to each chunk, which removes the second effect above for free.

### 2. Preferring onnx on importability alone can turn a working custom-model install into a 404 traceback

`resolve_embedding_backend()` selects onnx whenever `onnxruntime` and `tokenizers` import. `tokenizers` always ships with the `memory` extra via transformers, so onnxruntime alone decides — and plenty of unrelated packages pull it in. If `WREN_EMBEDDING_MODEL` points at a repo with no ONNX export, `_hf_file(repo, "onnx/model.onnx")` raises. Verified live:

```
_hf_file("prajjwal1/bert-tiny", "onnx/model.onnx")
  -> RemoteEntryNotFoundError (an OSError)
     404 Client Error. (Request ID: Root=1-6a9a2dcf-...)
```

`index` catches only `UnsupportedPoolingError` and `ValueError`, so this surfaces as a typer pretty-traceback whose most informative line is `404 Client Error`. Two things stand out. A configuration that worked before now fails; and it fails *with sentence-transformers installed and able to serve the request* — the opposite of the missing-extra case, which warns and falls back. Either fall back here too, or re-raise naming `WREN_EMBEDDING_BACKEND=sentence-transformers`, which is the courtesy the pooling guard already extends.

### 3. `memory-onnx` is documented nowhere

`README.md` lists every other extra on its own `pip install 'wrenai[...]'` line, and §6 is the memory walkthrough. This PR touches no docs, so the extra is discoverable only by reading `pyproject.toml`. One install line plus a sentence in §6 covers it. (`WREN_EMBEDDING_MODEL` is undocumented too, so env vars are not an established surface here — the extras list is.)

### 4. `_max_seq_length` silently defaults to 128 for exactly the repos an onnx user reaches for

ONNX-native mirrors publish `onnx/model.onnx` but no `sentence_bert_config.json` and no `1_Pooling/config.json` — confirmed via the HF API for `Xenova/paraphrase-multilingual-MiniLM-L12-v2` and `Xenova/all-MiniLM-L6-v2`. Run against the Xenova mirror of the default model, `_require_mean_pooling` passes silently and `_max_seq_length` returns 128.

Both outcomes are correct there, and the `_read_json` → `None` path does work: I checked it specifically because "or None when absent" looked like it could be dead code, and it isn't — huggingface-hub 1.8 raises `RemoteEntryNotFoundError`, which subclasses `OSError`. So this is not a defect, only an invisible default: a 512-length model would truncate at 128 with no signal. Note the fix is *not* to fall back to `config.json`'s `max_position_embeddings` — the default model reports 512 there against a real `max_seq_length` of 128, so that would diverge from sentence-transformers. A debug log saying the length was defaulted is enough.

### 5. The pooling guard's clean message reaches only `wren memory index`

`memory store`, `recall` and `watch` all embed, as does the agent path through `WrenMemory`, and none of them catch `UnsupportedPoolingError` — so a CLS-pooled model still tracebacks there. Related: `_read_json` catches `OSError` but not `json.JSONDecodeError`, which is a `ValueError`, so a corrupt `1_Pooling/config.json` would resurface as `Malformed manifest:` — the same misattribution this round just closed.

---

On severity for 1, since I'm calling it non-blocking but asking for it here: the deciding factor is the cost ratio, not the risk. It is roughly five lines in a file this PR already rewrites, the vectors do not change so the parity gate needs no adjustment, and deferring it saves close to nothing while knowingly shipping an unbounded-memory path. What tips it is who the extra is for: people who choose it to go from 817 MB to 133 MB are disproportionately running under a container memory limit, on a CI runner, or on a small VM. Moving the cost from install size to runtime OOM lands it on precisely the audience the extra exists to serve, and they find out by pointing it at a real warehouse.

If you'd rather ship as-is, that's safe and I won't hold the PR for it — but please open an issue rather than leaving it in this thread, because CI cannot see it and the largest batch under test is two rows.

audi0417:
All five are in the PR, in three commits (`320c072`, `15498ff`, `7b00446`). Four are small enough that an issue would cost more to track than to fix; the fifth I've partly declined and partly reframed, details below.

### 1. Chunking — `320c072`

`_encode` now loops at `_ENCODE_BATCH_SIZE = 32`, named against `SentenceTransformer.encode`'s default so the reason survives the next reader.

Two tests, because the two claims are independent. `test_a_large_batch_is_split_before_it_reaches_the_model` records batch shapes at a fake session and asserts `[32, 32, 6]` for 70 rows — on `f64608a` that assertion reads `[70]`, which is the finding stated as a test. `test_chunking_leaves_the_vectors_unchanged` runs the same 70 rows chunked and one at a time and compares with `assert_array_equal`, not `allclose` — the vectors are identical, not merely close, which is what your "the parity gate needs no adjustment" rests on.

Confirmed against the real default model rather than only the fakes: parity `1 passed, 128 deselected in 29.22s`.

### 2. Preferring onnx on importability alone — `15498ff`

Taking your second option, not the fallback, and I want to be explicit that this is a choice against the symmetry argument you made.

A missing extra is decided at import, deterministically. This one isn't: a repo with no ONNX export and a briefly unreachable HF both arrive as `OSError`. Falling back would mean that a network blip loads 817 MB of torch on a machine that installed `memory-onnx` specifically to not have it — landing the failure mode on the same audience you identified in your closing paragraph, just through a different door. So it raises `MissingOnnxExportError` naming both `WREN_EMBEDDING_BACKEND=sentence-transformers` and `WREN_ONNX_MODEL_FILE`, and the user makes that call knowingly.

If you'd rather have the fallback anyway, say so and I'll switch it — it's a smaller change than this paragraph.

### 3. `memory-onnx` documented — `7b00446`

Install line in the extras list, and a note in §6: same vectors so existing indexes stay valid, onnx wins when both are present, `WREN_EMBEDDING_BACKEND` overrides, `wren memory status` reports which is live. Kept to the extras surface as you framed it — the env var appears as prose in §6, not as a new documented API.

### 4. Defaulted sequence length — `15498ff`

Debug log, and the docstring records *why* `config.json`'s `max_position_embeddings` is not the fallback — the default model reports 512 there against a real 128, so reading it would diverge from sentence-transformers. That reasoning was in your comment and nowhere in the code; the next person to see the 128 will otherwise repeat it.

### 5. Two separate things, handled differently — `15498ff`

**The corrupt-JSON half is a fix, and a larger one than "catch `JSONDecodeError`".** Catching it and returning `None` would be worse than the status quo: `_require_mean_pooling` reads `None` as "no pooling config" and returns early, so a corrupt `1_Pooling/config.json` would wave a CLS-pooled model through with quietly wrong vectors — the exact outcome the guard exists to prevent, arrived at from the other side. It now raises, naming the cached path so the user can delete it. `_read_json` also no longer wraps `_hf_file` in the same `try`, so a genuinely absent file and a corrupt one are distinguishable.

**The reach half is a fix.** `UnsupportedPoolingError` and `MissingOnnxExportError` now share an `OnnxBackendError` base — still a `RuntimeError`, so `index` keeps its ordering against the `ValueError`/manifest branch, and there's a test asserting that hierarchy holds precisely because a later widening could break it silently. One `_report_backend_errors()` context manager covers `index`, `store` and `recall`. Six parametrized CLI tests, each error type against each command; five of the six fail on `f64608a`.

**On `watch` — I think the premise is off, and the real problem is elsewhere.** It never tracebacked: `watch.py:126` catches `Exception` and reports `reindex-error`. What actually happens is worse in a quieter way — it retries a permanent misconfiguration on every poll, and `on_event` receives only the event name, so the message naming the way out is dropped on the floor. Wrapping `_reindex` doesn't fix that; either the loop learns that some failures are terminal, or `on_event` learns to carry the exception. Both change `watch_loop`'s contract, which isn't something to bury in an embedding-backend PR — the same call I made on the seven `persist-credentials` checkouts. Happy to open it as an issue, or take it as a follow-up PR, whichever you prefer.

### Verification

macOS arm64, both extras installed. `tests/unit`: **1370 passed, 3 skipped**. `test_memory.py` + `test_index_backend.py`: **140 passed** under `WREN_EMBEDDING_BACKEND=onnx` and **140 passed** under `=sentence-transformers`. Parity against the real default model: **1 passed**. `ruff format --check` and `ruff check` clean on `src/wren/memory/` and the touched tests.

Net for the round: +11 tests. Nothing in the data format, the API, or the vectors moved.

goldmedal:
Round 2. Two things before I approve, plus a correction of mine.

I re-ran your claims rather than taking them: the new test file against `f64608a` gives 9 of 12 failing, the batch-shape assertion reads `[70]` there, and five of the six CLI cases fail — all as you described. Parity still passes against the real default model after chunking (`1 passed, 143 deselected in 20.57s`), which is what makes "the vectors did not move" checkable rather than asserted.

### 1. `memory tests` is red, and it is this PR's own test

```
tests/unit/test_memory.py::TestOnnxEmbeddings::test_a_repo_without_an_onnx_export_names_the_way_out FAILED
E   ModuleNotFoundError: No module named 'onnxruntime'
1 failed, 142 passed, 1 skipped in 23.14s
```

Reproduced locally on a sentence-transformers-only venv, same line: `_runtime` does `import onnxruntime` at `embeddings.py:296` before it reaches the `_hf_file` you stubbed. It is the only new test that enters the real `_runtime` — the others replace it wholesale — so `TestOnnxEmbeddings` has no class-level guard and this one went past it.

`pytest.importorskip("onnxruntime", reason="wren[memory-onnx] not installed")` in the test body matches how `TestOnnxVectorParity` gates itself. Worth naming where the blind spot came from: your verification ran with both extras installed, which is the one configuration that hides this. The `memory` job installs the default shape.

### 2. `fetch` is still unwrapped, and the tests encode the gap

`wren memory fetch -q …` → `get_context` → above the character threshold → `_search_schema` → embeds the query. Its docstring is "Get schema context for an LLM" and it is the second command in README §6, so it is a likely place to meet a backend error. `mem_store.load_queries(md_pairs, upsert=True)` at `cli.py:269` is outside the `with` block too — narrow, since it is only the first encode when `index_schema` embedded nothing at all, but the same gap.

I am asking for this one on scope rather than severity: `15498ff` is titled "route every onnx backend failure to the user, **from any command**", and `fetch` is a command that embeds. What makes it worth fixing now rather than later is that `TestBackendErrorsReachTheUser` enumerates exactly the three commands you wrapped, so the matrix cannot point at a fourth — a later review will not catch this, because this review is the one it had to survive. Wrapping the command body, or the store boundary, closes the class instead of the instances.

### 3. On `watch` — you are right and my claim was wrong

It does not traceback. I checked before conceding: `poll_once` re-raises after `on_event("reindex-error")`, and the swallow is `watch_loop` at `watch.py:172`, which catches `Exception`, emits `"error"`, sleeps and continues. So the behaviour is what you describe — a permanent misconfiguration retried every interval, with the message that names the way out dropped on the floor — and your two options are the right shape. Agreed it does not belong in this PR. Please open the issue, and write it against `watch_loop:172` rather than `poll_once`, which only re-raises.

`memory tests` green and `fetch` covered, and I will approve.

audi0417:
Both fixed, in `2e44632` and `0e8162b`. The `watch` issue is #2724.

### 1. The red test is mine, and so is the reason it got through

You diagnosed it exactly: I verified with both extras installed, which is the one configuration that hides it. I reproduced it in a `uv sync --locked --extra memory` venv before touching anything — same test, same `embeddings.py:296` — then gated it the way `TestOnnxVectorParity` gates itself.

The process fix matters more than the one-liner, so I've changed how I verify: every number below comes from three venvs built to match the three CI jobs, not from mine. That immediately paid for itself — my new `_read_json` test constructed a real `RemoteEntryNotFoundError`, which needs a live HTTP response object, so it imported `requests`. `huggingface-hub` 1.8 depends on `httpx`, not `requests`; `requests` reaches all three venvs transitively through other packages, so it would have passed. But nothing in `memory-onnx`'s declared closure requires it, so it's a test that passes for a reason unrelated to what it tests. Replaced with a stand-in that reproduces the hierarchy, which is the only property under test.

### 2. `fetch` — taking the wider option, because the narrow one has the same shape as the bug

You're right that wrapping the fourth site leaves the fifth, and that the matrix can only ever enumerate what exists. So rather than adding `fetch` and `load_queries` to the list, the handling moved to `_MemoryGroup.invoke`. The three call-site wrappers are gone and `_report_backend_errors` is deleted — cli.py is net simpler than before the fix.

Two things worth flagging, both with tests:

`typer.Exit` subclasses `RuntimeError`. A handler written against `RuntimeError` — the obvious way to write this — would rewrite every exit code in the sub-app to 1, silently. Scoped to `OnnxBackendError`, with `test_an_ordinary_exit_code_is_not_swallowed` holding it, because the next person to widen this will not know that.

`test_an_unwrapped_command_is_still_covered` registers a command that has no error handling of its own and asserts it is covered anyway. That is the claim "closes the class" actually makes, so it seemed worth asserting rather than describing.

### 3. CodeRabbit's `_read_json` finding is real, and it is the same hole a third time

It flagged that `_read_json` cannot distinguish a genuine 404 from a failed lookup. It's right, and it's worse than it reads: `_hf_file` falls back from cache to network, so a dropped connection, a rate limit or an offline run all arrive as `OSError` and were all reported as "file absent" — which `_require_mean_pooling` reads as "assume mean pooling". A network blip would index a CLS-pooled model with quietly wrong vectors. Same outcome as the corrupt-JSON case, reached from a third direction.

Now only a remote 404 counts as absent (`EntryNotFoundError` and not `LocalEntryNotFoundError` — the pair, rather than `RemoteEntryNotFoundError`, because the declared floor is `huggingface-hub>=0.23` and the class did not exist before 1.x; the discrimination holds in both).

`_max_seq_length` deliberately still degrades to the default when presence is unknown, and this asymmetry is the point rather than an oversight: an ONNX-native mirror legitimately ships no `sentence_bert_config.json`, so an offline run with the model otherwise cached must not die on being unable to confirm that. Getting the length wrong costs recall; getting the pooling wrong corrupts the index. Only the second is worth failing on.

I did not take CodeRabbit's other suggestion (wrapping `load_queries` specifically) because the group handler covers it and every sibling.

### 4. `watch` — #2724

Written against `watch_loop:172` as you asked, with `poll_once` noted only as the re-raiser. It separates the two fixes: carrying the exception through `on_event` so the reason survives, and letting `OnnxBackendError` terminate the loop rather than be retried forever. I've flagged that the second has a real design choice in it — how wide the terminal set should be — and asked which shape you want before writing it, since anything broader than `OnnxBackendError` risks killing a watcher over exactly the transient failure the current code was written to survive.

### Verification

Three venvs matching the three CI jobs:

| venv | `test_memory.py` |
|---|---|
| `--extra memory` (mirrors `memory tests`) | 136 passed, 2 skipped |
| `--extra memory-onnx` (mirrors `memory tests (onnx, torch-free)`) | 131 passed, 7 skipped |
| both extras | 149 passed (with `test_index_backend.py`) |

Full `tests/unit` with both extras: 1379 passed, 3 skipped. `ruff format --check` and `ruff check` clean on `src/wren/memory/` and the touched tests. On `f64608a`, the previously-red test is the one that now skips rather than fails.

+14 tests this round. No data format, API or vector change.

goldmedal:
Round 3. Two things, both about the same call: what an `OSError` from the hub means.

### 1. Raise the floor to `huggingface-hub>=0.25`

Choosing `EntryNotFoundError` + `not LocalEntryNotFoundError` over `RemoteEntryNotFoundError` is the right instinct about the declared floor. The reasoning stops one step short of the import path. I pulled the wheels and read the module:

```
0.23.0 | errors.py: EntryNotFoundError absent | LocalEntryNotFoundError absent   (only "from requests import HTTPError")
0.24.0 | errors.py: EntryNotFoundError absent | LocalEntryNotFoundError absent
0.25.0 | errors.py: both present
```

Before 0.25 both classes live in `huggingface_hub/utils/_errors.py` and are re-exported from `huggingface_hub.utils`; `huggingface_hub.errors` exists in 0.23/0.24 but holds unrelated classes. `pyproject.toml` declares `huggingface-hub>=0.23`, and nothing in `memory-onnx`'s closure raises that floor — `tokenizers>=0.15` allows `huggingface_hub>=0.16.4,<1.0`. So on a version this PR says it supports, every `_read_json` failure path raises `ImportError` *inside* the `except OSError` handler, replacing the error it was about to classify.

No CI job and no `uv sync` user can reach this, because the lock pins 1.8. The path that can is `pip install 'wrenai[memory-onnx]'` — the line this PR added to the README — resolving in an environment that already holds an older hub.

Bump the floor rather than moving the import to `huggingface_hub.utils`: `errors` is where these live now, and `utils` is the compatibility surface, so the pin is the honest half of the fix.

### 2. `_runtime` still reads every `OSError` as "no ONNX export"

`0e8162b` fixed this classification in `_read_json` and left the same conflation forty lines below it:

```python
try:
    onnx_path = _hf_file(self._repo_id, onnx_file)
except OSError as e:
    raise MissingOnnxExportError(
        f"'{onnx_file}' could not be fetched for '{self._repo_id}': {e}. "
        "The onnx embedding backend needs a repo that publishes an ONNX export. ..."
    )
```

So the file now holds two `OSError` semantics: `_read_json` distinguishes a remote 404 from an unreachable hub, and `_runtime` does not. A rate limit or a dropped connection tells the user that a repo which does publish an ONNX export does not — the same "presence unknown reported as absence" your own commit message describes, one call site over.

Two things keep it from being worse than it reads, and one hides it: the message interpolates `{e}`, so the real cause is printed immediately after the wrong headline, and the way out it names still works. What hides it is the test — `_missing` raises a bare `OSError("404 Client Error")`, so it passes identically for a connection error and cannot see the distinction.

Asking for it now rather than as a follow-up only because the discrimination is already written in this file: the same `isinstance` pair, and a message for the unknown-presence branch that says retry rather than switch backends.

`memory tests` is green and `fetch` is covered structurally — that part I'd merge as it stands. Note your #2724 question about the terminal set is still waiting on me; I'll answer it there.

audi0417:
Both fixed in `2cb00a1d`, taking your proposed shape for each.

### 1. The hub floor — you're right, and my reasoning stopped exactly where you said

I read the wheels rather than take the version claim:

```
0.23.0  errors.py present | EntryNotFoundError=False  LocalEntryNotFoundError=False   -> utils/_errors.py
0.24.0  errors.py present | EntryNotFoundError=False  LocalEntryNotFoundError=False   -> utils/_errors.py
0.25.0  errors.py present | both present                                              -> utils/_errors.py gone
```

So on a version the PR declared support for, both call sites raise `ImportError` inside the `except OSError` handler and destroy the error they were about to classify.

Bumped to `>=0.25` with the reason in the pin, rather than importing from `huggingface_hub.utils`, for the reason you gave. `uv lock` regenerated: one line, resolution unchanged since it was already at 1.8.

Added a test that reads the declared pin out of `pyproject.toml` and asserts `>= (0, 25)`. The failure is invisible to every CI job — the lock pins 1.8 — so a test against the installed version would assert nothing. Only the declared floor can be checked, and it is the thing that can regress.

### 2. `_runtime` — the duplication is the defect, so that is what got fixed

The discrimination is now one function, `_is_absent_from_repo`, called by both sites. Writing it inline twice is precisely what let the two drift apart inside a single commit, so adding a third copy in `_runtime` would have preserved the mechanism that produced this finding. `_runtime` raises `MissingOnnxExportError` only on a 404, and a plain `OnnxBackendError` naming retry otherwise.

You identified what hid it, and that is the more useful half of the report. `_missing` raised a bare `OSError("404 Client Error")` — a string the code never reads, since it switches on type — so a connection error and a 404 were the same input to that test. Split into the two cases: one asserting the 404 path names the backend override, one asserting the unreachable path is *not* a `MissingOnnxExportError` and says retry.

I checked the rest of the module for the same pattern. The other `OSError` handlers — `embeddings.py:113` and `_hf_file` itself — are local-then-online retries rather than classifications, so `_runtime` was the last one.

### Verification

| venv | `test_memory.py` |
|---|---|
| `--extra memory` | 137 passed, 3 skipped |
| `--extra memory-onnx` | 133 passed, 7 skipped |
| both extras | 151 passed (with `test_index_backend.py`) |

Full `tests/unit` with both extras: 1381 passed, 3 skipped. `ruff format --check` and `ruff check` clean. +3 tests, −1 replaced.

No rush on #2724 — the terminal-set question there is a real design call and I would rather have your answer than guess at it.

goldmedal (approved):
Thanks @audi0417 for working on this. LGTM 👍

PR #2719, opened 2026-09-01
fix(wren): make resolve_model_name's case-insensitive fallback deterministic

AmirF194:
## Summary

`resolve_model_name`'s unquoted case-insensitive fallback returns whichever
candidate the input `set` happens to iterate first when two model names
differ only in case. For a `set` of strings that order depends on string
hashing, which is randomized per process (`PYTHONHASHSEED`), so an identical
query can bind a different model across restarts of the same server. This
tracks the lexicographically smallest match in the same single pass instead,
so the result no longer depends on the interpreter's hash seed.

```python
from wren.policy import resolve_model_name
resolve_model_name("ORDERS", False, {"Orders", "orders"})
```

Running that same line in 8 fresh subprocesses on current `main`:

```
['Orders', 'orders', 'orders', 'orders', 'Orders', 'Orders', 'orders', 'orders']
```

Same input, same manifest, different answer per process: 5 runs resolved to
`orders`, 3 resolved to `Orders`. On this branch all 8 return `Orders`.

`test_case_sensitivity.py` already documents this pair as a legitimate manifest
shape, and its own `test_resolve_model_name_dual_case` accepts either result
for this exact case ("`USERS` unquoted may pick either manifest entry
depending on set iteration order"). This change does not reverse that: an
ambiguous unquoted reference still resolves to *a* case-insensitive match, and
that test still passes unmodified. It only removes the "depending on set
iteration order" part, so the same query gives the same answer on every run.

## How is it tested?

- `test_resolve_model_name_case_collision_is_deterministic` (new, in
  `tests/unit/test_policy.py`) runs the lookup 8 times and asserts the same
  result each time, mirroring the subprocess reproduction above.
- The existing `tests/unit/test_policy.py` and `tests/unit/test_case_sensitivity.py`
  suites (107 tests) pass unmodified, including the dual-case tests that
  intentionally keep `Users`/`users` as two distinct models.
- `ruff format --check src/` and `ruff check src/` pass.
- Ran the full `tests/unit/` suite (excluding the `memory`/`mcp` extras, matching
  CI): 1248 passed, 2 skipped, 3 pre-existing failures in
  `test_served_content_guard.py` unrelated to this change and present on `main`
  in the same environment (they need the `mcp`/`ui` extras to register every CLI
  subcommand the guard checks against).
- Not benchmarked: the loop this touches was already a full scan of
  `model_names` per unquoted reference, so this adds one `str` comparison only
  on the (rare) case-insensitive hit path, no new allocation or second pass.

## Duplicate check

`competing-prs.sh` swept all 72 open PRs; two drafts by the same author
(#2551, #2552) touch this file but only insert an unrelated `basic_safety_check`
function immediately after `resolve_model_name` and edit an error message
elsewhere in `_check_functions`. Neither touches the fallback loop this PR
changes, and no open PR modifies `resolve_model_name` itself.

## Summary by CodeRabbit

* **Bug Fixes**
  * Table references that match multiple model names differing only by letter case now resolve consistently by selecting the lexicographically smallest match.
  * Exact matches, quoted names, case-insensitive fallback behavior, and unmatched references continue to work as expected.

* **Tests**
  * Expanded coverage for matching precedence, case sensitivity, missing models, and deterministic results across different runtime environments.

### DIFF
--- a/core/wren/src/wren/policy.py
+++ b/core/wren/src/wren/policy.py
@@ -161,6 +161,14 @@ def resolve_model_name(
     case-sensitively; an unquoted identifier prefers an exact case match but
     falls back to a case-insensitive scan. Returns ``None`` if no model
     matches.
+
+    When an unquoted reference matches more than one model case-insensitively
+    (e.g. both ``Users`` and ``users`` are defined), the one picked is the
+    lexicographically smallest rather than whichever ``model_names`` happens
+    to iterate first: for a ``set`` of strings that iteration order is not
+    the insertion order, it depends on string hashing and varies across
+    processes under ``PYTHONHASHSEED``, so the same query could silently
+    bind a different model on every run.
     """
     model_set = (
         model_names if isinstance(model_names, (set, frozenset)) else set(model_names)
@@ -170,10 +178,14 @@ def resolve_model_name(
     if quoted:
         return None
     name_lower = name.lower()
+    # Single pass, no intermediate list: a tie only needs `<` against the
+    # best candidate seen so far, same cost as the exact-match scan this
+    # replaces.
+    best: str | None = None
     for candidate in model_set:
-        if candidate.lower() == name_lower:
-            return candidate
-    return None
+        if candidate.lower() == name_lower and (best is None or candidate < best):
+            best = candidate
+    return best
 
 
 # Statement shapes that are read-only queries. ``SetOperation`` is the base class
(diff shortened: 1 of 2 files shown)


AmirF194:
Good catch, the loop was proving the wrong thing. Within one process the old buggy code also returns the same answer every time (fixed hash seed, stable iteration order), so it would have passed on unfixed code too. Pushed a version that spawns a fresh subprocess per PYTHONHASHSEED instead, which fails on the pre-fix code and passes on this one.

goldmedal (approved):
Thanks @AmirF194 👍

AmirF194:
Appreciate the review. The existing case_sensitivity test made it clear the dual-case behavior was intentional, so narrowing this to determinism only rather than rejecting the collision was the right call.

PR #2726, opened 2026-09-07
fix(bigquery): support Application Default Credentials

### LINKED ISSUE #2660: ADC support for Bigquery connection
The current documentation and feature set for connecting wren with Bigquery requires a service account https://docs.getwren.ai/oss/guide/connect/bigquery#prerequisites

This might be fine for a long running production service.

However, in many enterprises short lived credentials are the preferred approach. And normal users who would quickly want to spin up a demo do not have the permissions to create service accounts. 

It would be neat that ADC the google application default credentials would be supported to within the context of that user access bigquery.

## details

out of scope

- automatic renewal of the credentials. it is explicitly OK that people have to re-auth with the gcloud CLI manually

AmirF194:
## Summary
Close https://github.com/Canner/WrenAI/issues/2660

`BigQueryConnectionInfo.credentials` was a required base64 service-account JSON, and `BigQueryConnector.__init__` always decoded it. A user on Application Default Credentials (`gcloud auth application-default login`, workload identity, or the GCE/Cloud Run metadata server), or without permission to create a service-account key, could not connect at all. `credentials` is now optional; when it is absent, the connector calls `google.auth.default()` with the same scopes used for a service account.

```python
>>> from wren.model import BigQueryDatasetConnectionInfo
>>> BigQueryDatasetConnectionInfo(project_id="my-project", dataset_id="my_dataset")
pydantic_core._pydantic_core.ValidationError: 1 validation error for BigQueryDatasetConnectionInfo
credentials
  Field required [type=missing, input_value={'project_id': 'my-projec...taset_id': 'my_dataset'}, input_type=dict]
```

`google-auth` is already listed alongside `google-cloud-bigquery` under the `bigquery` extra, so no new dependency is needed.

## How is it tested?

- New `tests/unit/test_bigquery_adc.py`, run against `c59ec230` (unmodified `main`) and against this branch on python 3.11: `test_connection_info_allows_missing_credentials` and `test_connector_falls_back_to_adc_when_credentials_absent` fail on `main` (`ValidationError`) and pass here; `test_connector_uses_service_account_when_credentials_present` passes on both, pinning that the credentialed path is unchanged.
- `tests/unit/test_bigquery_semicolon.py` (existing BigQuery suite) stays green.
- `ruff format --check` and `ruff check` clean on the changed files.
- I have not tried this against a real GCE/Cloud Run metadata server or an actual `gcloud` session; `google.auth.default()` is stubbed in the test, the same way the existing suite stubs `google.cloud.bigquery`.

## Duplicate check

Swept all 68 open PRs against both changed files. #2258 (YTsaurus connector) also touches `model/__init__.py` but only inserts a new class after `ClickHouseConnectionInfo`; no overlap with the BigQuery classes.

## Summary by CodeRabbit

* **New Features**
  * BigQuery connections can use Application Default Credentials when service-account credentials are omitted.
  * Supported options include gcloud credentials, workload identity, and cloud platform metadata services.
  * Service-account credentials remain supported for existing configurations.
  * Connection settings now explain available authentication methods and when credentials may be left blank.
  * BigQuery external tables using Drive or Sheets still require a service account when using gcloud-based Application Default Credentials.

### DIFF
--- a/core/wren/src/wren/connector/bigquery.py
+++ b/core/wren/src/wren/connector/bigquery.py
@@ -21,26 +21,33 @@ def _apply_limit(sql: str, limit: int) -> str:
     return f"SELECT * FROM ({cleaned}) AS _sub LIMIT {limit}"
 
 
+_SCOPES = [
+    "https://www.googleapis.com/auth/drive",
+    "https://www.googleapis.com/auth/cloud-platform",
+]
+
+
 class BigQueryConnector(ConnectorABC):
     def __init__(self, connection_info):
         from google.cloud import bigquery  # noqa: PLC0415
-        from google.oauth2 import service_account  # noqa: PLC0415
 
         self.connection_info = connection_info
-        credits_json = loads(
-            base64.b64decode(connection_info.credentials.get_secret_value()).decode(
-                "utf-8"
+        if connection_info.credentials is not None:
+            from google.oauth2 import service_account  # noqa: PLC0415
+
+            credits_json = loads(
+                base64.b64decode(connection_info.credentials.get_secret_value()).decode(
+                    "utf-8"
+                )
             )
-        )
-        credentials = service_account.Credentials.from_service_account_info(
-            credits_json
-        )
-        credentials = credentials.with_scopes(
-            [
-                "https://www.googleapis.com/auth/drive",
-                "https://www.googleapis.com/auth/cloud-platform",
-            ]
-        )
+            credentials = service_account.Credentials.from_service_account_info(
+                credits_json
+            )
+            credentials = credentials.with_scopes(_SCOPES)
+        else:
+            import google.auth  # noqa: PLC0415
+
+            credentials, _ = google.auth.default(scopes=_SCOPES)
         client = bigquery.Client(
             credentials=credentials,
             project=connection_info.get_billing_project_id(),
--- a/core/wren/src/wren/connector/bigquery.py
+++ b/core/wren/src/wren/connector/bigquery.py
@@ -32,7 +32,7 @@ class BigQueryConnector(ConnectorABC):
         from google.cloud import bigquery  # noqa: PLC0415
 
         self.connection_info = connection_info
-        if connection_info.credentials is not None:
+        if connection_info.credentials:
             from google.oauth2 import service_account  # noqa: PLC0415
 
             credits_json = loads(
--- a/core/wren/src/wren/model/__init__.py
+++ b/core/wren/src/wren/model/__init__.py
@@ -38,8 +38,14 @@ class BaseConnectionInfo(BaseModel):
 
 
 class BigQueryConnectionInfo(BaseConnectionInfo):
-    credentials: SecretStr = Field(
-        description="Base64 encode `credentials.json`", examples=["eyJ..."]
+    credentials: SecretStr | None = Field(
+        default=None,
+        description=(
+            "Base64 encode `credentials.json`. Omit to use Application Default "
+            "Credentials (gcloud auth application-default login, workload "
+            "identity, or the GCE/Cloud Run metadata server)."
+        ),
+        examples=["eyJ..."],
     )
     job_timeout_ms: int | None = Field(default=None)
 
--- a/core/wren/src/wren/model/__init__.py
+++ b/core/wren/src/wren/model/__init__.py
@@ -43,7 +43,10 @@ class BigQueryConnectionInfo(BaseConnectionInfo):
         description=(
             "Base64 encode `credentials.json`. Omit to use Application Default "
             "Credentials (gcloud auth application-default login, workload "
-            "identity, or the GCE/Cloud Run metadata server)."
+            "identity, or the GCE/Cloud Run metadata server). ADC obtained via "
+            "gcloud auth application-default login does not carry the Drive "
+            "scope; BigQuery external tables over Drive/Sheets still require a "
+            "service account."
         ),
         examples=["eyJ..."],
     )
--- a/core/wren/docs/connections.md
+++ b/core/wren/docs/connections.md
@@ -75,6 +75,12 @@ Both formats are accepted. The CLI auto-flattens the envelope format.
 }
 ```
 
+`credentials` is optional: if omitted, the connector falls back to
+Application Default Credentials (`gcloud auth application-default login`,
+workload identity, or the GCE/Cloud Run metadata server). ADC obtained via
+`gcloud auth application-default login` does not carry the Drive scope;
+BigQuery external tables over Drive/Sheets still need a service account.
+
 ## Snowflake
 
 ```json
--- a/core/wren/src/wren/model/field_registry.py
+++ b/core/wren/src/wren/model/field_registry.py
@@ -94,14 +94,14 @@ _MODEL_UI_OVERRIDES: dict[str, dict[str, dict]] = {
         "credentials": {
             "input_type": "file_base64",
             "accept": ".json",
-            "hint": "Upload your GCP service account credentials.json file. It will be base64-encoded automatically.",
+            "hint": "Upload your GCP service account credentials.json file (it will be base64-encoded automatically), or leave blank to use Application Default Credentials.",
         },
     },
     "BigQueryProjectConnectionInfo": {
         "credentials": {
             "input_type": "file_base64",
             "accept": ".json",
-            "hint": "Upload your GCP service account credentials.json file. It will be base64-encoded automatically.",
+            "hint": "Upload your GCP service account credentials.json file (it will be base64-encoded automatically), or leave blank to use Application Default Credentials.",
         },
     },
     "SnowflakeConnectionInfo": {
--- a/core/wren/tests/unit/test_bigquery_adc.py
+++ b/core/wren/tests/unit/test_bigquery_adc.py
@@ -155,3 +155,21 @@ def test_connector_falls_back_to_adc_when_credentials_absent() -> None:
     ]
     assert _client_calls[0]["credentials"] is _adc_credentials
     assert _client_calls[0]["project"] == "my-project"
+
+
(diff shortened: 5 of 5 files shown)


AmirF194:
Good catch. Pushed 72deef6: the stub is now an autouse fixture using monkeypatch.setitem instead of the module-level setdefault, so it overrides an already-loaded google (protobuf/grpc use that same namespace) and restores the prior state after each test. Ran it alongside test_bigquery_semicolon.py and test_athena_connector.py, all 40 pass.

core/wren/src/wren/connector/bigquery.py
```diff
@@ -21,26 +21,33 @@ def _apply_limit(sql: str, limit: int) -> str:
...
-            ]
-        )
+            credentials = service_account.Credentials.from_service_account_info(
+                credits_json
+            )
+            credentials = credentials.with_scopes(_SCOPES)
+        else:
+            import google.auth  # noqa: PLC0415
+
+            credentials, _ = google.auth.default(scopes=_SCOPES)
```
goldmedal, line 50: `scopes=_SCOPES` is silently a no-op for the `gcloud auth application-default login` flavour of ADC -- which is the case the linked issue leads with.

`google.auth.default()` routes through `with_scopes_if_required`, guarded by `isinstance(credentials, Scoped) and credentials.requires_scopes`. gcloud user credentials (`google.oauth2.credentials.Credentials`) are not `Scoped` at all, so the guard short-circuits on the first condition and the credential comes back untouched.

Measured on this branch against a live gcloud ADC session, introspecting the actual issued token:

```
requested _SCOPES  : ['.../auth/drive', '.../auth/cloud-platform']
returned class     : google.oauth2.credentials.Credentials
isinstance(Scoped) : False          <- guard short-circuits here
effective .scopes  : None

ACTUAL token scopes (tokeninfo endpoint):
   - email
   - https://www.googleapis.com/auth/cloud-platform
   - https://www.googleapis.com/auth/sqlservice.login
   - https://www.googleapis.com/auth/userinfo.email
   - openid

drive scope present in the real token? False
```

The same `_SCOPES` list on the service-account path *does* apply (`creds.scopes == ['.../auth/drive', '.../auth/cloud-platform']`, also verified live). So the branches read as equivalent but are not:

| ADC flavour | credential class | `_SCOPES` applied? |
| --- | --- | --- |
| `GOOGLE_APPLICATION_CREDENTIALS` -> SA file | `oauth2.service_account.Credentials` | yes (verified live) |
| GCE / Cloud Run metadata | `auth.compute_engine.Credentials` | yes (`requires_scopes` = `not self._scopes`) |
| `gcloud auth application-default login` | `oauth2.credentials.Credentials` | **no -- not `Scoped`, silently ignored** (verified live) |

This is deliberate upstream: the comment above the `checkers` tuple in `_default.py` reads *"Avoid passing scopes here to prevent passing scopes to user credentials"*, because a user token's scopes are fixed at consent time. And gcloud's ADC consent cannot grant Drive -- the default set is `openid` / `userinfo.email` / `cloud-platform` / `sqlservice.login`, and the gcloud docs state that adding Drive requires supplying your own OAuth client ID via `--client-id-file`.

Practical consequence: on gcloud ADC ordinary BigQuery tables work fine (I queried them), but a BigQuery **external table backed by Google Drive or Sheets** cannot be read, whereas the service-account path -- which deliberately requests `auth/drive` -- can. I had no Drive-backed table available, so the 403 itself is inferred from the missing scope rather than observed.

Not a reason to hold the PR; ADC without Drive is strictly better than no ADC. But it deserves a sentence in the `credentials` description or the docs, e.g. *"ADC obtained via `gcloud auth application-default login` does not carry the Drive scope; BigQuery external tables over Drive/Sheets still require a service account."*

core/wren/src/wren/connector/bigquery.py
```diff
@@ -21,26 +21,33 @@ def _apply_limit(sql: str, limit: int) -> str:
...
 class BigQueryConnector(ConnectorABC):
     def __init__(self, connection_info):
         from google.cloud import bigquery  # noqa: PLC0415
-        from google.oauth2 import service_account  # noqa: PLC0415
 
         self.connection_info = connection_info
-        credits_json = loads(
-            base64.b64decode(connection_info.credentials.get_secret_value()).decode(
-                "utf-8"
+        if connection_info.credentials is not None:
```
goldmedal: Minor: `is not None` treats an empty string as "use a service account". Measured on the real path:

```
BigQueryDatasetConnectionInfo(project_id=..., dataset_id=..., credentials="")
-> json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

(whitespace-only behaves the same). The realistic route is a profile with `"credentials": "${SOME_VAR}"` where the variable is *set but empty* -- `_expand_obj` raises only when the name is absent, so an empty value flows straight through, and someone switching to ADC by blanking their env var lands on a JSON parse error instead of ADC.

`SecretStr("")` is falsy, so this is a one-character fix:

```suggestion
        if connection_info.credentials:
```

goldmedal (commented):
**LGTM.** I checked this out at `72deef68`, installed it with the `bigquery` extra, and exercised both auth branches against a real BigQuery dataset (TPC-H `orders`, 15 000 rows). It works. The three inline comments below are documentation-level; none of them block.

### What I verified

| Path | Result |
| --- | --- |
| `credentials` omitted -> `google.auth.default()` | `INFORMATION_SCHEMA.TABLES` listing, `SELECT *` with `limit=3` (the `_apply_limit` wrap), `dry_run`, and `COUNT(*) = 15000` all succeed |
| `credentials` present -> service account | identical results; scopes still `[drive, cloud-platform]`, path unchanged |
| End-to-end through the CLI | `wren profile add --from-file` with a BigQuery profile carrying only `project_id` + `dataset_id` prints `Connection validated`, and `wren -m mdl.json -s "SELECT o_orderstatus, COUNT(*), SUM(o_totalprice) ... GROUP BY 1"` returns correct rows through the semantic layer |
| On `main` | `BigQueryDatasetConnectionInfo(project_id=..., dataset_id=...)` -> `ValidationError: credentials`, confirming the regression this repairs |

`ruff format --check src/` and `ruff check src/` clean, `pytest tests/unit/test_bigquery_adc.py` 3 passed, all CI jobs green.

Gating on `credentials is not None` keeps the service-account path byte-for-byte identical; `required` in `field_registry` is derived from `is_required()`, so the connect form and `wren connect --interactive` relax automatically; and `test_save_skips_empty_fields` means a blank field in the web form is dropped rather than submitted as `""`. Nice, tight change.

### Nits

- `BigQueryProjectConnectionInfo` isn't covered by the new tests. I confirmed it works by inheritance, but a one-line parametrization would pin it.
- The autouse fixture replaces the top-level `google` entry in `sys.modules` for each test. Harmless today, but any future import of a real `google.*` submodule from inside these tests would resolve against the stub -- a comment there might save someone ten minutes.
- On user-credential ADC without a quota project, every connection prints Google's `UserWarning: ... authenticated using end user credentials ... without a quota project`. Environmental rather than a defect here, but a docs line pointing at `gcloud auth application-default set-quota-project <project>` would spare users the noise.

core/wren/src/wren/model/__init__.py
```diff
@@ -38,8 +38,14 @@ def _normalize(value):
 
 
 class BigQueryConnectionInfo(BaseConnectionInfo):
-    credentials: SecretStr = Field(
-        description="Base64 encode `credentials.json`", examples=["eyJ..."]
+    credentials: SecretStr | None = Field(
```
goldmedal, line 41: The new ADC wording never reaches a user.

`FieldDef.hint` is populated only from `_MODEL_UI_OVERRIDES`, never from the Pydantic `description`, so the connect form and `wren connect --interactive` still show *"Upload your GCP service account credentials.json file"* with nothing to indicate that leaving it blank now means ADC. `core/wren/docs/connections.md` likewise still lists `credentials` in the BigQuery example unconditionally, and the public connect/BigQuery guide still states a service account is a prerequisite -- the exact wording the linked issue opens with.

Worth updating the `credentials` override hint for both `BigQueryDatasetConnectionInfo` and `BigQueryProjectConnectionInfo` in `field_registry.py`, plus the docs example, in this PR -- otherwise the feature ships undiscoverable.

AmirF194:
Pushed 000a013f.

`is not None` -> `if connection_info.credentials:` on line 35. Only other read of that field in the file is `.get_secret_value()` inside the same branch, so nothing else depended on the old check. Added `test_connector_falls_back_to_adc_when_credentials_is_empty_string`: constructs `BigQueryDatasetConnectionInfo(..., credentials="")` and asserts it hits `google.auth.default()` instead of the service-account path. With the old `is not None` check it reproduces your `JSONDecodeError` exactly; with the fix it passes.

Updated both `credentials` hints in `field_registry.py` (`BigQueryDatasetConnectionInfo` and `BigQueryProjectConnectionInfo`) to mention ADC, and added a paragraph to the BigQuery section of `connections.md` stating `credentials` is optional and what it falls back to.

Took the Drive-scope note too, one sentence in the `credentials` field description in `model/__init__.py` plus the same line in `connections.md`: ADC via `gcloud auth application-default login` does not carry the Drive scope, so Drive/Sheets-backed external tables still need a service account.

Left `BigQueryProjectConnectionInfo` test parametrization and the sys.modules stub comment from your top-level nits alone; those weren't in the three inline threads and I didn't want to stretch this PR further without checking with you first.

ruff format/check clean, `pytest tests/unit/test_bigquery_adc.py tests/unit/test_bigquery_semicolon.py` 9 passed, in a python:3.11-slim container.

goldmedal (approved):
Thanks @AmirF194 LGTM

AmirF194:
Thanks for taking it all the way through a live BigQuery run, that's more coverage than I had locally.

PR #2728, opened 2026-09-07
fix(wren): postgres LIMIT pushdown breaks on SQL ending in a trailing comment

### LINKED ISSUE #2727: fix(wren): postgres LIMIT pushdown breaks on SQL ending in a trailing comment
`PostgresConnector.query()` wraps the caller's SQL for LIMIT pushdown as `SELECT * FROM (<sql>) AS _sub LIMIT n`, built on one line (`dry_run()` does the same for its `LIMIT 0` probe). When `<sql>` ends in a single-line SQL comment, the comment swallows everything after it on the same physical line, including the closing paren, the alias, and the LIMIT clause, so the wrapped query is unbalanced and the engine rejects it with a syntax error.

Repro, no DB needed:

```python
import sqlglot
sql = "SELECT name FROM customers -- top spenders"
wrapped = f"SELECT * FROM ({sql}) AS _sub LIMIT 5"
sqlglot.parse_one(wrapped, read="postgres")
# sqlglot.errors.ParseError: Expecting ). Line 1, Col: 41.
```

Expected: a query ending in a trailing comment runs with its limit applied, same as any other query.

athena.py and snowflake.py wrap the same way and already guard against this (#2457, #2456) by putting the wrapped SQL on its own line so a trailing comment stops at the newline. postgres.py's `query()` and `dry_run()` still build the wrap on one line.

AmirF194:
Fixes #2727.

PostgresConnector wraps the caller's SQL for LIMIT pushdown and for dry_run as `SELECT * FROM (<sql>) AS _sub LIMIT n`, built on one line. When `<sql>` ends in a single-line SQL comment, the comment eats the rest of the line, including the closing paren, the alias, and the LIMIT clause, so the query fails with a syntax error instead of running with the limit applied.

athena.py and snowflake.py have the identical wrap and fixed the same defect (#2457, #2456) by putting the wrapped SQL on its own line, so a trailing comment stops at the newline instead of continuing into the wrap. This applies the same fix to postgres.py's `query()` and `dry_run()`.

The other single-line connectors (redshift, canner, duckdb, trino, oracle, clickhouse, databricks' dry_run, datafusion) have the same shape and are not touched here, to keep this PR to the one connector.

Updated the existing wrap-format assertion for the new multiline shape and added a regression test per path (query, dry_run) with a trailing comment. Ran `tests/unit/` (1252 passed, 2 skipped, one unrelated pre-existing CLI-introspection file excluded for a missing extra in my environment) and `ruff format`/`ruff check` on both changed files, all clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed PostgreSQL queries containing trailing line comments so wrapped queries continue to apply row limits correctly.
  - Corrected limited-query and dry-run behavior to prevent comments from interfering with generated SQL.
  - Preserved expected results for regular limited queries and dry runs, including queries ending with semicolons or line comments.
  - Improved generated SQL formatting to ensure trailing comments do not unintentionally consume closing clauses or limit statements.

### DIFF
--- a/core/wren/src/wren/connector/postgres.py
+++ b/core/wren/src/wren/connector/postgres.py
@@ -317,7 +317,10 @@ class PostgresConnector(ConnectorABC):
         # client-pasted statements match dry_run / limited composition rules.
         sql = strip_trailing_semicolon(sql)
         if limit is not None:
-            sql = f"SELECT * FROM ({sql}) AS _sub LIMIT {limit}"
+            # Multiline wrap so a trailing `-- line comment` in the inner SQL
+            # is terminated by the newline instead of swallowing the closing
+            # `) AS _sub LIMIT n` (same technique as athena.py/snowflake.py).
+            sql = f"SELECT * FROM (\n{sql}\n) AS _sub LIMIT {limit}"
 
         try:
             with self.connection.cursor() as cursor:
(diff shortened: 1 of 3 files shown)


goldmedal (approved):
Thanks @AmirF194 👍

Issue #2733, opened 2026-09-09
LIMIT pushdown and dry-run subquery wraps break on SQL ending in a line comment (9 connectors)


goldmedal:
#2728 fixed this for `postgres`, but the same single-line `SELECT * FROM ({sql}) ...` wrap is still in place across nine other connectors. When the user SQL ends in a `--` line comment, the closing paren and everything after it land inside that comment, so the statement is truncated mid-expression.

### Reproduction

No credentials needed -- DuckDB, on `main`:

```python
import duckdb
con = duckdb.connect()
con.execute("CREATE TABLE t AS SELECT * FROM range(10) tbl(i)")
inner = "SELECT i FROM t -- pick the column"

con.execute(f"SELECT * FROM ({inner}) AS _q LIMIT 3")        # duckdb.py:88 shape
# duckdb.ParserException: Parser Error: syntax error at end of input

con.execute(f"SELECT * FROM (\n{inner}\n) AS _q LIMIT 3")    # postgres/athena/snowflake shape
# OK, 3 rows

con.execute(inner)                                           # unlimited path, no wrap
# OK, 10 rows
```

Also confirmed live against BigQuery, where the generated SQL is

```sql
SELECT * FROM (SELECT o_orderkey FROM `<project>.<dataset>.orders` -- pick the key column) AS _sub LIMIT 3
```

```
BadRequest: 400 Syntax error: Expected ")" but got end of script at [1:104]
```

The same query without a limit succeeds and returns all rows, so the SQL itself is valid -- only the wrap is broken.

### Affected sites

| Connector | Line | Path |
| --- | --- | --- |
| `bigquery.py` | 21 | `_apply_limit` |
| `canner.py` | 256, 277 | query, dry_run |
| `clickhouse.py` | 402, 419 | query, dry_run |
| `databricks.py` | 64 | dry_run |
| `datafusion.py` | 35 | query |
| `duckdb.py` | 88, 105 | query, dry_run |
| `oracle.py` | 187, 205 | query, dry_run (`) t WHERE ROWNUM <= n`) |
| `redshift.py` | 49, 64 | query, dry_run |
| `trino.py` | 493, 513 | query, dry_run |

Verified live for BigQuery and DuckDB; the remaining connectors are identified by identical code shape rather than execution, since a `--` comment runs to end of line in every dialect involved.

[...]

PR #2734, opened 2026-09-09
fix(wren): bigquery and duckdb LIMIT wrap breaks on SQL ending in a trailing comment

AmirF194:
Refs #2733. Only bigquery.py and duckdb.py here, so this doesn't close it: see the scope note below.

bigquery.py's `_apply_limit` and duckdb.py's `query()`/`dry_run()` wrap on one line too, so a trailing SQL line comment eats the closing paren, alias and LIMIT clause the same way postgres.py did before #2728. Wrapped both onto their own line, matching the postgres.py/athena.py shape.

bigquery.py's `dry_run()` never wraps (no LIMIT there), so it needed no change.

Scoped to these two connectors: they cover both cases you verified live in the issue (DuckDB and BigQuery). canner.py, clickhouse.py, databricks.py, oracle.py, redshift.py and trino.py have the same shape and are a natural fast-follow, left out here to keep the diff small.

Added a regression test per call site with a trailing-comment SQL string; each fails on the unmodified wrap and passes with it multiline. Ran `tests/unit/` (1450 passed, 2 skipped) and `ruff format`/`ruff check` on `src/`, all clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed limited queries containing trailing SQL line comments so closing syntax and `LIMIT` clauses are processed correctly.
  * Improved consistency of wrapped query formatting across BigQuery and DuckDB connectors.

* **Tests**
  * Added coverage for limited and dry-run queries containing trailing line comments.
  * Updated expectations for multiline wrapped queries.

### DIFF
--- a/core/wren/src/wren/connector/bigquery.py
+++ b/core/wren/src/wren/connector/bigquery.py
@@ -15,10 +15,13 @@ def _apply_limit(sql: str, limit: int) -> str:
     (after stripping a trailing semicolon) short-circuits the engine and
     correctly enforces the caller's limit even when *sql* already contains an
     inner ``LIMIT`` (the outer limit always wins / can only reduce rows).
-    Avoids comment-sensitive outer-LIMIT detection heuristics.
+    Avoids comment-sensitive outer-LIMIT detection heuristics. The wrap
+    itself is multiline so a trailing ``-- line comment`` in *sql* is
+    terminated by the newline instead of swallowing the closing paren,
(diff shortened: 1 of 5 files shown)


goldmedal (approved):
Thanks @AmirF194 👍

AmirF194:
Glad it landed. The trailing-comment case was the kind of thing that only shows up once someone's SQL happens to end that way.

PR #2736, opened 2026-09-09
fix(wren): oracle, redshift and trino LIMIT wrap breaks on SQL ending in a trailing comment

AmirF194:
Fast follow to #2728 (postgres) and #2734 (bigquery/duckdb): oracle.py, redshift.py and trino.py wrap the caller's SQL for LIMIT pushdown and dry-run the same way, on one line, so a trailing SQL line comment eats the closing paren, alias/ROWNUM clause and LIMIT clause:

```
SELECT * FROM (SELECT i FROM t -- pick the column) AS _q LIMIT 3
```

Moved the wrap onto its own line for all six call sites, same technique as postgres.py.

Scoped to these three connectors on purpose. canner.py, clickhouse.py, databricks.py and datafusion.py are the remaining four from the issue's table; two open drafts (#2551, #2552) already touch those same wrap lines for an unrelated change (routing the limit through a parameterized query instead of an f-string), so I left them for whoever picks that up rather than adding a third diff to the same lines. I read both drafts before scoping this: neither fixes the trailing-comment case, they keep the identical single-line wrap.

Added a regression test per connector (query and dry_run) with SQL ending in a trailing comment, fails on the unmodified wrap and passes with it multiline. `tests/unit/` (excluding memory/mcp, matching wren-ci.yml's test-unit job): 1457 passed, 1 skipped, in a container with the oracle/redshift/trino extras installed so those tests actually run instead of skipping. ruff clean on the changed files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Oracle, Redshift, and Trino query wrappers so trailing SQL line comments no longer interfere with generated limits or closing syntax.
  - Improved handling of wrapped queries and dry runs involving trailing semicolons and comments.
  - Preserved generated query limits when SQL ends with a line comment.

- **Tests**
  - Added coverage for trailing line comments in regular queries and dry runs.
  - Updated expectations for multiline wrapped query formatting.

### DIFF
--- a/core/wren/src/wren/connector/oracle.py
+++ b/core/wren/src/wren/connector/oracle.py
@@ -184,7 +184,10 @@ class OracleConnector(ConnectorABC):
         # though engines accept multi-statement scripts elsewhere.
         sql = strip_trailing_semicolon(sql)
         if limit is not None:
-            sql = f"SELECT * FROM ({sql}) t WHERE ROWNUM <= {limit}"
+            # Multiline wrap so a trailing `-- line comment` in the inner SQL
+            # is terminated by the newline instead of swallowing the closing
+            # `) t WHERE ROWNUM <= n` (same technique as postgres.py).
+            sql = f"SELECT * FROM (\n{sql}\n) t WHERE ROWNUM <= {limit}"
         try:
             with self.connection.cursor() as cursor:
                 cursor.execute(sql)
@@ -202,7 +205,7 @@ class OracleConnector(ConnectorABC):
             try:
                 with self.connection.cursor() as cursor:
                     cursor.execute(
-                        f"SELECT * FROM ({strip_trailing_semicolon(sql)}) t "
+                        f"SELECT * FROM (\n{strip_trailing_semicolon(sql)}\n) t "
                         f"WHERE ROWNUM <= 0"
                     )
             except oracledb.DatabaseError as e:
--- a/core/wren/src/wren/connector/redshift.py
+++ b/core/wren/src/wren/connector/redshift.py
@@ -46,7 +46,9 @@ class RedshiftConnector(ConnectorABC):
     def query(self, sql: str, limit: int | None = None) -> pa.Table:
         limit = coerce_limit(limit)
         if limit is not None:
-            sql = f"SELECT * FROM ({strip_trailing_semicolon(sql)}) AS _q LIMIT {limit}"
+            # Multiline wrap so a trailing line comment in the inner SQL is
+            # terminated by the newline instead of eating the closing paren.
+            sql = f"SELECT * FROM (\n{strip_trailing_semicolon(sql)}\n) AS _q LIMIT {limit}"
         else:
             # Unlimited path also rejects trailing ``;`` for single statements
             # depending on driver/session settings — strip for consistency.
@@ -61,7 +63,7 @@ class RedshiftConnector(ConnectorABC):
     def dry_run(self, sql: str) -> None:
         with closing(self.connection.cursor()) as cursor:
             cursor.execute(
-                f"SELECT * FROM ({strip_trailing_semicolon(sql)}) AS sub LIMIT 0"
+                f"SELECT * FROM (\n{strip_trailing_semicolon(sql)}\n) AS sub LIMIT 0"
             )
 
     def close(self) -> None:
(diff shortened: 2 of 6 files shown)


goldmedal:
**Verified at `b00d75ef`.** Validated the generated SQL per dialect with sqlglot: the old single-line form is a `ParseError: Expecting )` in the oracle, trino and redshift dialects, while the multiline form parses with the `LIMIT` / `ROWNUM` predicate intact. `tests/unit/` with the oracle/redshift/trino extras installed gives 1457 passed / 1 skipped, matching the description exactly. Reverting only the three `src/` files fails all six new tests. Disjoint from #2734, so no interaction now that it has landed.

The diff is correct. Two notes about what happens *after* this PR, plus one nit.

**1. #2551 would revert #2728, #2734 and this PR in a single commit.**

Your reading is right that it keeps the single-line wrap — but it does more than leave the four remaining connectors alone. It adds `ConnectorABC._apply_limit_param()` to `base.py` built on a single-line wrap, and routes postgres, redshift, trino, duckdb, canner and athena through it. In its own diff, athena's multiline wrap is removed together with the comment that documents why it exists:

```python
-            # Multiline wrap so a trailing `-- line comment` in the inner SQL
-            # is terminated by the newline instead of swallowing the closing
-            # `) AS _wren_sub LIMIT n`. (Single-line sibling connectors don't guard this.)
-            executed = (
-                "SELECT * FROM (\n"
-                f"{strip_trailing_semicolon(sql)}\n"
-                f") AS _wren_sub LIMIT {int(limit)}"
-            )
+            executed, params = self._apply_limit_param(sql, limit, param_style="qmark")
```

#2728 and #2734 are already on `main`, so if that draft lands as written it regresses shipped fixes across every connector at once, under a title about security hardening and parameterized queries. Worth flagging on #2551 now that the shared helper needs to emit the multiline form — a one-line change there, and otherwise a silent repo-wide regression that no test would catch, since the tests guarding it are the ones being replaced.

**2. The scoping rationale does not separate the four deferred connectors from the three fixed here.**

#2551 and #2552 each touch all 14 connector files, including `oracle.py`, `redshift.py` and `trino.py`. So "two open drafts already touch those same wrap lines" applies equally to what this PR changes — the drafts do not distinguish the deferred four from the fixed three. Given they are 52 files / +627 −264, last updated six weeks ago, and carry an unrelated `MAX_ROW_LIMIT = 10000` clamp, I would rather finish canner/clickhouse/databricks/datafusion in one more PR than leave four user-visible parse errors blocked behind a draft that will need rework regardless.

**3. Nit:** `redshift.py:51` is 92 characters, over the project's `line-length = 88`. Every sibling connector hoists `stripped = strip_trailing_semicolon(sql)` above the branch, and `query()` here calls it in both branches anyway — hoisting shortens the line and matches the house shape. CI will not catch this one: `E501` is ignored and `ruff format` cannot split an f-string.

AmirF194:
Pushed the redshift.py hoist so line 51 matches the oracle/trino shape and stays under 88 chars.

Agreed on scoping: rather than wait on #2551's rework, finishing canner, clickhouse, databricks and datafusion in one more PR is the right call. Good catch on #2551 dropping the multiline wrap when it lands, that would undo this fix silently.

goldmedal:
**Re-verified at `975b0dcf`.** The hoist is the only delta and it's semantically identical — `coerce_limit()` still runs first, both branches already called `strip_trailing_semicolon`, so this just calls it once. Line 52 is now 71 chars and `ruff format --check src/` is clean. All 11 checks pass on this SHA; `BLOCKED` is the required approval, not a conflict. LGTM.

**On CodeRabbit's outside-diff finding** (`SELECT 1; -- comment`): it's factually right — `strip_trailing_semicolon` only removes a *terminating* run of `;`, so a semicolon before a trailing comment survives into the subquery. I confirmed it with sqlglot: `ParseError: Expecting ). Line 2, Col: 13` in the oracle, redshift and trino dialects.

I'd leave it out of this PR. It's pre-existing and repo-wide — the multiline wraps already on `main` (postgres, bigquery, duckdb, athena) fail on that input identically, so this PR neither introduces nor worsens it. And the right fix is one change in `strip_trailing_semicolon` in `base.py`, which fixes every connector at once; adding a third per-connector normalization here would be the thing that has to be undone later.

**Scope for the follow-up PR**, since you're taking it: six single-line wrap sites remain across four files — `canner.py:256,277`, `clickhouse.py:402,419`, `databricks.py:64` (dry-run only; `query()` limits via `fetchmany_arrow` and never wraps), `datafusion.py:35` (`query()` only; `dry_run` goes through `ctx.dry_run`). That's 6, not 8, if you were sizing it from the issue's table.

goldmedal (approved):
Thanks @AmirF194 👍

AmirF194:
Confirmed the sqlglot repro on my end too, and agree base.py is the right layer for that fix rather than duplicating it per connector. Queuing the six remaining wrap sites as a follow-up.

PR #2748, opened 2026-09-16
fix(wren): resolve relative Trino verify path against the connect-time cwd

### LINKED ISSUE #2747: [BUG] Trino connector: relative verify CA path resolves against CWD, not a stable root
## Bug

When a relative path is provided for the `verify` SSL kwarg in `TrinoConnectionInfo.kwargs` (e.g. `verify: certs/ca-chain.pem`), it is passed as-is through `_build_trino_connect_kwargs` → `_apply_trino_ssl_overrides` → `trino.dbapi.connect()` → `requests`.

The `requests` library resolves relative paths against `os.getcwd()`. This means the CA bundle is only found when the process happens to start from the project root. If the application is launched from a different working directory (e.g. as a subprocess of another service, from an IDE, or a CI runner), the connection fails with a certificate error even though the file exists at the expected relative location.

## Where

`wren/connector/trino.py` — `_apply_trino_ssl_overrides()` already coerces string values like `"false"` / `"true"` into proper booleans, but does not resolve relative file paths.

## Expected behavior

A relative `verify` path should be resolved to an absolute path before being passed to `trino.dbapi.connect()`, so SSL verification works regardless of the process's current working directory.

## Suggested fix

In `_apply_trino_ssl_overrides`, after coercion, resolve relative paths with `pathlib.Path`:

```python
from pathlib import Path

def _apply_trino_ssl_overrides(connect_kwargs: dict) -> dict:
    if "verify" in connect_kwargs:
        coerced = _coerce_trino_verify(connect_kwargs.pop("verify"))
        if coerced is not None:
            # Resolve relative CA paths so they don't depend on CWD
            if isinstance(coerced, str) and not Path(coerced).is_absolute():
                coerced = str(Path(coerced).resolve())
            connect_kwargs["verify"] = coerced
    # ... rest of function unchanged
```

Alternatively, accept an optional `root_path` parameter to resolve against, which would let callers anchor to their project root rather than relying on `Path.resolve()` (which still depends on CWD).

## Reproduction

```python
import os
os.chdir("/tmp")  # any directory that isn't the project root

from wren import DataSource, WrenEngine
# configure with verify: "certs/my-ca.pem" (relative path)
# connection fails — requests looks for /tmp/certs/my-ca.pem
```

## Environment

- wrenai 0.14.0
- Python 3.14
- macOS / Linux

AmirF194:
## Summary

`_apply_trino_ssl_overrides` accepts a `verify` CA-bundle path but never resolves it, so a relative path is handed straight to the trino client and ends up depending on whatever the process's working directory happens to be at connect time.

Fixes #2747.

`trino.dbapi.connect(verify="certs/ca-chain.pem")` passes that string through to `requests.adapters.HTTPAdapter.cert_verify`, which calls `os.path.exists(cert_loc)` on it as-is, resolved against `os.getcwd()` at connect time rather than wherever the connection config was built. Reproduced in a clean `python:3.11-slim` container against HEAD `cebd8b20`, importing the real module and calling `_apply_trino_ssl_overrides` after `os.chdir("/tmp")`:

```
cwd: /tmp
output verify: certs/ca-chain.pem
is absolute: False
```

The same override loop already normalizes `verify`/`ssl_verify`/`insecure` into a single boolean or path (`#1502`), so the fix sits in the one place both aliases converge: after that normalization, a relative path is resolved with `pathlib.Path(...).resolve()` before it reaches the trino client, anchoring it to the cwd that is actually live at connection-build time instead of leaving it to whatever cwd happens to be current whenever the query later runs.

## How is it tested?

- New unit tests in `tests/unit/test_trino_ssl_verify.py`: a relative `verify` path and a relative `ssl_verify` alias path both resolve to an absolute path under a `monkeypatch.chdir`'d cwd; an already-absolute path is left untouched. Coverage confirms both new lines in `_apply_trino_ssl_overrides` are exercised.
- Ran the same test file both before and after the fix in a `python:3.11-slim` container, `uv sync --locked` (resolves the published `wren-core-py==0.8.0` wheel, no local Rust build needed): 2 failed / 6 passed on plain `main` (`cebd8b20`), 8 passed with this branch's change applied.
- Full `tests/unit/` (excluding `test_memory.py`/`test_mcp_server.py`, matching `wren-ci.yml`'s `test-unit` job) and `ruff format --check` / `ruff check` on `src/`: 1427 passed, 2 skipped, lint clean. `test_cloud.py`'s 44 failures happen the same way on plain `main` in this same container, unrelated to this change.
- Not run: `tests/connectors/test_trino.py`, the live-Trino integration suite. It needs `testcontainers` and a real Trino container, which I could not stand up here; it does not touch SSL/verify handling either way.

## Duplicate check

`competing-prs.sh` against `core/wren/src/wren/connector/trino.py`, run at scout time and again immediately before this push (64/64 open PRs swept both times): 2 candidates, `#2551` and `#2552`, both drafts from the same author reworking `query()`/`dry_run()` to bind `LIMIT` as a parameter. Neither touches `_apply_trino_ssl_overrides` or SSL/verify handling.

## Summary by CodeRabbit

* **Bug Fixes**
  * SSL certificate verification paths now consistently support relative paths, absolute paths, and paths beginning with `~`.
  * Invalid or missing certificate paths now produce a clear connection error that includes the resolved path.
  * Existing absolute certificate path behavior remains supported.

* **Tests**
  * Added coverage for relative, absolute, home-directory, and missing SSL certificate verification paths.


goldmedal:
Thanks for the thorough writeup — the container repro and the duplicate sweep are both more rigorous than most PRs get.

I don't think the change repairs #2747 though, and the reason points at a design question worth settling before anything lands here.

## The fix anchors to the same cwd `requests` would have used

`Path(verify).resolve()` resolves against `os.getcwd()` — the exact cwd that `HTTPAdapter.cert_verify` would have resolved against. In the failure #2747 describes (process started somewhere other than where the cert lives), the result is the same wrong location, now spelled absolutely. Against this branch:

```
cwd at build: /private/tmp
verify -> /private/tmp/certs/ca-chain.pem
exists?  False
```

The PR reframes the bug as "cwd changed between config-build and query", and against that framing the change is correct. But that window is empty here: `_build_trino_connect_kwargs` is called at `trino.py:474` inside `TrinoConnector.__init__` and `trino_connect(**connect_kwargs)` runs at `:487` — same function — and `git grep chdir -- core/wren` has no production hits, only `monkeypatch.chdir` in tests. So build-time cwd is always connect-time cwd, and the change is a no-op for correctness.

The new tests inherit this: `test_apply_relative_verify_path_resolved_against_cwd` chdirs away from the cert and asserts the path now points into that other directory. That is the failing scenario asserted as the expected result, so the test passes exactly when the bug still bites. A discriminating version writes the CA file under a root, chdirs elsewhere, and asserts `os.path.exists(out["verify"])` — it fails on this branch today, which is the point.

## What should a relative `verify` path mean?

The real question is the anchor, and cwd can't be it: this value is authored in `~/.wren/profiles.yml` (`profile.py`, `_PROFILES_FILE`), a global file with no relationship to whatever directory the process was launched from.

I'd propose the anchor is **the source the value was read from**:

| Value comes from | Anchor |
| --- | --- |
| `~/.wren/profiles.yml` | `$WREN_HOME` — the directory holding that file |
| `--connection-file foo.json` | `dirname(foo.json)` |
| inline `--connection-info '{...}'`, or `?verify=` in a connection URL | cwd |

That last row isn't a concession — for a value typed at a shell prompt, cwd is genuinely what the user meant. The rule covers all three sources with one implementation, and it explains why the current approach misreads the profile case: it treats a value written in a file as if it had been typed on a command line.

**Why the profile file rather than the project root**, since both are stable:

- A profile is global and shared. One `profiles.yml`, and several projects can pin the same profile by name via `wren_project.yml`. Anchoring to the project makes one `verify: certs/ca.pem` line mean different files depending on which project is active — but a CA bundle is a property of the server you're connecting to, not of the project connecting to it.
- It can be read locally. Whoever wrote the line can see the file sitting next to it. Anchoring to the project means a single config line can only be understood after resolving `WREN_PROJECT_HOME` / a cwd walk / the `default_project` in `~/.wren/config.yml` — trading a cwd dependency for a more indirect one.
- A project isn't guaranteed to exist. Running a query with an explicit `--connection-info` and no project at all is legitimate; `$WREN_HOME` exists whenever `profiles.yml` was readable.
- It's the ordinary convention — dbt, docker-compose, tsconfig and `.gitmodules` all resolve relative paths against the file that contains them.

There's already precedent for the general shape in this codebase: `dbt.py:462` resolves a relative db path against `target.project_dir` rather than cwd.

## Where that resolution has to live

Not in `_apply_trino_ssl_overrides`. That function receives a dict and cannot tell whether the values came from `profiles.yml`, a `--connection-file`, or the command line — at that layer the anchor can only be guessed. Resolution belongs where the source is still known (profile loading / connection-file loading), or the anchor has to be passed down explicitly.

It also isn't Trino-specific: `DataFusionConnectionInfo.source` already does a cwd-relative `.resolve()` (`datafusion.py:27`, with `./data` in its `examples`), and `LocalFileConnectionInfo.url` is a path field too. This probably wants one helper at the config layer rather than each connector picking its own meaning.

## A smaller version that could land now

If you'd rather not take on the anchor question, two changes are independent of it, don't foreclose any of the options above, and cover most of the real pain:

1. **`expanduser()`.** `verify: ~/certs/ca.pem` currently becomes `<cwd>/~/certs/ca.pem` — `~` isn't absolute, so it gets joined onto cwd. Since `profiles.yml` is hand-edited, `~` is a likely input, and it's already the convention elsewhere (`context.py:391`/`:396`, `profile_cli.py:131`, `dbt.py:148`). This alone gives users a cwd-independent way to write the path.
2. **Fail fast when the bundle isn't there, naming the path that was tried.** Today the failure surfaces as a certificate error from inside `requests`, which never reveals where it looked.

Plus a line in the docs/warning saying relative paths have no defined base yet — use an absolute or `~/` path. That resolves the practical complaint in #2747 without the current risk: looking fixed while the failure is unchanged.

## What's good

Placement is right — after `verify` / `ssl_verify` / `insecure` converge, so there's one normalization point instead of three. The `isinstance(verify, str)` guard correctly leaves the coerced booleans alone, absolute paths pass through untouched, and the local job list you ran matches `wren-ci.yml`.

One note on `.resolve()` regardless of which direction this goes: it also collapses symlinks, while `os.path.abspath` (or `Path.cwd() / value`) normalizes without following them. Mounted cert bundles are often reached through a rotating symlink, and pinning through it outlives the rotation.

AmirF194:
You're right, the resolve()-at-normalization framing doesn't hold.
Checked: _build_trino_connect_kwargs and trino_connect() run in the
same __init__ call with no chdir between them, so resolving here just
spells the same cwd-dependent path absolutely.

Took the smaller path. Pushed 4a976fe8: expanduser() so `~/certs/ca.pem`
works, and a WrenError naming the resolved path when it's missing
instead of an opaque cert error from requests. Left the anchor question
(profile file vs cwd vs connection-file dir) alone, it's a config-loading
decision beyond one connector.

Rewrote the two relative-path tests to create the file and assert
against it, and the missing-file test fails on the old code exactly the
way you'd expect (raises nothing, since resolve() alone can't tell
missing from present). ruff clean, 10/10 unit tests pass in the repo's
own Docker verify command.

Issue #2498, opened 2026-07-14
Strip trailing semicolon on dry_run / EXPLAIN / describe across connectors


goldmedal:
## Context

Follow-up to #2441 / #2480 (shared `strip_trailing_semicolon` in `connector/base.py`).

Validation paths — `dry_run`, `EXPLAIN`, Snowflake `describe` — don't uniformly strip the trailing `;`, so a trailing semicolon breaks validation even when execution would succeed. This issue tracks routing every connector's validation path through the shared helper, in **one PR**.

**Prerequisite:** blocked by #2480 — rebase onto it so each change is "import + one call", not a new local copy.

### Fold in and close
- [ ] datafusion — dry_run half of #2483
- [ ] spark — dry_run half of #2464
- [ ] snowflake `describe` #2487

### Close as redundant
- [ ] #2463 (datafusion dry_run) — superseded by #2483, which covers datafusion dry_run **and** the unlimited-query path

The query/unlimited-query half of #2483 and #2464 is tracked in #2497.

## Use the shared helper — don't add local copies

spark and snowflake are not migrated by #2480, so their entries here must `import strip_trailing_semicolon` from `base`, not introduce a per-connector `_strip_trailing_semicolon`.

## Done when
- Every connector strips on its `dry_run` / `EXPLAIN` / `describe` path via the shared helper
- Listed PRs folded in and closed

PR #2752, opened 2026-09-20
fix(wren): strip semicolon before trailing comment in strip_trailing_semicolon

### LINKED ISSUE #2740: strip_trailing_semicolon doesn't strip a semicolon before a trailing SQL comment
`strip_trailing_semicolon` in `connector/base.py` only strips a `;` that is literally the last thing in the string (`_TRAILING_SEMICOLONS_RE = re.compile(r"[;\s]+\Z")`). A comment after the semicolon on the same line defeats it:

```python
from wren.connector.base import strip_trailing_semicolon
strip_trailing_semicolon("SELECT 1 AS x; -- pick")
# 'SELECT 1 AS x; -- pick', unchanged
```

Every connector calls this before wrapping for LIMIT/dry_run, so the surviving `;` lands inside the subquery:

```
SELECT * FROM (
SELECT 1 AS x; -- pick
) AS _sub LIMIT 0
```

sqlglot rejects that with `ParseError: Expecting ). Line 2, Col: 13.`

Distinct from #2498/#2499: those route each connector's dry_run/query calls through this same helper, but the helper has this gap no matter where it's called from, so hoisting the call to one boundary wouldn't fix it either.

Fix looks like widening the regex to also eat a trailing line comment (and whitespace) after the semicolon, plus a case in `test_strip_trailing_semicolon.py`. Happy to send a PR if useful.

lihongyuan99:
Closes #2740

## Why

`strip_trailing_semicolon` used `[;\s]+\Z`, so a line or block comment after the terminating `;` defeated stripping and the semicolon survived into the subquery wrap. Observed on current `main` (`871118e`) by calling the real function — not asserted from reading:

```py
>>> strip_trailing_semicolon('SELECT 1 AS x; -- pick')
'SELECT 1 AS x; -- pick'      # unchanged; the ';' is still there
>>> strip_trailing_semicolon('SELECT 1 AS x; /* b */')
'SELECT 1 AS x; /* b */'      # block comment: same
>>> strip_trailing_semicolon('SELECT 1 AS x;')
'SELECT 1 AS x'               # no comment: already worked
```

Reachability: every connector calls this before wrapping for LIMIT/dry_run (e.g. `clickhouse.py:402`, `databricks.py:64`), and the wrapped `SELECT * FROM (SELECT 1 AS x; -- pick ...)` then fails with sqlglot `ParseError` — as #2740 describes. The issue's repro is the same one reproduced above; I confirmed it independently on a clean checkout before touching anything.

## What changed

From the caller's side:

- SQL ending with `; -- comment` or `; /* comment */` now has the terminating `;` (and the trailing comment after it, which is dead text past the statement end) stripped, so subquery wrapping and EXPLAIN succeed.
- Behaviour promises are kept and made explicit: `;` inside string literals is still preserved (`SELECT 'a;b'`), a trailing comment **without** a terminating `;` is now preserved untouched (`SELECT 1 -- keep me` returns unchanged — previously it also happened to be, but by accident of the regex), and `;` inside comments is never treated as a terminator.
- The implementation is a small state scan (single/double-quoted strings, line comments, block comments, with `''`/`""` escape handling) instead of a regex, because the cases above cannot be distinguished by an end-anchored character class without mis-stripping. An unterminated block comment (`SELECT 1 /* ;`) is left untouched.

No connector call sites changed. Note this does **not** attempt #2733 / #2498/#2499's separate scope (single-line wrap with trailing `--` comment and no semicolon at all) — different change, deliberately left out per "one change, once".

## Tests

`tests/unit/test_strip_trailing_semicolon.py` gains 8 parametrized cases (all exercise the real function; each is capable of failing):

- Before the fix: `7 failed, 9 passed` (the 2 new "no semicolon, comment preserved" cases pass both before and after on purpose — they are anti-regression guards).
- After the fix: `16 passed`.
- Full module suite `uv run pytest tests/unit -m "unit and not slow"`: `842 passed, 81 skipped, 7 failed` — the same 7 failures (`test_cloud` git-identity + 6 ONNX tests needing model downloads) exist unchanged on untouched `main`, so zero regressions.
- `ruff check` and `ruff format --check` clean on both touched files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved SQL handling when trailing semicolons are followed by whitespace or comments.
  - Preserved semicolons within quoted strings and comments.
  - Correctly leaves SQL unchanged when a semicolon appears inside a line comment.

- **Tests**
  - Added coverage for line comments, block comments, multiline trailing comments, and semicolons within strings.


goldmedal (changes requested):
**Verdict: request changes.** The direction is right, but the hand-rolled scanner regresses inputs that `main` handles today, and sqlglot's tokenizer already does this job dialect-aware with less code.

### Why the current scan regresses

An end-anchored regex can't tell a trailing comment from `--` inside a string, so moving to a scanner is correct. But the scanner only models `''`/`""` escapes. On any other lexical form it loses sync, never sees the terminating `;`, and returns the SQL unchanged, so the `;` lands inside the subquery wrap. That is the same failure as #2740, now on inputs that used to work. Verified by calling both versions:

| Input | `main` | This PR |
| --- | --- | --- |
| `SELECT 'it\'s' AS x;` (backslash escape: default in MySQL, BigQuery, Spark, Databricks, ClickHouse) | `SELECT 'it\'s' AS x` | unchanged, `;` kept |
| `SELECT E'it\'s' AS x;` (Postgres) | stripped | `;` kept |
| `SELECT $$it's$$ AS x;` (Postgres/DuckDB dollar quoting) | stripped | `;` kept |
| ``SELECT `a--b` FROM t;`` (backtick identifier) | stripped | `;` kept |
| `SELECT [a--b] FROM t;` (T-SQL bracket identifier) | stripped | `;` kept |
| `SELECT 1; 'x'` | unchanged | `SELECT 1` (over-strip, CodeRabbit's point) |

The first row is the realistic one: LLM-generated MySQL/BigQuery SQL uses `\'` routinely. The unit suite is green only because it has no backslash-escape case.

Adding more states to the scanner (backslash, `$$`, backticks, brackets, `q'[]'`, `#` comments) is possible but means maintaining a per-dialect lexer in `connector/base.py`, and every new connector reopens the question of which escapes are modelled.

### Suggested approach: sqlglot's tokenizer

sqlglot is already a dependency and its tokenizer is dialect-aware. Comments attach to the preceding token, so "only comments after the `;`" needs no handling at all: the last token is simply `SEMICOLON`. The tokenizer is far more permissive than the parser, so it does not gate on sqlglot understanding the statement.

```python
from sqlglot import Dialect
from sqlglot.errors import TokenError
from sqlglot.tokens import TokenType

_TRAILING_SEMICOLONS_RE = re.compile(r"[;\s]+\Z")


def strip_trailing_semicolon(sql: str, dialect: str | None = None) -> str:
    try:
        tokens = Dialect.get_or_raise(dialect).tokenize(sql)
    except TokenError:
        # Unterminated string/comment or a form the tokenizer rejects:
        # degrade to the previous end-anchored strip, never worse than before.
        return _TRAILING_SEMICOLONS_RE.sub("", sql)
    cut = None
    for token in reversed(tokens):
        if token.token_type is not TokenType.SEMICOLON:
            break
        cut = token.start
    return sql[:cut].rstrip() if cut is not None else sql
```

I ran this against all 22 parametrized cases in the PR plus the six rows above and dialect-specific forms (Snowflake `@stage`, Oracle `q'[...]'`, DuckDB `$$..$$`, Databricks `:param`, MySQL `#` comments). All correct; the Oracle `q'[]'` case hits `TokenError` and the regex fallback still gives the right answer. Cost is ~0.4 ms on a 200-column query.

**Wiring the dialect:** `get_connector(data_source, connection_info)` in `connector/factory.py` already has the `DataSource`, and `get_sqlglot_dialect()` in `mdl/cte_rewriter.py` already maps it to a sqlglot dialect name (the custom `wren` dialect and `doris` both tokenize fine). Suggest a `dialect` attribute on `ConnectorABC` set by the factory, and a small `self._strip(sql)` helper that the 15 call sites use. Moving the dialect map next to `DataSource` would avoid the connector layer importing from `mdl/`, but that's optional.

### Minor

- The `test_trino_parser.py` expectation change is fine; the old line only recorded behaviour at the time, it wasn't a design decision.
- Please add the six inputs above as regression cases, with the dialect each one needs.
- The docstring should say that anything after the terminating `;` (including tag-style comments) is dropped, which it now does. Good.

lihongyuan99:
Thanks for the detailed review — I ran all three implementations side by side before touching anything, and your table reproduces exactly. The five lexical forms (`\'`, `E'\'`, `$$..$$`, backticks, `[brackets]`) all made my scanner lose sync, never reach the terminating `;` and return the SQL unchanged, i.e. the `;` lands inside the wrap; `SELECT 1; 'x'` over-strips as you and CodeRabbit said. Adopting the tokenizer.

Pushed as `6f7de14`, branch rebased onto current `main`.

What I verified locally, so the numbers are here if useful:

- Your implementation passes the PR's original 16 parametrized cases unchanged (16/16), so the rewrite gives up no existing semantics.
- The six inputs you listed are added as regression cases with the dialect each needs, extended to `bigquery`/`spark`/`databricks`/`clickhouse` for the `\'` row: **all 12 fail against the scanner in `bf11e4c1`**, which is what makes them worth keeping.
- `uv run pytest tests/unit`: `1547 -> 1567 passed`, `7 failed` identical on both sides (the pre-existing `test_cloud` git-identity + 6 ONNX-download cases), `83 skipped` both sides. `ruff check src/` / `format --check src/` clean.
- Three sites keep the public function with an explicit dialect rather than `self._strip`, because existing tests exercise them unbound: `mssql._raw_cursor_sql`, `mssql._flatten_pagination_limit` (called as `MSSqlConnector._flatten_pagination_limit(None, sql)`), and the module-level `_apply_limit` in `mysql.py`/`bigquery.py` (called with two positional args). Otherwise the conversion breaks them.

Two judgement calls I would rather flag than bury:

1. **`tests/unit/test_oracle_strip_unlimited_query.py` had to change.** It asserted on `oracle.py`'s source text (`assert "sql = strip_trailing_semicolon(sql)" in src`), so the rename breaks it. I did not want to delete the guard its comment records ("unlimited branch used to pass `sql` through unchanged"), so it now drives `query()` through a recording connection and asserts on the SQL that reaches the cursor. Its second test built the old code path inside the test itself and asserted nothing about production, so that one is gone. If you would rather keep a source-text assertion, say so and I will restore a dialect-aware version of it.

2. **One behaviour this design gives up.** `SELECT 1; /* unterminated` raises `TokenError`, so the regex fallback runs and the `;` survives — my scanner used to strip it. It is malformed SQL the engine rejects anyway, and guessing where an unclosed comment ends is what this function stopped doing, so I kept your fallback and pinned the result in an explicit test rather than leaving it implicit. If you would prefer unclosed comments to be treated as running to end-of-string, that is a small change and I will make it.

I did not add a guard for an unresolvable dialect name. Checking it first: every value `get_sqlglot_dialect` can return resolves, including `datafusion` → the custom `wren` dialect, which is registered by the `wren/__init__` → `engine` → `cte_rewriter` → `mdl/wren_dialect.py` chain before any connector module loads. No reachable bad value, so per the contribution bar it would have been dead code.

Two asks when you have a moment: CI on `6f7de14` is back to `action_required` (the earlier green run was against the superseded head), so the workflow approval needs re-granting; and since a push does not dismiss a `CHANGES_REQUESTED`, this needs another look from you before it can move.

goldmedal (approved):
Both judgement calls are the right ones, keep them as they are:

1. The Oracle test rewrite is an improvement, not a compromise. Asserting on the SQL that reaches the cursor is what the old comment said it cared about; grepping the connector source only ever pinned a spelling. No source-text assertion needed.
2. Leaving `SELECT 1; /* unterminated` alone is correct. Treating an unclosed comment as running to end-of-string is guessing at a lexical boundary, which is exactly what this function stopped doing. The explicit test pinning the fallback is the right way to document it.

On the dialect guard: agreed, dropped. I checked the registration chain you describe (`wren/__init__` → `engine` → `cte_rewriter` → `wren_dialect`) on `6f7de14` and it holds, and every current `DataSource` resolves. A guard for an unreachable value would be dead code.

Verified locally on `6f7de14`: the touched unit files plus `tests/connectors/test_mssql.py` pass (97 passed, 1 skipped), no import cycle from `factory.py` → `mdl/cte_rewriter.py`, and the 12 regression cases are exactly the ones that matter, given they all fail against the `bf11e4c` scanner.

CI note: the `action_required` state is stale, the run on `6f7de14` has already completed and all 12 checks are green.

One optional nit, not blocking: two added lines in `tests/unit/test_strip_trailing_semicolon.py` (the `test_dialect_specific_lexical_forms` signature and the final assert) aren't `ruff format` clean. CI lint only covers `src/`, so this doesn't fail anything, but the file was format-clean on `main`. Fine to leave or fold into a follow-up.

Approving.

goldmedal:
Thanks @lihongyuan99 👍

